PR #8, opened 2017-02-23
remove broken late static binding compat code

hairmare:
Problem: CentOS still installs php 5.4 if you take the easy route to install it. So far I haven't found any complete features that depend on 5.5 to work. Some of the code has been written in an incompatible fashion without any real need for it.

Solution: Don't use any php 5.4 features for the time being.

If more changes had been needed I would have switched to 5.5 (or even 7) in a hearbeat, turns out the changes we need to support all version are rather minimal and make things simpler over the board.

* [x] don't use `static::class` in a class that does not have a parent class, use `__CLASS__` instead.
* [x] don't use `<ClassName>::class`, since we already know what class we want `"<ClassName>"` ist just fine.


hairmare:
Closing in favor of #14.

PR #10, opened 2017-02-23
Rewrite config from /etc/airtime-saas to plain /etc/airtime

hairmare:
This is the results of 

```bash
sed -i -e 's|/etc/airtime-saas/|/etc/airtime/|' `grep -irl 'airtime-saas' airtime_mvc/ python_apps/` 
```

It might need more testing, the airtime-saas part never really made sense, zf1 has environments for that, ie you would create a saas env based on production for instance.

I beleive legacy upstream was using this to share configuration between customers (ie. analyser runs only once and writes to a shared S3 bucket). I assume they mount the airtime-saas folder onto individual customers instances with a global config. Like I said, I don't feel that this makes sense since all it does is make hacking at the configs in airtime-saas a bit easier. A serious SaaS operation should be using something like puppet or ansible to achieve this.

### DIFF
--- a/airtime_mvc/application/configs/constants.php
+++ b/airtime_mvc/application/configs/constants.php
@@ -110,7 +110,7 @@ define("WHMCS_API_URL", "https://account.sourcefabric.com/includes/api.php");
 define("SUBDOMAIN_WHMCS_CUSTOM_FIELD_NAME", "Choose your domain");
 
 //Sentry error logging
-define('SENTRY_CONFIG_PATH', '/etc/airtime-saas/sentry.airtime_web.ini');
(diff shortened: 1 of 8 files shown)


hairmare:
I redid this after merging #14. It probably needs some testing as I haven't exercised the python parts enough yet.

hairmare:
I tested this and all that's left after #26 is merged will be these two:

```bash
sudo ln -s /etc/airtime /etc/airtime/production
sudo cp ./airtime_mvc/tests/conf/testing/cloud_storage.conf /etc/airtime/
```

Everything is now tracked in #24.

PR #11, opened 2017-02-23
Problem: Billing is always on

hairmare:
Solution: Make billing configurable through LIBRETIME_ENABLE_BILLING and deactivate it

This should catch all the changes needed to deactive billing in LibreTime.

* [x] only call billing when it is enabled
* [x] let super admins edit their info
* [x] dont link to billing if it is disabled


hairmare:
And I'm done for now...

Issue #4, opened 2017-02-21
Problem: ZF1 might not be at its latest version


hairmare:
Since the current install process relies on the OS to install Zend Framework 1 it can sometimes get an an outdated version installed that misses some features (mainly security stuff).

For instance, Debian will be dropping the package with stretchy: [debian#831418](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=831418).

Solution: Use composer to install a vendorized version of the zf1 at version 1.12 and upgrade the code to be compatible with the new framework version.

This way it gets installed the same way as propel1 already does on the master branch. 

As far as I can tell the main changes needed are rather small and maybe all we need to do is implement `Zend_Rest_Controller::headAction` on some controllers in addition to bumping.

I'm not sure why the missing headAction is not throwing errors left and right since [debian](https://tracker.debian.org/pkg/zendframework) doesn't seem to have any patches concerning this.

Issue #5, opened 2017-02-22
Server setup and installation


gusaus:
Resetting [some of the discussion](https://tree.taiga.io/project/robbt-libreairwaves-airtime-fork/issue/27) pertaining to installing LibreTime on a server like AWS. Assuming we'll want to have a guide similar to the [upstream branch](http://sourcefabric.booktype.pro/airtime-25-for-broadcasters/preparing-the-server/), what steps are necessary to get there?

hairmare:
Thanks for moving this from the taiga to GitHub.

That's a nice guide. It contains some alsa stuff and I can't quite see how that would work on AWS (or any other hosted virtualization platform not tailored for such use).

My [legacy upstream fork](https://github.com/radiorabe/airtime#quickstart) has some info on how to install on CentOS parts of which may be generally useful. Some of the more opinionated parts of the install script should also be moved to documentation rather than just blindly installing (ie. it doesn't make sense to tell a user to change the rabbitmq password on the Web-UI when it's too late; if we automate this we should prompt for one or generate one and use that).

gusaus:
Sounds like there are some tasks for the community to work through. Per our previous discussion, possibly we can provide an AWS instance to test install and provide a demo site.

hairmare:
To have everything in one space, this is my feedback from taiga:
> I'm not sure if LibreTime will be lightweight enough for a demo to fit 
> onto the AWS [free tier](https://aws.amazon.com/s/dm/optimization/server-side-test/free-tier/free_np/).
> 
> When we have the install fixed  and things properly packaged for debian. 
> Setting up LibreTime should be relatively straightforward. A demo would 
> still be nice, though I'm afraid of the logistics involved in one. Once the UI 
> is rebranded we should at least put up screenshots.
> 
> One idea I have been thinking about is to additionally offer libretime as a 
> set of docker containers, a docker-compose file and maybe a kubernetes 
> kubelet. That could be built in a way that it's easy to do a quick demo locally.

My point being that we should make it very easy to get up and running for a demo locally.

I believe having a public demo instance would be too much of a burden at this moment.

hairmare:
Amazon Linux seems to be a flavor of CentOS, the RPMs I plan on building should work there!

ergonlogic:
[...]

PR #14, opened 2017-02-27
Vendorize ZF1, fix PHPUnit and configure travis

hairmare:
This a a rather large commit due to the nature of the stuff it is touching. To get PHPUnit up and running again I had to update some deps and I did so by vendorizing them. The vendorizing of zf1 makes sense since distros are already considering to drop it from their repos.

* [x] install vendorized zf1 with composer
* [x] load composer autoloader before zf1
* [x] Implement headAction for all Zend_Rest_Controller based controllers
* [x] switch to yml dataset to get around string only limitations of xml sets (also removed warning in readme)
* [x] use year 2044 as hardcoded date for tests since it is in the future and has the same days like previously used 2016
* [x] make tests easier to run when accessing phpunit directly
* [x] clean up test helper to always use airtime.conf
* [x] switch test dbname to libretime_test
* [x] test db username password switched to libretime/libretime
* [x] install phpunit with composer in a clear version (make tests easier to reproduce on other platforms)
* [x] remove local libs from airtime repo (most of airtime_mvc/library was not needed of in vendor already)
* [x] configure composer autoloading and use it (also removed requires that are not needed anymore)
* [x] add LibreTime prefix for FileNotFoundException (phing had a similar class and these are all pre-namespace style)
* [x] add .travis.yml file
* [x] make etc and logdir configurable with LIBRETIME_CONF_DIR and LIBRETIME_LOG_DIR env (so travis can change it)
* [x] slight cleanup in config for travis not to fail
* [x] add cloud_storage.conf for during test runs
* [x] rewrite mvc testing docs and move them to docs/ folder
* [x] don't use `static::class` in a class that does not have a parent class, use `__CLASS__` instead.
* [x] don't use `<ClassName>::class`, since we already know what class we want `"<ClassName>"` ist just fine.
* [x] fix "can't use method in write context" errors on 5.4 (also helps the optimizer)
* [x] add build status badge on main README.md
* [x] make setup use vendorized libs
* [x] use working copy local config directly during tests
* [x] add -h localhost to test setup instructions
* [x] get rid of `Cannot redeclare class Zend session` in setup
* [x] make sure all parts (including setup, rabbitmq part) use vendorized libs
* [x] update to namespace style amqp lib
* [x] add postgresql user setup part to TESTING.md docs (I'm installing it on ubuntu so I can see this in action, this seems to differ from what upstream postgresql recommends)
* [x] make it not fail with unconfigured soundcloud
* [x] fix installing vendorized autoloader by introducing `php` subdir under `/usr/share/airtime/`
* [x] get rid of StringHelper dependency on login form

Fixes https://github.com/LibreTime/libretime/issues/4

The PHP parts of https://github.com/LibreTime/libretime/pull/10 get obsoleted by this change and it will need rebasing.

This also contains https://github.com/LibreTime/libretime/pull/8, the late static binding compat code was broken for no reason and until CentOS drops php 5.4 there is no reason I'm aware of not to support it. I inlined #8 since the test would be failing on php 5.4 without the change.

If you want to run tests you need to run `composer install` in the root directory and then `cd airtime_mvc/tests && ../../vendor/bin/phpunit`. For the tests to run the user `libretime` needs to be allowed to create the `libretime_test` database. 

Since this is a large one, the only reasonable way to test it is probably to run it locally. Instructions may be found in [docs/TESTING.md](https://github.com/radiorabe/libretime/blob/625f92fe4459bb23ae4476b8a5943f6a0c128290/docs/TESTING.md). With this PHPUnit is fine again and we stop relying on 3rd party installs of zf1.

I hope there will not be a need for anything this large anytime soon again. From here on out the PHP parts should get easier to review since the tests establish a baseline.


hairmare:
First of all, thanks for the feedback. The installer was not touched by this change, as such I would not have expected more of it to work. I have yet to start working on #5 (setup and install).

The tests are meant to be run against a working copy of LibreTime and not an installed version. 

> I fired up a fresh Ubuntu 14.04 VirtualBox to test this and ran into some issues.
> First the install failed due to this line loudCmd "pip install setuptools" 
> When I put in loudCmd "pip install setuptools --upgrade" it works. This was one of the install script fixes I did.

This PR doesn't touch anything related to python. I have been concentrating on the php parts first. Getting python tests to run will be covered #15. 

> But following that it ran through the python aspect of the set-up successfully but when I tried to  access the web site I got this error "Error: Propel not found. Did you install Airtime's third-party dependencies with composer? (Check the README.)"

This should have been solved by installing vendor/ later on.

> When I tried to run the tests as described I got the following errors.
> PHP Warning:  parse_ini_file(/etc/airtime/production/cloud_storage.conf): failed to open stream: No such file or directory in /home/robbt/libretime/airtime_mvc/application/configs/conf.php on line 58
> PHP Warning:  parse_ini_file(/etc/airtime/production/airtime.conf): failed to open stream: No such file or directory in /home/robbt/libretime/airtime_mvc/application/configs/conf.php on line 93

These need to be set up manually (and actually go into `/etc/airtime/testing`). I think I have an idea how they need not be touched manually and will update this.

> PHP Fatal error:  Cannot redeclare class Zend_Session in /home/robbt/libretime/vendor/zendframework/zendframework1/library/Zend/Session.php on line 50

This one is weird. It looks like I missed a require somewhere, but I can't quite tell where.

It looks like you don't have [xdebug](https://xdebug.org/) installed. Installing it would decorate php errors with a stacktrace letting us know where Session.php was loaded.

> Also as per the database instructions for the psql commands to work I needed to add -h localhost to each of them. I'm not sure if this is specific to Ubuntu or not.

I assumed `-h localhost` to be the default. The example was taken from the Travis-CI setup which is based on Ubuntu10 where it works without. I'll add the -h to the docs in any case.

> What is the linux distro you are currently developing this on and how did you set it up ?

I'm still all over the place. For the most part this was developed in a centos based docker container. I tested most parts of it using that. Since travis uses ubuntu that should also have gotten some coverage.

My setup is very much just a collection of development machines and containers already on my laptop. I'm working on breaking this down so I can share it, but it's been more complicated than expected.

Most likely I'll install a Ubuntu box as well, I already have a Debian one that is getting close to useable for dev work.

For the most part the distro should not matter that much. Apart from the outdated upstart stuff everything we do is pretty much distro agnostic.

> The install script doesn't copy the vendor files to the directory it installs to. This was part of my install script fix.
> +    cp -R vendor ${web_root}/../ 
> installed by the Install script Libretime excepts /usr/share/vendor to contain all of the composer files. But this doesn't happen by default. Manually copying the directory allowed the web application to work for me. Not sure if this is a workaround or a fix.

I'd say fix. The vendor files absolutely need to be installed. I plan on packaging these as part of .rpms and .debs we make.

> Also the web-based portion of the  installation failed with this error in the apache log. PHP Fatal error:  require_once(): Failed opening required '/usr/share/airtime/library/php-amqplib/amqp.inc' (include_path='.:
/usr/share/php:/usr/share/pear') in /usr/share/airtime/public/setup/rabbitmq-setup.php on line 3

Good catch, I missed that require and am removing it right now. The setup part missed out on most of the vendorizing as it seems.

> The version I installed via tarball had doctrine  phing  php-amqplib  ZFDebug in the library directory, so not sure why they are missing with this branch.

I moved most of them over to composer since there is no real reason for them to not to be.
* doctrine is still in `library/` since I haven't figured out where it is being used
* phing was never used in the first place and was only put there as a dependency of zf1, to use it there would need to be a `build.xml` file somewhere.
* php-amqplib is install in composer and not explicitly  listed since it is pulled by celery: `massivescale/celery-php  dev-master  requires  videlalvaro/php-amqplib`
* ZFDebug is in `require-dev` as `jokkedk/zfdebug`. I have yet to see it in action though.

hairmare:
I think I touched on all the things you found in my last commits.

I could not reproduce the `Cannot redeclare class Zend_Session` one, but hope it was fixed by the autoloader changes.

The installer will still be broken, I'm looking forward to your PR for that. At RaBe our deploy doesn't really fit the installer since we are deploying in a more enterprisey fashion. I can do more doxing on that a soon as we have packages built as there is no way around them in our deploy.

Robbt:
Ok, from a clean install on Ubuntu 14.04 I still got the error related to Cannot redeclare class Zend session - PHP Fatal error:  Cannot redeclare class Zend_Session in /home/robbt/libretime/vendor/zendframework/zendframework1/library/Zend/Session.php on line 50

Also in order to set-up postgresql I had to set the password for the default user using the following commands. So this might be helpful to add to the documentation for people just starting from a blank slate.

sudo -u postgres psql postgres

# \password postgres

Enter new password: 

The setup got an error with ampq as well. I think it got one screen further and failed at the RabbitMQ part.   
PHP Fatal error:  Class 'AMQPConnection' not found in /usr/share/airtime/public/setup/rabbitmq-setup.php on line 65
Perhaps its not autoloading the classes from vendor ?

hairmare:
I tried to load vendor/ early enough in https://github.com/LibreTime/libretime/pull/14/commits/4936b16bc1dbf48b2322175e62e9ebe0dfe67de8. Looks like i might have missed something, sorry.

Looks like I'll be setting up Ubuntu to reproduce `Cannot redeclare class Zend session`. I haven't been able to see that one on CentOS or Debian.

I'll let you know when it works on a local Ubuntu box.

Robbt:
I don't think you missed anything, I just think that the setup files operate independent of index.php and thus aren't including any of the files/libraries referenced by index.php. I'll try to see if I can do any further debugging to determine why the class is getting loaded prior to that statement on the Ubuntu phpunit test.

hairmare:
https://github.com/LibreTime/libretime/pull/14/commits/21a319767f9b26462e83dc1c79923d7fa041954a is what I had missed. I'm still trying to figure out how to reproduce the session error.

For the reference, here is how I installed on a clean ubuntu trusty to narrow it down.
```bash
git clone https://github.com/radiorabe/libretime.git
cd libretime
git checkout feature/phpunit
sudo apt-get install curl # can't opengroup just add this to posix :/
sudo pip install setuptools --upgrade # needed so install won't fail badly
# hack vendor install 2-liner
vi install
sudo ./install -fiap # with curl and upgraded setuptools

# xdebug to help narrow now errors (adds backtraces to error logs)
sudo apt-get install php5-xdebug 
sudo service apache2 restart
```

After that the installer worked, to get the web interface almost up and running I did:
```
sudo mkdir /etc/airtime/production
sudo cp ./airtime_mvc/tests/conf/testing/cloud_storage.conf /etc/airtime/production/
sudo bash -c 'echo soundcloud_client_id = 0 >> /etc/airtime/airtime.conf'
sudo bash -c 'echo soundcloud_client_secret = 0 >> /etc/airtime/airtime.conf'
sudo bash -c 'echo soundcloud_redirect_uri = http://example.org >> /etc/airtime/airtime.conf'
sudo cp /etc/airtime/airtime.conf /etc/airtime/production/
```

With this, I'm close to a working install. The remaining issue with the installed version is with the autoloader referencing `/usr/share/airtime_mvc` and the installer installing to `/usr/share/airtime`. I'll have to dig into composers docs to figure out how to rewrite this properly or might have to change the install location to make more sense (ie. `/usr/share/airtime/php/{vendor,airtime_mvc}` with the new `php` subdir to properly contain everything).

I added more tasks to the issue and will update this again when I have them figured out.

hairmare:
And I also finally got the `Cannot redeclare class Zend_Session` error, I hadn't realized it was happening when running PHPUnit in `airtime_mvc/`.

hairmare:
It's pulling the system install of Zend_Session as well as the new vendorized install. A workaround is `sudo apt-get purge libzend-framework-php` and I'm removing the package from the requirements. I'll also take care of keeping it out of the include path as well.

hairmare:
I hope to have caught them all this time :)

I added better ubuntu instructions to  [TESTING.md](https://github.com/radiorabe/libretime/blob/898a1ce5453247ac81c0ead3a67953a297be0e7e/docs/TESTING.md) and tested the install all the way through. 

While I was at it I fixed the install to work with vendorized zf1 and added the vendor copying lines in the install script to this pr. 

I stumbled upon some classes with names that where clashing with pluginloading vs. autoloading and went back to manually loading them.

I also removed the zf1 install status from /?config and the status page since this is now taken care of with vendor and should need no user interaction like installing a missing pkg.

The soundcloud service was updated to not fail completely when called without a soundcloud id and secret.

Sorry about all the things that were originally broken in this PR, I'm really looking foward to being able to do smaller ones that are easier to review :P

Robbt:
Cool I tried again, only ran into an issue with airtime_analyzer looking for Failed to open config file at /etc/airtime-saas/production/cloud_storage.conf: No such file or directory.
This is easy to fix but I figure we want to modify it to point it elsewhere since I don't think we are using /etc/airtime-saas anymore.

The tests ran fine if I ran them sudo ../../vendor/bin/phpunit  otherwise they ran into Zend_Log_Exception: "/var/log/airtime/zendphp.log" cannot be opened with mode "a"
Once I did that it ran fine. 

Thanks for all of your work.

hairmare:
I was thinking about doing a largish refactor of the config loading stuff after haven looked at it in python a bit more. It's a bit all over the place right now.

The python parts of rewriting the config path are still in #10. When I update #10 to reflect what was done in here I'll make the path configurable through LIBRETIME_CONF_DIR like I did with php. Let me know if you'd prefer to have those changes on this branch as well.

Having to run tests as root should be fixed with https://github.com/LibreTime/libretime/pull/14/commits/39fb738945b6a668712a96ffed9a8216748fdd2a.

Robbt:
Let's do the refactoring of the python stuff in a separate PR/branch just for the sake of clarity. That makes sense. I think that this is pretty much ready to be committed. I'll give it a final test and then merge it. If I have time I plan on working on the install script/manual install instructions, but I don't know how soon I'll find the time.

hairmare:
Sweet. I'll be adding to TESTING.md when I do the python stuff. Possibly some of it can be reused for the install docs (like the postgresql setup).

I've noticed some cleanup potential in python already. On my hacky dev branch I have tests for airtime_analyzer up and running and I only need to stomp a replaygain in virtualenv on travis bug that has so far eluded a fix. The 90% of analyzer tests that do not need replaygain already work and I'm sure pypo will be testable as well :)

Robbt:
Great. I'm going to push this through. Everything works as far as testing. The install still needs a few workarounds that you included above. Such as adding blank keys for soundcloud to the config and copying a few config files. We can fix those on a subsequent commit to the install script.

hairmare:
Cool. Thanks for the merge!

I think I broke the soundcloud parts when I removed a hardcoded private key from sourcefabric from the repo. I haven't been quite able to figure out how and where the installer creates the initial config file.

I'll be rebasing most of my other commits before I finalize figuring out what is wrong with python tests on travis.

PR #20, opened 2017-03-03
Add Vagrant support

ergonlogic:
Since the installation script appears not to be functioning at the moment, I figured adding a Vagrantfile would make for simpler local dev/test environment setup and teardown.

I've currently got our docs being built and served via Mkdocs within the Vagrant VM, to simplify documentation. I figure to add provisioning via our install script, so that we end up with a full-functioning LibreTime install.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -1,0 +1,13 @@
+# -*- mode: ruby -*-
+# vi: set ft=ruby :
+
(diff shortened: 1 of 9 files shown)


hairmare:
Thanks for the contribution, I tested it and am looking into getting the installer to work in vagrant.

Issue #15, opened 2017-02-27
Run python tests through travis


hairmare:
Currently the available python tests are not getting run regularly.

Travis was configured to run php tests in #14 and python needs to catch up.

Solution: Add python to travis setup and run tests.

This needs a bit of a complex setup since we install python in addition to the already available php.

Travis is on ubuntu10 so I don't see any surprises there. An example of how to install python on a project that is on a different language is at https://github.com/travis-ci/travis-ci/issues/4090#issuecomment-184811689.

* [x] analyzer
* [ ] pypo
* [ ] api_clients

hairmare:
I've got the analyzer mostly figured out and will do a pr on that after cleaning up my local branch.

I looked into pypo and api_clients, to no avail. I'd be glad to pass on the torch for getting those test sets running after i finish the analyzer parts.

hairmare:
Now that analyzer is mostly tested in #21. I'm de-assigning myself from this issue. I'm just not that much of a python buff to figure out what is wrong with the pypo and api_clients tests.

Robbt:
I'm looking into revisiting this, or at least the python testing in general. It seems like tests were wrote for Analyzer using nose but I don't really see any comprehensive testing for pypo, there is a test of the API which just tests that it can connect and a test_modify_cue_in test and that depends upon pytest which needs to be installed with `sudo pip install pytest`

That results in an error with importing the PypoPush class - ```

ImportError while importing test module '/vagrant/python_apps/pypo/tests/test_modify_cue_in.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
tests/test_modify_cue_in.py:1: in <module>
    from pypopush import PypoPush
E   ImportError: No module named pypopush

```

So that will require some troubleshoot or perhaps we should try to rewrite the tests. I have an intuition that proper testing might help us troubleshoot some of the inconsistent behavior we have with pypo in production #232  but I'm also not sure how to go about testing properly since it involves working with liquidsoap.

hairmare:
IMHO pytest is the way to go anyhow.

I've seen that the analyzer nosetests are runnable with pytest, so we're good there.

I haven't had any luck with the pypo tests either (a stumbled over the pypopush error as well).

When it comes to the API we probably won't get away with unit testing alone. Some of the tests are end to end test and not really focussed on units. We will probably have to set up and tear down an API instance to test against for those cases. This also concerns the selenium tests as per #16.

[...]

PR #21, opened 2017-03-03
mostly run analyzer tests on travis

hairmare:
* [x] regonfigured the build matrix with more php jobs and a separate python job (we can add more python jobs later)
* [x] run tests on travis' trusty beta container (it's closer to what we need anyway)
* [x] install packages needed for analyzer tests in build env
* [x] added docs on how to run nosetests locally
* [x] don't run initctl in analyzer setup so setup can also be used on travis (and add it to the install script directly)
* [x] ignore replaygain checks on travis (it has proven quite impossible to get the needed python-gi module to work in the provided virtualenv)
* [x] make sure conditional tests don't kill the build (! return status instead of X)

I tried a lot of solutions to get the replaygain checks to run. I needed to decide that this has gone far enough, maybe someone who is more of a pythonista than me can take a crack at it and get it solved. Even without running those tests on CI/CD there are still plenty others.

This PR only has parts of what are needed for getting python tests running on travis as per #15. I only took a quick shot at anything not analyzer and figured I would not be able to "fix" them without digging a bit deeper (ie. also getting rid of std_err_override).


hairmare:
I noticed a snag, the failing test is showing up as ! and not as X, I'm sorting that out as I type.

Robbt:
Ok, so I started testing this. I realized I needed to type ./travis/python.sh to run the travis tests.
And I ran into 
Building docs...
+ mkdocs build --clean -q
./travis/python.sh: line 13: mkdocs: command not found

So I'm figuring there is something missing. I re-ran the install and it didn't add mkdocs. 

Also wasn't sure what is going on with the Merge branch 'master' into feature/analyzer-on-travis commit. It looks like it is touching a lot of stuff not specifically mentioned here. 

If that is intentional and this PR is just supposed to fix the travis python tests then I can go ahead and commit it as that appears to be working.

Robbt:
Ok, I noticed the docs directory with documentation.md with instructions on how to install mkdocs. Do we want to do that by default or test to see if docs are included or just add a footnote for how to install mkdocs to the python.sh ? 

Either way I think this is ready for commit once we make a decision on whether to fix this now or later.

hairmare:
`mkdocs` gets installed on travis by [master/.travis.yml#L14](https://github.com/LibreTime/libretime/blob/master/.travis.yml#L14). `travis/python.sh` is really only for travis.

Normally you should not need to run it locally, to run tests the recommended way is still `cd airtime_mvc/tests && ../../vendor/bin/phpunit` as per http://libretime.org/testing/. 

I'd rather not add it to the installer since mkdocs is not needed during runtime. The Vagrant setup from #20 will also help working with the docs.

Issue #3, opened 2017-02-20
Uploading tracks, stops randomly.


korotkov13:
When uploading tracks, thru uploader file process stops randomly. It can stop on first file or after upload 20 tracks and stop. Uploading can be resumed only after canceling uploading hanged file. 

Got this in zendphp.log
```
2017-02-19T13:02:12+03:00 INFO (6): ***radio****.com Checking autoplaylist poll
2017-02-19T13:02:20+03:00 ERR (3): ***radio****.com [MediaController.php:140 - postAction()] - chmod(): No such file or directory
#0 [internal function]: exception_error_handler(2, 'chmod(): No suc...', '/usr/share/airt...', 943, Array)
#1 /usr/share/airtime/application/models/StoredFile.php(943): chmod('', 420)
#2 /usr/share/airtime/application/services/MediaService.php(40): Application_Model_StoredFile::moveFileToStor('', 'Quanta - Center...', false)
#3 /usr/share/airtime/application/models/airtime/CcFiles.php(160): Application_Service_MediaService::importFileToLibrary('http://***radio****....', '', 'Quanta - Center...', 1, false)
#4 /usr/share/airtime/application/models/airtime/CcFiles.php(99): CcFiles::createAndImport(Array, '', 'Quanta - Center...')
#5 /usr/share/airtime/application/modules/rest/controllers/MediaController.php(124): CcFiles::createFromUpload(Array)
#6 /usr/share/php/libzend-framework-php/Zend/Controller/Action.php(516): Rest_MediaController->postAction()
#7 /usr/share/php/libzend-framework-php/Zend/Controller/Dispatcher/Standard.php(295): Zend_Controller_Action->dispatch('postAction')
#8 /usr/share/php/libzend-framework-php/Zend/Controller/Front.php(954): Zend_Controller_Dispatcher_Standard->dispatch(Object(Zend_Controller_Request_Http), Object(Zend_Controller_Response_Http))
#9 /usr/share/php/libzend-framework-php/Zend/Application/Bootstrap/Bootstrap.php(97): Zend_Controller_Front->dispatch()
#10 /usr/share/php/libzend-framework-php/Zend/Application.php(366): Zend_Application_Bootstrap_Bootstrap->run()
#11 /usr/share/airtime/application/airtime-boot.php(88): Zend_Application->run()
#12 /usr/share/airtime/public/index.php(66): require_once('/usr/share/airt...')
#13 {main}

[...]

PR #22, opened 2017-03-04
Use Zend_File_Transfer instead of $_FILES

hairmare:
This should give us more information in the case of an error and is the framework idiomatic way to handle a RESTful file upload.

I'm hoping this helps debug https://github.com/LibreTime/libretime/issues/3


Robbt:
This appears to work just fine with some testing. So I'm going to merge it.

PR #23, opened 2017-03-04
stop using reflection in production code

hairmare:
Problem: Reflection is a debugging tool and not really meant for use on production. There are easier ways to check if a class implements an interface with `is_a()`.

Solution: Get rid of reflection use and switch to `is_a()` in TaskManager, refactor the Enum thing that does not make sense since it is not being used.

The `is_a()` solution is really straightforward and has been supported in php since 5.3.9 (dereleased in 2014, probably not compatible anyway, I know of no distro that backports security fixes to it).

The Enum thing was a copy paste hack from stackoverflow and ignored the simple solution mentioned there while not using the features of the advanced one.


Robbt:
This looks good to me.

PR #28, opened 2017-03-04
Podcast naming feature as per #26

### LINKED ISSUE #26: Problem: Podcasts often lack an album ID3 tag
This was on Robbs branch here https://github.com/Robbt/airtime/compare/34db703381f8ccf7fdee3706c3df58f347da7e00...b6556043d49f97fa05facc154b16b8512d786ac2.

My only nitpick was that I'd prefer already set ID3 tags to get left alone (https://github.com/Robbt/Airtime/commit/77c2e57c86dc912956a75b93b6b753752fea934e#commitcomment-20994564). ~I'll merge it without that change and implement that later if it turns out to be an issue for anyone.~ 

This issue is mainly for tracking in the [1 milestone](https://github.com/LibreTime/libretime/milestone/1)

hairmare:
I went ahead and cherry picked what is need to fix #26 and also added a check to not overwrite the album tag if one already exists.

@Robbt Could you check if this (https://github.com/LibreTime/libretime/pull/28/commits/2e9fc432b14899b9463260a26eeca9b890d2a954) still works with your pacifica downloads and I'll merge it.

### DIFF
--- a/airtime_mvc/application/services/PodcastEpisodeService.php
+++ b/airtime_mvc/application/services/PodcastEpisodeService.php
@@ -146,6 +146,7 @@ class Application_Service_PodcastEpisodeService extends Application_Service_Thir
             'url'           => $url,
             'callback_url'  => $stationUrl . 'rest/media',
             'api_key'       => $CC_CONFIG["apiKey"][0],
+            'podcast_name'  => "best podcast",
         );
         $task = $this->_executeTask(static::$_CELERY_TASKS[self::DOWNLOAD], $data);
         // Get the created ThirdPartyTaskReference and set the episode ID so
--- a/airtime_mvc/application/services/PodcastEpisodeService.php
+++ b/airtime_mvc/application/services/PodcastEpisodeService.php
@@ -45,7 +45,8 @@ class Application_Service_PodcastEpisodeService extends Application_Service_Thir
      */
     public function importEpisode($podcastId, $episode) {
         $e = $this->addPlaceholder($podcastId, $episode);
-        $this->_download($e->getDbId(), $e->getDbDownloadUrl());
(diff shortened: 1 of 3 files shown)


Robbt:
Something isn't quite right here - got this error while testing.
[PodcastController.php:106 - postAction()] - Call to undefined method SimplePie::getAuthor() in /usr/share/airtime/php/airtime_mvc/application/services/PodcastService.php on line 94

hairmare:
Shoot, I thought I had tested this in #14. SimplePie was vendorized there as well, looks like it has some compatibilty issues. I'm on it.

hairmare:
The `$rss` getting called is created in [PodCastService.php#L47-L51](https://github.com/LibreTime/libretime/blob/bd7091ae59bc40ac866d7ca3162f06d29e011c2e/airtime_mvc/application/services/PodcastService.php#L47-L51) and it's a basic SimplePie class.

SimplePie seems to call the method `get_author()`, I'm trying to figure out when this changed in simplepie (it was always installed in composer and should not have been updated by #14 so I'm stumped why this is even wrong in the first place). The `get_author()`, `get_authors()` methods seems to implement something completely different than what we are using.

The relevant parts of `composer.lock` are unchanged since https://github.com/LibreTime/libretime/commit/727abac0b84de3af4cc75a4591388a578f0dcdbd and [SimplePie.php#L2129](https://github.com/simplepie/simplepie/blob/6102a15d768c7d81fe48f7e2f1d8cf9e71e5959f/library/SimplePie.php#L2129) already has getAuthor there at that point.

Maybe there used to be another SimplePie on the include_path that we lost. I could not find one in any of the removed library paths.

Since SimplePie is still maintained I will try bumping to it's latest release and then fix forward.

PR #31, opened 2017-03-04
[WIP] Remove Airtime.pro option from settings

hairmare:


Robbt:
This does remove Airtime.pro but it doesn't allow the user to change the server or any of the settings whereas I think before if they selected Custom they could set their own server. We also need to configure the show_source etc. so that it can be modified by the user. 

I think we should probably make it so that the user can at least input their custom server information and remove the default grayed out values that point to airtime.pro from this page before we commit it. I might be able to take a stab at it.

hairmare:
We might need to switch this in the defaults somewhere to make the fields not grayed out. I think if you save it once the fields will become editable.

hairmare:
I need to wipe my DB before testing things like this.

hairmare:
I just retested this with vagrant from #39, the `cc_prefs` fix from https://github.com/LibreTime/libretime/pull/31/commits/4fbaa331c7ba0b74ab6940cb1ca6526169abf44a is not enough to get it working :(

At least the workaround with saving once to make it editable as advertised.

Robbt:
I think the reason it isn't working is due to the way the $useDefaults is set in  airtime_mvc/application/forms/StreamSettingSubForm.php $useDefaults = !Application_Model_Preference::getUsingCustomStreamSettings();
But I'm not sure. I don't know why it isn't editable but I know it is getting the default values which is where the in.airtime.pro comes from when it's first started up.

hairmare:
What was I thinking :man_facepalming:. All this needs is `s/Airtime Pro Streaming/Default Stream Settings/` in the label.

Robbt:
Any reason we closed this without merging ?

hairmare:
Yes, I think #61 is works and looks smarter (we still need to clean up the "out.airtime.pro" default settings elsewhere though).

PR #33, opened 2017-03-04
Added production directory during setup and used --upgrade parameter …

### LINKED ISSUE #32: Problem: Install.sh fails on Ubuntu when installing setuptools
There is a failure for the install script when installing on Ubuntu, there is also a simple fix that I plan to commit.

Robbt:
…for pip install setuptools fixes #32


hairmare:
LGTM, I'll merge as soon as travis is through

Issue #24, opened 2017-03-04
Problem: Installer is broken due to missing soundcloud, env and rabbitmq settings


hairmare:
This was introduced by #14 and needs to be fixed. I looks like it needs a multi pronged fix since there are two main failures.

1. missing soundcloud config
2. no support for env subdir in conf dir
3. setup does not create `cloud_storage.conf` and leaves `rabbitmq-analyzer.ini` empty

Workaround from https://github.com/LibreTime/libretime/pull/14#issuecomment-283634730

```bash
# after completing setup (ie no error from rabbitmq, but it hangs in the end)
#sudo bash -c 'echo soundcloud_client_id = 0 >> /etc/airtime/airtime.conf'
#sudo bash -c 'echo soundcloud_client_secret = 0 >> /etc/airtime/airtime.conf'
#sudo bash -c 'echo soundcloud_redirect_uri = http://example.org >> /etc/airtime/airtime.conf'

# sudo ln -s /etc/airtime /etc/airtime/production
#sudo cp ./airtime_mvc/tests/conf/testing/cloud_storage.conf /etc/airtime/

# do these before running sudo commands in the last setup step!
sudo bash -c 'cat > /etc/airtime/rabbitmq-analyzer.ini <<EOD
[rabbitmq]
host = 127.0.0.1
port = 5672
user = airtime
password = airtime
vhost = /airtime
EOD'

# start icecast
sudo sed -i -e 's/ENABLE=false/ENABLE=true/' /etc/default/icecast2
sudo service icecast2 start

# reload page and you should be on the default radio page
```
Solution: Add soundcloud_* defaults to config file during initial generation and make `/etc/airtime/<env>/` installer location sane.

* [x] add soundcloud to default config #25 
* [x] make default production env and symlink it to conf dir in setup (so we catch all python things as well)
* [x] add a quick and dirty `cloud_storage.conf` install to the setup script
* [ ] figure out why `rabbitmq-analyzer.ini` is empty and fix that
* [x] use upgraded setuptools #32 #33 

We can refactor it to fall back on `LIBRETIME_CONF_DIR/*.conf` if `LIBRETIME_CONF_DIR/ENVIRONMENT/*.conf` is missing later, I need to read up on how to best do that in python.

Do we want to have nice setup steps for the soundcloud and cloud storage steps? Is anyone currently not using a `file` backend in `cloud_storage.conf` that can share an S3 example?

Robbt:
[...]

PR #39, opened 2017-03-06
Install LibreTime in vagrant env

hairmare:
Uses the old installer to get LibreTime installed quick and dirty on vagrant.

It uses the `install -fIap` command that does a local install and points the apache config
directly to the local working copy mounted in /vagrant.

While we don't have fancy autoloading for libretime like the docs do, this way
it's already easy to work against a local branch.

I'm not sure if the `-I` arg to the installer also covers the python parts of libretime.

* [x] update vagrant box to grab more memory (the default 512 MB results in oom kills) e394df83a5709459df61163f68b9982f0c2ee80d
* [x] solve port mapping dilemma bd2f96acf2cd76b26b8c47b6fa1408b354e2cf46
* [x] accessible icecast (it is still up to the user to enable it, an example is in #24)
* [x] virtual audio interface for testing line in/out features (works with pulseaudio out of the box)
* [x] more vagrant [docs](https://github.com/radiorabe/libretime/blob/feature/vagrant-install/docs/vagrant.md)


hairmare:
This will really only be useable to hack on the php parts out of the box. analyzer needs some coaxing due to the config being based on 8080 from the webui and the actual apache running on 80.

I ended up hacking the apache config and letting the vhost listen on both ports to get the results up and running as a dev env. I'm not sure where analyzer is getting it's config from, it' doesn't seem to be `base_port` in `/etc/airtime/airtime.conf`.

hairmare:
There external `8080` also somewhat conflicts with icecast running on `8080` internally. Does anyone have a preference when it comes to ports? Should we switch to a [private network ip](https://www.vagrantup.com/docs/networking/private_network.html)?

Robbt:
I'm arbitrarily using 9080. A private network sounds fine as long as its documented, do we have docs for running vagrant to test anywhere ? 

Analyzer pulls the port from the conf after the web setup is ran so I think it is /etc/airtime/airtime.conf, at least based on my testing.

Does this write a default rabbitmq.ini file ?

hairmare:
Some docs are on [getting started](https://github.com/LibreTime/libretime/blob/master/docs/index.md#getting-started) in the main docs page.

I think 9080 might be a sensible choice, at least locally I have nothing there most of the time. Analyzer does pull the port from the config, the problem is with this install the port gets NATed `8080 -> 80` which leads to the setup UI writing to `8080` to the file per default. I'll need to do some testing with private and/or public networks to figure out which one works for us.

All this does is call `install -fIap`, so no it doesn't fix the rabbitmq file issue.

hairmare:
This should now install the whole shebang including virtual audio and expose all relevant ports in a straightforward way ready for devs.

The [docs](https://github.com/radiorabe/libretime/blob/2350650e2624546ea934727b354783d2a2ec86d7/docs/index.md) have been updated to reflect what ports to expect after install. You might have to `vagrant reload` to get audio working after the initial install (and LibreTime still needs #44 and #40 and similar fixes, I tested audio out with `mpg123` on the cli).

install
```diff
@@ -414,12 +434,22 @@ if [ "$apache" = "t" ]; then
...
-            sed -e "s@WEB_ROOT@${web_root}@g" ${SCRIPT_DIR}/installer/apache/airtime-vhost-2.4 > /etc/apache2/sites-available/${airtimeconfigfile}
-        else
-            sed -e "s@WEB_ROOT@${web_root}@g" ${SCRIPT_DIR}/installer/apache/airtime-vhost > /etc/apache2/sites-available/${airtimeconfigfile}
+        apache_template_file=${SCRIPT_DIR}/installer/apache/airtime-vhost-2.4
+        if [ "$apacheversion" = "1" ]; then
+            # fall back to apache 1 config
+            apache_template_file=${SCRIPT_DIR}/installer/apache/airtime-vhost
         fi
+        sed \
+            -e "s@WEB_PORT_LISTEN@Listen ${listen_port}@g" \
```
hairmare: This will break installs on the regular port 😞

hairmare:
I just changed the milestone on this after installing it one final time on a clean working copy. This has been my php dev install since I created it and it finally has working playout thanks to 0e8d0868e8eae8e7f245ee82fbae41652853e5bb.

The whole install process works through setup and the welcome guide through to having a red on-air button and listening to an "airtime pro" stream through the stream player.

Robbt:
It works. Awesome.

PR #40, opened 2017-03-06
Hardware audio output added, install modifications, autoplaylist.

comiconomenclaturist:
You may just want to just grab the hardware audio output section in airtime_mvc/application/forms/StreamSetting.php

The rest are basically the same changes as the dropbox version to get the installer working.


hairmare:
I created #41 since I did not want to push to your master branch. Feel free to op a new pr if you want anything else sorted. You can also create issues if we missed something from Robbs tarball that hasn't got an issue yet.

Thanks!

PR #52, opened 2017-03-09
adding the default rabbitmq-analyzer.ini file to fix install script

### LINKED ISSUE #24: Problem: Installer is broken due to missing soundcloud, env and rabbitmq settings

Robbt:
This fixes #24 by adding a rabbitmq-analyzer.ini file.

### DIFF
--- a/airtime_mvc/build/rabbitmq-analyzer.ini
+++ b/airtime_mvc/build/rabbitmq-analyzer.ini
@@ -1,0 +1,7 @@
+[rabbitmq]
+host = 127.0.0.1
+port = 5672
+user = airtime
+password = airtime
+vhost = /airtime
+


hairmare:
Thanks! I did a vagrant re-install and uploading works out of the box in #39. I'm off to fixing a regular install from that branch.

PR #60, opened 2017-03-11
Fix unicode issues in podcast downloader

hairmare:
The podcast downloader fails pretty badly when the podcast name contains non ascii chars. The main fail happens during logging; I have learnt way to much about pythons stupid unicode implementation.

This adds addtional debug logging and also outputs the real reason a download fails properly. The content of the tags should be written as UTF-8 or whater is input into it, this commit mainly touches (and fixes) logging.

### DIFF
--- a/python_apps/airtime-celery/airtime-celery/tasks.py
+++ b/python_apps/airtime-celery/airtime-celery/tasks.py
@@ -8,6 +8,7 @@ import urlparse
 import posixpath
 import shutil
 import tempfile
+import traceback
 from mutagen.mp3 import MP3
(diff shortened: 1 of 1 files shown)


Robbt:
Cool, in regards to the Album downloading. What would be the easiest way to make it so that there was an option to override the Album (even if it exists). In my testing I realized that even some podcasts that have Album's have bad metadata, DemocracyNow for instance has the Year as the album. 
In my opinion there needs to be some easy way to connect tracks from a podcast back to the podcast they came from for SmartBlocks and the like. But it could make sense to have an option that could be configured. I just don't know the best way to query that in the python side of things.

PR #64, opened 2017-03-12
Vagrant Debian support (and experimental CentOS)

### LINKED ISSUE #47: Problem: Debian install has issues
The docs [recommend Debian over Ubuntu](https://github.com/LibreTime/libretime/blob/0f8c3b7c92496af1a59aba04b1573dacd235e0ad/docs/manual/preparing-the-server/index.md) but the actual install currently only works on Ubuntu. I think Debian is a better recommendation than Ubuntu since it is a less complex OS. If you ask **rms** he would go on a tangent about Debian being the more Free 😉 

Since this was only tested on Ubuntu lately nothing has been done to keep Debian working. Most of the issues are rather small stuff (like Debian not having `curl` on the CLI per default). Some of the newer python apps like analyzer and celery only have upstart scripts (where in older cases like pypo upstart falls back to the existing SysV scripts).

Solution: Test and fix the installer for Debian, convert init scripts to SysV for maximum compatibility. Fix the version recommendation in the docs.

Ubuntu will still be using upstart to keep the SysV scripts running, I think this offers the best of both worlds. Other distros (ie. CentoOS and Debian testing) will be supported by systemd anyway and we will only need one set of scripts once SysV is completely gone from our support matrix.

hairmare:
This changes the Vagrant setup to support multiple installations as multiple
boxes. In addition to Ubuntu Vagrant can now be used to install on Debian
as well as on CentOS.

I took the chance to clean up the .deb install a bit and backported analyzer
and celery to SysV proper so it runs there. Some of the distro specfics were
moved to the install script from the python setup scripts to acheive this.

For the CentOS support I added a rather involved OS prepare script. In the
long term this will be added to the preparing-the-server docs we already have.

I had to switch the default port to http-alt (8080). On CentOS 9080 is registered
for ocsp and getting it to work for apache without hacking SELinux is hard. I
think 8080 is the RFC way to go anyhow. If anyone want to override this it
should be rather easy using the --web-port arg and by hacking Vagrantfile.

The PyOpenSSL code has been refactored for all the distros that the Vagrantfile
now supports.

As far as my checks go, I tried this code with all the distros, uploaded a track
and downloaded a unicode and a ssl podcast and was able to listen to them
in each case.

In the experimental CentOS case, the UI is not up to spec since services
need to get scheduled through systemctl and the status overview (ie. on the /?config page)
do not work properly. They need to be as follows:

```
sudo systemctl start airtime-playout
sudo systemctl start airtime-liquidsoap
sudo systemctl start airtime_analyzer.service
sudo systemctl start airtime-celery.service
```

Also, the boxes can't run at the same time since they use the same ports. You can `vagrant suspend <os>` and `vagrant up <os>` or `vagrant reload <os>` and keep more than one ready to run.

Fixes #47


Robbt:
Gave it a shot but ran into an issue with the Centos install regarding the cryptography install. 
So perhaps we need to add the centos equivalent of build-essential to add gcc etc.

==> centos:     unable to execute gcc: No such file or directory
==> centos:     error: command 'gcc' failed with exit status 1
==> centos:     
==> centos:     ----------------------------------------
==> centos: Command "/usr/bin/python2 -u -c "import setuptools, tokenize;__file__='/tmp/pip-build-KlFLgP/cryptography/setup.py';exec(compile(getattr(tokenize, 'open', open)(__file__).read().replace('\r\n', '\n'), __file__, 'exec'))" install --record /tmp/pip-SzBEqb-record/install-record.txt --single-version-externally-managed --compile" failed with error code 1 in /tmp/pip-build-KlFLgP/cryptography/
==> centos: You are using pip version 8.1.2, however version 9.0.1 is available.
==> centos: You should consider upgrading via the 'pip install --upgrade pip' command.

Robbt:
Also ran into the need to install nfsd w/ this install. It worked fine after I installed it but we should document the dependency (if we havent already).

Robbt:
I'm going to merge this through, since it appears to work for Debian, and we can refine the CentOS stuff in another PR. Also since I started using it for my development environment.

hairmare:
Thanks for the merge.

I think gcc was installed by the vbguest plugin in my local install automatically. Did you need to install nfsd on the host or in the guest?

I think we are ready for a 3.0.0-alpha release. I'll do some more testing (and re-install my 3 boxes once more) and then tag a release if everything looks ok (most likely in a day or two).

Could you assign any issues you find that need fixing before the release to the [3.0.0-alpha milestone](https://github.com/LibreTime/libretime/milestone/1) so I notice them should I proceed to tag the release.

PR #74, opened 2017-03-13
modified smart block to not exceed the time limits.

### LINKED ISSUE #73: Problem: Smart Block time-based limits fill tracks until they exceed the limit.
If you have a smart block limited to 30 minutes it currently adds tracks until the time of the tracks is more than 30 minutes. For shorter limits for instance PSAs or filler tracks, one longer track can throw off the whole playlist. In my opinion in concordance with the language, tracks that exceed the limit should not be added and the algorithm should instead try to get as close to the limit as possible.

Robbt:
this fixes #73 - it doesn't apply an optimal search algorithm to make tracks meet the length, it just doesn't add tracks that would exceed the time length, and cycles through up to 500 tracks adding items it can without exceeding the time length.

### DIFF
--- a/airtime_mvc/application/models/Block.php
+++ b/airtime_mvc/application/models/Block.php
@@ -1310,10 +1310,18 @@ SQL;
             $id = $iterator->current()->getDbId();
             $fileLength = $iterator->current()->getCueLength();
             $length = Application_Common_DateHelper::calculateLengthInSeconds($fileLength);
-            $insertList[] = array('id'=>$id, 'length'=>$length);
-            $totalTime += $length;
-            $totalItems++;
-            
+            // need to check to determine if the track will make the playlist exceed the totalTime before adding it
+            // this can be quite processor consuming so as a workaround I used the totalItems limit to prevent the
+            // algorithm from parsing too many items.
+                $projectedTime = $totalTime + $length;
+            if ($projectedTime > $limit['time']) {
+                $totalItems++;
+                		}
+	        else {
+                $insertList[] = array('id' => $id, 'length' => $length);
+                $totalTime += $length;
+                $totalItems++;
+            }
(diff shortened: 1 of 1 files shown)


hairmare:
I can't seem to get this to work. When I schedule a 50 min smart block to a 15 minute show  I still get massive overbooking.

Robbt:
This doesn't prevent overbooking on the show level, but instead it keeps the smart block from overbooking it's own limit. For instance a 50 minute limit should be a hard limit, it should not schedule 51+ minutes of tracks. Since overbooking happens and might be necessary to prevent dead-air (in instances where there is no exact fit of tracks) we should be probably allow that to happen still.

hairmare:
Ok, now I get how it operates on smart blocks. Your reasoning makes sense to me.

I still cannot figure out how it works though. Does the smart block need anything special? I can't get it to work with either static or dynamic smart block.

After creating the 15 min smart block I just schedule it through the calendar, is that ok?

Robbt:
Yeah. And then it should add tracks to get close to but not exceed 15 minutes. It will tend to cluster short tracks at the end as it rejects tracks that will exceed the length (until it cycles through 500 tracks). When I tested it with 1 minute track it appeared to be working for me.

hairmare:
Weird, I was doing it the right way... This is my smart block:

screen shot 2017-03-13 at 17 51 01

And these are the results I'm seeing after adding the block to the show.

screen shot 2017-03-13 at 17 51 31

Am I wrong to assume that the last track should have not made it into the show?

Robbt:
Weird, yeah, that shouldn't be happening if my code is correct. Let me test it some more.

Robbt:
I just tested it and it's working perfectly fine for me. Maybe the modified Blocks.php file isn't being used by your testing environment for some reason.

hairmare:
I'm not quite sure what I changed, somehow my test env was not picking up on the branch switch, once I touched the file if worked:

screen shot 2017-03-13 at 18 24 02

PR #84, opened 2017-03-14
[WIP] Making show and master source ports and mount points editable.

Robbt:
This works to make it so that the ports and mount points are editable and the stream URL is derived from the host.

Currently the python script has an error when attempting to regenerate the liquid soap configuration.

Also the override option wasn't implemented yet because it was mainly implemented in javascript and I wanted to get this working first.

It could probably be simplified as there were a few changes that were unneeded but they didn't have a negative impact. Mainly switching from holding the static values in an array to actually writing them to the database.


Robbt:
Currently getting the following error in the regenerate_liquidsoap_conf function (which I copied and pasted from the 2.5.x version).

  File "/usr/local/lib/python2.7/dist-packages/airtime_playout-1.0-py2.7.egg/pypo/pypofetch.py", line 258, in regenerate_liquidsoap_conf
    if "output_sound_device" in s[u'keyname'] or "icecast_vorbis_metadata" in s[u'keyname']:
TypeError: string indices must be integers

Line 257-25 is
         for k, s in setting:
            if "output_sound_device" in s[u'keyname'] or "icecast_vorbis_metadata" in s[u'keyname']:
  Need to ensure the correct data type is being passed into the function via rabbitmq.

Robbt:
So the k and s were simply the keys and their value without the need to reference it via s[u'keyname'] that became k and s was s[u'value'] and it worked and rewrote the liquidsoap config but I'm not 100% sure that all of the checks are working as intended.

airtime_mvc/application/forms/LiveStreamingPreferences.php
```diff
@@ -7,6 +7,7 @@ public function init()
     {
         $CC_CONFIG = Config::getConfig();
         $isDemo = isset($CC_CONFIG['demo']) && $CC_CONFIG['demo'] == 1;
+        $isStreamConfigable = Application_Model_Preference::GetEnableStreamConf() == "true";
```
hairmare: This was used in 2.5 to do something with the demo mode, it's currently unused and this line can be removed.

airtime_mvc/application/models/StreamSetting.php
```diff
@@ -464,7 +464,7 @@ public static function setDjLiveStreamPort($value)
 
     public static function getDjLiveStreamPort()
     {
-        return self::getValue("dj_live_stream_port", 8001);
+        return self::getValue("dj_live_stream_port", 8002);
```
hairmare: Do we need to change the default port? I think localhost:8001/show works fine.

airtime_mvc/application/controllers/PreferenceController.php
```diff
@@ -262,6 +262,41 @@ public function streamSettingAction()
                     //Application_Model_RabbitMq::PushSchedule();
                 }
 
+                // pulling this from the 2.5.x branch
+                if (!Application_Model_Preference::GetMasterDjConnectionUrlOverride()) {
+                    $master_connection_url = "http://".$_SERVER['SERVER_NAME'].":".$values["master_source_port"].$values["master_source_mount"];
+                    if (empty($values["master_source_port"]) || empty($values["master_source_port"])) {
```
hairmare: This should probably be:

```php
if (empty($values["master_source_port"]) || empty($values["master_source_mount"])) {
```

I kinda liked how SaaS was using `parse_url(`) on `Application_Model_Preference::GetMasterDJSourceConnectionURL()` to achieve this without only keeping `MasterDJSourceConnectionURL` for display purposes.

Robbt: Yeah honestly this was just there from my attempt to integrate the override functionality before I abandoned it for this commit. The question is do we want to integrate the ability to override the URL ? It seemed like the functionality was primarily a edge-case with a firewall requiring the rewriting of the URL.

airtime_mvc/application/forms/LiveStreamingPreferences.php
```diff
@@ -140,7 +153,53 @@ public function updateVariables()
...
+                    $sock = socket_create(AF_INET, SOCK_STREAM, SOL_TCP);
+                    try {
+                        socket_bind($sock, 0, $show_source_port);
+                    } catch (Exception $e) {
+                        $element = $this->getElement("show_source_port");
+                        $element->addError(sprintf(_("Port %s is not available"), $show_source_port));
+                        $isValid = false;
+                    }
+                    socket_close($sock);
+                }
```
Robbt: This isn't working. I'm not sure why.

airtime_mvc/application/models/StreamSetting.php
Robbt: The Master port should be 8001 and the DJ port should be 8002 (or something different), since they can't be the same.

airtime_mvc/application/forms/LiveStreamingPreferences.php
```diff
@@ -125,12 +138,12 @@ public function updateVariables()
...
-                      'show_source_port'            => isset($showSourceParams["port"])?$showSourceParams["port"]:"",
-                      'show_source_mount'           => isset($showSourceParams["path"])?$showSourceParams["path"]:"",
+                      'master_source_host'          => isset($masterSourceHost)?$masterSourceParams["host"]:"",
+                      'master_source_port'          => isset($masterSourcePort)?$masterSourceParams["port"]:"",
+                      'master_source_mount'         => isset($masterSourceMount)?$masterSourceParams["path"]:"",
+                      'show_source_host'            => isset($showSourceHost)?$showSourceParams["host"]:"",
+                      'show_source_port'            => isset($showSourcePort)?$showSourceParams["port"]:"",
+                      'show_source_mount'           => isset($showSourceMount)?$showSourceParams["path"]:"",
                       'isDemo'                      => $isDemo,
                     )
```
Robbt: I wasn't sure if this was important. The masterSourceParams was an array that never got filled that came from the GetMasterConnectionUrl function (but that function seemed to return a URL), so I wasn't sure how the key value array was supposed to be built for the airtime.pro functionality of the form.

airtime_mvc/application/models/StreamSetting.php
hairmare: Sounds ok, I was assuming it would only use different mountpoints. We should NAT 8002 in vagrant per default as well in this case.

airtime_mvc/application/controllers/PreferenceController.php
hairmare: Is it an edge case though? Most folks that will be using the instreamer will do so from the internet and I'm not sure everyone has the means to do this over a proper VPN tunnel. For smaller stations using a consumer grade internet uplink I always saw this as a handy way for them to communicate the proper IP or DynDNS entry to staff that needs it.

airtime_mvc/application/forms/LiveStreamingPreferences.php
hairmare: Personally I would not do keep this check, better add a warning telling users they should check the port on their own if they want to change it. Personally this check won't make sense for my use-case since we plan on having dedicated playout systems that only have db, amqp and api access.

airtime_mvc/application/forms/LiveStreamingPreferences.php
hairmare: I think is was using `parse_url` to create the array from the url out of the config but never saw it in action due to my url pref always being empty.

Robbt: Ok that makes sense. I wasn't familiar with how the function parse_url worked. Perhaps that is a more elegant solution than putting everything in as a separate key, but I think that the python liquidsoap reconfigure function depends upon them being a separate key for its checks. So without refactoring that.

airtime_mvc/application/models/StreamSetting.php
Robbt: Yeah, I'm just going off the code requiring them to be different. NATing 8002 is a good idea.

airtime_mvc/application/controllers/PreferenceController.php
Robbt: Yeah fair enough. I think that we might want to do that. I can work on reconfiguring the javascript.

airtime_mvc/application/forms/LiveStreamingPreferences.php
Robbt: Ok, sounds like a plan, any ideas on the best way to add a warning ?

hairmare: Either with a tooltip or by calling `setDescription()` on the form element (like in with the CORS form). Also remember to wrap the string in the `_()` method call so it can be translated.

airtime_mvc/application/controllers/PreferenceController.php
hairmare: I also found [Cool Mic](https://coolmic.net/) which can read a QR code with the URL to make set up easy. Adding such a code is very easy as long as we have a nice URL, Cool Mic would make for a great rewrite of parts of the smartphone journalism docs.

airtime_mvc/application/forms/LiveStreamingPreferences.php
hairmare: Python seems to have a [`urlparse`](https://docs.python.org/2/library/urlparse.html) module as well. I need to step through the file generation if I want to make any sense of it as I don't currently really understand what (and why) it is doing. It looks rather convoluted and I have a hard time believing that it can't be done in a simpler fashion.

airtime_mvc/application/controllers/PreferenceController.php
Robbt: That is a awesome find. We should definitely integrate support for that.

airtime_mvc/application/forms/LiveStreamingPreferences.php
Robbt: Yeah. the code does seem to be working. I'm not sure that we need to overhaul it at this point. If someone wants to spend the time picking it apart it could definitely use refactoring as it is kind of cludgy. The only way I could get the log data from airtime-playout though was to run it sudo -u www-data airtimeplayout from the CLI. The logs for pypo don't appear to be working as they only have a few lines of data.

Robbt:
Ok, I think this should be mostly working. It needs testing, and possibly more clean-up. I noticed that for some reason the URLs had two slashes in them even though I commented this out. I did manage to fix most of the javascript, although it is still using an ajax call to save the results (which is possibly where the two slashes are coming from).

Robbt:
This should be ready to go as far as the functionality. 
The validation where it attempted to manually check the port was stripped out but the ZendForm based validators don't appear to be working. I'm not sure what the problem is here as I'm not 100% familiar with the way ZendForm works. We

airtime_mvc/application/forms/LiveStreamingPreferences.php
```diff
@@ -153,53 +154,7 @@ public function updateVariables()
...
-                    socket_close($sock);
-                }
-            } else {
-                $isValid = false;
-            }
-        }
-
-        return $isValid;
-
+        return $isValid = parent::isValid($data);
```
hairmare: This will always return `true` since the assigning the results of `parent::isValid($data)` always evaluates to `true`.

I you want to just have the parents `isValid()` method do it's thing I would recommend removing the whole method fom the subclass.

airtime_mvc/application/forms/LiveStreamingPreferences.php
```diff
@@ -63,42 +64,55 @@ public function init()
 
         // Master source connection url parameters
         $masterSourceHost = new Zend_Form_Element_Text('master_source_host');
-        $masterSourceHost->setAttrib('readonly', true)
-            ->setLabel(_('Host:'))
-            ->setValue(isset($masterSourceParams["host"])?$masterSourceParams["host"]:"");
+        $masterSourceHost->setLabel(_('Host:'))
```
hairmare: s/Host/Master Source URL/

airtime_mvc/application/forms/LiveStreamingPreferences.php
```diff
@@ -63,42 +64,55 @@ public function init()
...
         $this->addElement($masterSourceHost);
 
+        //liquidsoap harbor.input port
+        $betweenValidator = Application_Form_Helper_ValidationTypes::overrideBetweenValidator(1024, 49151);
+        $m_port = Application_Model_StreamSetting::getMasterLiveStreamPort();
         $masterSourcePort = new Zend_Form_Element_Text('master_source_port');
-        $masterSourcePort->setAttrib('readonly', true)
-            ->setLabel(_('Port:'))
-            ->setValue(isset($masterSourceParams["port"])?$masterSourceParams["port"]:"");
+        $masterSourcePort->setLabel(_('Master Source Port:'))
```
hairmare: s/Port/URL/

hairmare:
From my tests it looks like form validation is actually working but the error is not properly being displayed due to the element being part of a subform. I haven't been able to figure out why though.

Robbt:
I reverted back to the original form from saas-dev and tried to see if validation was working to make it digits only on the switch transition fade and it was not. It allowed me to enter any characters I wanted and gave no error messages and even saved the letters to the database. 

So validation is currently broken on this subform as far as I can tell. At least I've determined it's not something wrong with the validation code. I'm trying to figure it out but I don't understand the zend framework enough for it to be obvious.

hairmare:
I just reproduced it as well. We should address this somewhere else as it affects all the elements and probably other forms as well. Maybe it's time to switch to client side html5 validation in addition to fixing what is wrong with zf1.

Robbt:
I think that what is happening is unique to this form as validation works on every other form including the general preferences form. I think what may be different is the form isn't actually getting submitted via post due to the javascript in streamsetting.js which gives the pop-up notification.

Here is a relavent excerpt - 
481         var confirm_pypo_restart_text = $.i18n._("WARNING: This will restart your stream and may cause a short dropout for your listeners!");
482         if (confirm(confirm_pypo_restart_text)) {
483             var data = $('#stream_form').serialize();
484             var url = baseUrl+'Preference/stream-setting';

hairmare:
The XHR is going to the right controller, to me it looks like the validation in `streamSettingAction` is quite redundant and doesn't really handle subforms. `$value` seems to be what gets validated but the values are never added to the array.

airtime_mvc/public/js/airtime/preferences/streamsetting.js
```diff
@@ -159,9 +161,12 @@ function setLiveSourceConnectionOverrideListener(){
     });
     
     master_dj_actions.find("#reset").click(function(event){
-        var port = $("#master_harbor_input_port").val();
-        var mount = $("#master_harbor_input_mount_point").val();
-        var url = "http://"+location.hostname+":"+port+"/"+mount;
+        var port = $("#master_source_port").val();
+        var mount = $("#master_source_mount").val();
+        if (mount.charAt(0) != '/') {
```
hairmare: This makes the mount part of `url = 'N/A'` below unreachable.

airtime_mvc/public/js/airtime/preferences/streamsetting.js
```diff
@@ -111,34 +111,36 @@ function checkLiquidsoapStatus(){
...
     
     live_dj_actions.find("#reset").click(function(event){
     	event.preventDefault();
-        var port = $("#dj_harbor_input_port").val();
-        var mount = $("#dj_harbor_input_mount_point").val();
-        var url = "http://"+location.hostname+":"+port+"/"+mount;
+        var port = $("#show_source_port").val();
+        var mount = $("#show_source_mount").val();
+        if (mount.charAt(0) != '/') {
+            mount = ('/').concat(mount);
```
hairmare: This makes the mount part of `url = 'N/A'` below unreachable.

airtime_mvc/public/js/airtime/preferences/streamsetting.js
Robbt: I see what you are saying. I'm not exactly sure what the 'N/A' is supposed to be there for except for if the user deleted both port and mount from the form.

Robbt:
Yeah so the ZendForm validation doesn't appear to happen at all because it is passed via ajax to the streamSettingAction. What do you suggest ?

hairmare:
During my tests the missing validation made it way to easy to save a state that led to sad tape errors on the settings page. 

I have some ideas on how this might be fixed but need to read up on zf1 validation first. I'll try giving this a go over the weekend.

hairmare:
Allrighty then... half of the form was missing at the moment when the validation that decided if saving is ok was run. I moved all the form building to the top so the form (and its validators) get built before the the initial validation.

There was also some js stuff that just reloaded the page instead of doing the proper job of displaying the errors that were already being returned.

I think validation is now pretty much ok, I tried it with some stream settings and it looked like it was fine.

I want to take a closer look at the python parts of this PR before we merge it. I have the slight feeling that we might have missed something.

hairmare:
It kinda looks like `TODO: This function needs to be way shorter, and refactored :/ - MK` was done but the comment was never removed. At least when I revert your changes to the python parts locally it still seems to update `/etc/airtime/liquidsoap.cfg`. I'm currently testing if anything doesn't work w/o the python changes and haven't found any issues so far.

hairmare:
I tested all the things on the ui w/o the python changes and on a rather clean install and everything worked, so pushed a revert for the python changes. If it doesn't works after you `python python_apps/pypo/setup.py install` and restart the services I'll take another look at the matter.

Other than that I think this is pretty much where it needs to be to land it on master.

Robbt:
I think you are right in your assertion that the configuration was refactored to occur elsewhere. I noticed range validation didn't appear to  be working for the master port field. I think I removed it inadvertantly while trying to figure things out. I reinserted it and it rejected the value, displayed the error message and then refreshed the screen with the original value in the form. Better than what was happening before but still not ideal.

PR #91, opened 2017-03-16
basic podcast album override option in preferences

### LINKED ISSUE #89: Problem: There is no way to identify podcast tracks with ambiguous album metadata to the podcast they originated from
The current implementation of the album podcast album metadata insertion will only add the podcast name as the album for tracks where the id3 tag lacks any information for Album. This should work for most cases where the album id3 tag is already a unique and useful identifier but there are some podcasts that have album metadata that is not useful for identifying the podcast. In this case we need to provide an option to override album metadata even if it exists.

I personally think that overriding should be on by default but we could provide an option that disables it either on a podcast specific setting or a general sitewide setting that applies to every podcast. 

This is necessary for my station to produce smart blocks that target specific podcasts and would also be helpful in the automatic creation of smart blocks to coincide with podcasts. If we don't override the album metadata there is no way to link a track back to the podcast it originates from. For this reason alone I think it may make sense to just override the album metadata even if it exists. I know this is also the default behavior of iTunes which is by far the most widely used podcast software and also putting the podcast name in the album is the recommended naming convention as per the iTunes standards for id3 tags.

Robbt:
This adds an option to preferences that makes it so that downloaded podcasts will always have their album ID3 tag  overridden, even if it isn't empty. 

There needs to be some troubleshooting because it is always overriding even if this option is disabled. 

This is either due to the wrong boolean being passed via rabbitMQ to the celery task or the python code in the celery tasks.py not being able to interpret the bool correctly. I'll work to fix it but if someone else notices a possible issue feel free to point it out. 

This fixes #89

### DIFF
--- a/airtime_mvc/application/controllers/PreferenceController.php
+++ b/airtime_mvc/application/controllers/PreferenceController.php
@@ -45,6 +45,7 @@ class PreferenceController extends Zend_Controller_Action
                 Application_Model_Preference::SetDefaultCrossfadeDuration($values["stationDefaultCrossfadeDuration"]);
                 Application_Model_Preference::SetDefaultFadeIn($values["stationDefaultFadeIn"]);
                 Application_Model_Preference::SetDefaultFadeOut($values["stationDefaultFadeOut"]);
+                Application_Model_Preference::SetPodcastAlbumOverride($values["podcastAlbumOverride"]);
                 Application_Model_Preference::SetAllow3rdPartyApi($values["thirdPartyApi"]);
                 Application_Model_Preference::SetAllowedCorsUrls($values["allowedCorsUrls"]);
                 Application_Model_Preference::SetDefaultLocale($values["locale"]);
--- a/airtime_mvc/application/forms/GeneralPreferences.php
+++ b/airtime_mvc/application/forms/GeneralPreferences.php
@@ -100,6 +100,25 @@ class Application_Form_GeneralPreferences extends Zend_Form_SubForm
             'value' => $defaultFadeOut,
         ));
 
+
+        $podcast_album_override = new Zend_Form_Element_Radio('podcastAlbumOverride');
+        $podcast_album_override->setLabel(_('Podcast Album Override'));
+        $podcast_album_override->setDescription(_('Always replace downloaded podcast track album with podcast name.'));
+        $podcast_album_override->setMultiOptions(array(
+            _("Disabled"),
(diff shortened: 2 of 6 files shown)


hairmare:
Should it only be a global option though? I could see myself wanting to override this on a per feed level.

Robbt:
I think we could make it a per feed setting for the podcast and go under the podcast setting. It shouldn't be too hard to implement in that way instead. 

One reason I went this route because it was easier to implement without touching the database and adding another field and all of the getters and setters to the podcast class.

hairmare:
Let me know if you think this is ready after my changes and I'll merge it. I did the per podcast override config on a feature branch and will open the respective PR as soon as this is merged.

Robbt:
I tried the latest version and got this issue. I tried composer install. I'm trying destroying wiping the vbox and trying a completely fresh vagrant box.

2017-03-14T01:22:43+00:00 ERR (3): localhost [ThirdPartyCeleryService.php:33 - _executeTask()] - Invalid request: Unable to execute INSERT statement [INSERT INTO "third_party_track_references" ("id", "service") VALUES (:p0, :p1)] [wrapped: SQLSTATE[23502]: Not null violation: 7 ERROR:  null value in column "file_id" violates not-null constraint
DETAIL:  Failing row contains (2, podcast, null, null, null, null).]

2017-03-14T01:22:43+00:00 ERR (3): localhost [PodcastEpisodesController.php:122 - postAction()] - Unable to execute INSERT statement [INSERT INTO "third_party_track_references" ("id", "service") VALUES (:p0, :p1)] [wrapped: SQLSTATE[23502]: Not null violation: 7 ERROR:  null value in column "file_id" violates not-null constraint
DETAIL:  Failing row contains (2, podcast, null, null, null, null).]

Robbt:
Ok so on fresh install it worked and I was able to turn on and off the album override.

One issue I've noticed that I'm not sure is unique to this PR, is that the Podcast Episode view never shows that the track succesfully downloaded but instead continues to display the rotating icon. I am pretty sure this worked in my original saas-dev fork. I'll do more testing.

Robbt:
So the issue isn't related to this PR. So I think we can move forward with it but will need to figure out why the UI isn't being updated.

hairmare:
I figured the issue out, I'm merging this since the fix needs some db changes and I can lump them together with the per podcast override option I already have on a branch.

PR #95, opened 2017-03-18
per podcast album override

### LINKED ISSUE #92: Problem: Podcast Web UI doesn't confirm track download
Something in the podcast code changes we did made it so podcast downloads aren't confirmed back to the UI. Normally they should have a green check mark when they are done downloading but instead the ongoing download dialog continues to display.

This is probably an issue with the celery task callback.

hairmare:
This allows you to override the album name in a downloaded podcast at the per podcast level. The global option takes precedence to it so it can only activate overriding if that is disabled. I should probably change that to it only being used as the default for new podcasts. Let me know what you think.

It also has the fix for getting a green checkmark after downloads have processed.

screen shot 2017-03-18 at 11 41 19

Setting it results in

screen shot 2017-03-18 at 11 51 14

vs.

screen shot 2017-03-18 at 11 51 07

Fixed #92

### DIFF
--- a/airtime_mvc/application/controllers/downgrade_sql/airtime_3.0.0-alpha.1/downgrade.sql
+++ b/airtime_mvc/application/controllers/downgrade_sql/airtime_3.0.0-alpha.1/downgrade.sql
@@ -1,1 +1,3 @@
 ALTER TABLE imported_podcast DROP COLUMN IF EXISTS album_override;
+ALTER TABLE third_party_track_references ALTER COLUMN file_id DROP DEFAULT;
+ALTER TABLE third_party_track_references ALTER COLUMN file_id SET NOT NULL;
--- a/airtime_mvc/application/controllers/downgrade_sql/airtime_3.0.0-alpha.1/downgrade.sql
+++ b/airtime_mvc/application/controllers/downgrade_sql/airtime_3.0.0-alpha.1/downgrade.sql
@@ -1,0 +1,1 @@
+ALTER TABLE imported_podcast DROP COLUMN IF EXISTS album_override;
--- a/airtime_mvc/application/controllers/upgrade_sql/airtime_3.0.0-alpha.1/upgrade.sql
+++ b/airtime_mvc/application/controllers/upgrade_sql/airtime_3.0.0-alpha.1/upgrade.sql
@@ -1,0 +1,1 @@
+ALTER TABLE imported_podcast ADD COLUMN album_override boolean default 'f' NOT NULL;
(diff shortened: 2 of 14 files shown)


Robbt:
This is awesome. Only issue I noticed was the UI for the label was off in firefox. 
screenshot from 2017-03-18 10-16-28

It has to do with this CSS - if you set the width to 40% it stays in one line but they are no longer lined up with the field above.
.podcast-metadata-field {
    display: inline-block;
    width: 60%;
    min-width: 60%;
    max-width: 60%;

Need to run to the dentist, let me know if you want to try to fix this before we commit.

hairmare:
I'll give it a look in Firefox, should be an easy fix & I wasn't totally happy with the placement anyway.

hairmare:
I can't seem to reproduce this anymore (and the placement issue are also fine now).

screen shot 2017-03-18 at 17 20 07

Maybe we both had some stale CSS in our caches 😕

hairmare:
Found it, you have a larger screen than me :)

hairmare:
the label ui should be positioned properly with 5684689a1200fe4116fff97c1f31c593ba3e06c5.

Robbt:
It still looks off for my firefox but its fine in Chrome. In tweaking the CSS I couldn't find an easy fix that made sense. It appears to be a rendering bug from what I can tell as it isn't behaving as the CSS width directive indicates.

hairmare:
Is it off in all resolutions or just in fullscreen?

Robbt:
I just pushed it to but I guess it came in as a new branch. It was an issue with the no-float class on the input checkbox. I created a float-right class and changed it. But I guess I wasn't sure how to push my commit correctly. Feel free to look at that code or give me instructions on how to commit to this ?
Do I need to add your fork as a remote then push to it ?

hairmare:
Yes, you have to add my fork as a remote and to push to it. This is the "Allow edits from maintainers" feature.

hairmare:
I seem to have mindlessly mis-clicked the delete before pulling, could you push somewhere again?

Robbt:
Yeah I added your remote but I'm seemingly failing to understand the proper git push to commit to it.
I fail when I type git push radiorabe feature/per-podcast-album-override
it gives error: src refspec feature/per-podcast-album-override does not match any.

git push radiorabe:feature/per-podcast-album-override gives me this error 
warning: push.default is unset; its implicit value has changed in
Git 2.0 from 'matching' to 'simple'. To squelch this message
and maintain the traditional behavior, use:

hairmare:
Your push looks ok to me, did you `git fetch radiorabe` after adding the remote?

Robbt:
I think I figured it out I needed to do git push radiorabe mylocalbranchname:yourbranchname

Robbt:
Yeah, the issue was odd, the float:none for the input box was causing any subsequent lines to be screwed up for some reason. We should keep that in mind in case we run into this issue again when adding dialog.

Robbt:
Let me know if this works for you and then we can merge it.

hairmare:
It does, I couldn't reproduce the ui glitches again.

PR #102, opened 2017-03-20
Re-implement version check

hairmare:
This makes LibreTime check its version against github releases and lets the user know when to update. It uses the red exclamation point when there is a patch release or if LibreTime is more than one major release ahead. The orange icon is used when LibreTime is on a git install, a single major update is available, or a pre-release version is installed. The green update icon gets used to signify that a new minor release is available. Finally the green checkmark will be used when you are on a stable release.

It might display some slightly incorrect info while we only have pre-release packages but should work fine afterward. You'll have to run `composer install` after switching the branch since it installs some the version comparing code used.

The actual version check should only get executed once a week. Making this configurable  (ie. choosing daily, weekly monthly and deactivating the check from general preferences) should anyone have such a requirement.

### DIFF
--- a/airtime_mvc/application/configs/constants.php
+++ b/airtime_mvc/application/configs/constants.php
@@ -25,6 +25,7 @@ define('AIRTIME_TRANSIFEX_URL'      , 'http://libretime.org/translating/');
 define('SUPPORT_TICKET_URL'         , 'https://github.com/LibreTime/libretime/issues');
 define('UI_REVAMP_EMBED_URL'        , 'https://www.youtube.com/embed/nqpNnCKGluY');
 define('LIBRETIME_WHATS_NEW_URL'    , 'https://github.com/LibreTime/libretime/releases');
+define('LIBRETIME_UPDATE_FEED'      , 'https://github.com/LibreTime/libretime/releases.atom');
 define('LIBRETIME_EMAIL_FROM'       , 'noreply@libretime.org');
 
 define('LICENSE_VERSION'    , 'GNU AGPL v.3');
--- a/airtime_mvc/application/controllers/LocaleController.php
+++ b/airtime_mvc/application/controllers/LocaleController.php
@@ -39,9 +39,11 @@ class LocaleController extends Zend_Controller_Action
             //dashboard/versiontooltip.js
             "You are running the latest version" => _("You are running the latest version"),
             "New version available: " => _("New version available: "),
-            "This version will soon be obsolete." => _("This version will soon be obsolete."),
-            "This version is no longer supported." => _("This version is no longer supported."),
-            "Please upgrade to " => _("Please upgrade to "),
+            "You have a pre-release version of LibreTime intalled." => _("You have a pre-release version of LibreTime intalled."),
+            "A patch update for your LibreTime installation is available." => _("A patch update for your LibreTime installation is available."),
+            "A feature update for your LibreTime installation is available." => _("A feature update for your LibreTime installation is available."),
+            "A major update for your LibreTime installation is available." => _("A major update for your LibreTime installation is available."),
+            "Multiple major updates for LibreTime installation are available. Please upgrade as soon as possible." => _("Multiple major updates for LibreTime installation are available. Please upgrade as soon as possible."),
             //library/events/library_playlistbuilder.js
             "Add to current playlist" => _("Add to current playlist"),
             "Add to current smart block" => _("Add to current smart block"),
--- a/airtime_mvc/application/models/Preference.php
+++ b/airtime_mvc/application/models/Preference.php
@@ -879,8 +879,27 @@ class Application_Model_Preference
     public static function GetLatestVersion()
     {
         $config = Config::getConfig();
-        $latest = self::getValue("latest_version");
-        if ($latest == null || strlen($latest) == 0) {
+
+        $latest = json_decode(self::getValue('latest_version'));
+        $nextCheck = self::getValue('latest_version_nextcheck');
+        if ($latest && $nextCheck > time()) {
+            return $latest;
+        }
+
+        $rss = new SimplePie();
+        $rss->set_feed_url(array(LIBRETIME_UPDATE_FEED));
+        $rss->enable_cache(false);
+        $rss->init();
+        $rss->handle_content_type();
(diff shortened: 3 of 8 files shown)


Robbt:
This appears to work and make sense, hopefully all of our testers who update will realize the need to run composer install again. I'm going to merge the PR.

korotkov13:
@Robbt @hairmare 
i'm new in git and linux mostly.
I want to understand updating process. To update my test vps I need to:
`sudo git pull` it will update git folder.
And to update the applications i run `sudo composer install` in libretime folder.
Thats all? or I need to do something else? 
Sorry for the noob question, but i'm realy noob in it:(

hairmare:
@korotkov13 It depends on how you installed. If you used the bundled install script then you might also have to rerun parts of it.

Basically, the `git pull` will update the copy of LibreTime in you current folder, `composer install` makes sure that the contents of the `vendor/` folder are complete. What is missing is installing the code to the right place on the system. I.e. the installer takes care of calling `python python_apps/airtime_analyzer/setup.py install` among others. The right place in this context is `/usr/lib/python` for that and somewhere in `/var/www` or similar for the php code.

Since the install script already does a `composer install` when it is run pulling from git and re-running the installer should get you there. You can skip things like configuring psql and rabbitmq when you do the update install. I recommend you create a backup of /etc/airtime before you start, the installer does some nasty stuff to the folder and if you have a backup you can restore it instead of running the setup steps in the web-ui.

korotkov13:
@hairmare 
Unfortunately i had 
screen shot 2017-03-23 at 14 37 55
After Forced reinstall to defaults, had the same.
Will try to install on clean server.

hairmare:
Is there anything in `/var/log/airtime/zendphp.log` that might help me make the code more robust?

korotkov13:
Unfortunately did not think of looking at the logs. Now installing on clean vps.
I'll try to reproduce the situation and will look.

korotkov13:
reinstalled on clean VPS and got the same OOPS error.

```
2017-03-23T15:16:25+03:00 ERR (3): old.server.host [ErrorController.php:54 - errorAction()] - An internal application error has occurred.: exception 'ErrorException' with message 'Undefined offset: 1' in /usr/share/airtime/php/airtime_mvc/application/views/helpers/VersionNotify.php:33
Stack trace:
#0 /usr/share/airtime/php/airtime_mvc/application/views/helpers/VersionNotify.php(33): exception_error_handler(8, 'Undefined offse...', '/usr/share/airt...', 33, Array)
#1 [internal function]: Airtime_View_Helper_VersionNotify->versionNotify()
#2 /usr/share/airtime/php/vendor/zendframework/zendframework1/library/Zend/View/Abstract.php(350): call_user_func_array(Array, Array)
#3 /usr/share/airtime/php/airtime_mvc/application/layouts/scripts/layout.phtml(32): Zend_View_Abstract->__call('versionNotify', Array)
#4 /usr/share/airtime/php/airtime_mvc/application/layouts/scripts/layout.phtml(32): Zend_View->versionNotify()
#5 /usr/share/airtime/php/vendor/zendframework/zendframework1/library/Zend/View.php(157): include('/usr/share/airt...')
#6 /usr/share/airtime/php/vendor/zendframework/zendframework1/library/Zend/View/Abstract.php(888): Zend_View->_run('/usr/share/airt...')
#7 /usr/share/airtime/php/vendor/zendframework/zendframework1/library/Zend/Layout.php(796): Zend_View_Abstract->render('layout.phtml')
#8 /usr/share/airtime/php/vendor/zendframework/zendframework1/library/Zend/Layout/Controller/Plugin/Layout.php(143): Zend_Layout->render()
#9 /usr/share/airtime/php/vendor/zendframework/zendframework1/library/Zend/Controller/Plugin/Broker.php(333): Zend_Layout_Controller_Plugin_Layout->postDispatch(Object(Zend_Controller_Request_Http))
#10 /usr/share/airtime/php/vendor/zendframework/zendframework1/library/Zend/Controller/Front.php(965): Zend_Controller_Plugin_Broker->postDispatch(Object(Zend_Controller_Request_Http))
#11 /usr/share/airtime/php/vendor/zendframework/zendframework1/library/Zend/Application/Bootstrap/Bootstrap.php(105): Zend_Controller_Front->dispatch()
#12 /usr/share/airtime/php/vendor/zendframework/zendframework1/library/Zend/Application.php(384): Zend_Application_Bootstrap_Bootstrap->run()
#13 /usr/share/airtime/php/airtime_mvc/application/airtime-boot.php(84): Zend_Application->run()
#14 /usr/share/airtime/php/airtime_mvc/public/index.php(68): require_once('/usr/share/airt...')
#15 {main}
2017-03-23T15:16:38+03:00 INFO (6): old.server.host Checking autoplaylist poll

```

hairmare:
With an empty database?

korotkov13:
Yes. After clean reinstall.
Will try again. maybe something was wrong during install. but it's strange.

hairmare:
I'll do something to make the code more robust. It should do a bit more validation that what is in it anyway.

hairmare:
The part that fails has to do with the contents of your `VERSION` file. What is in it on your box?

korotkov13:
@hairmare 
01aef51

hairmare:
Looks ok, I'll need to dig a bit deeper.

hairmare:
The developer version is based on running `git rev-parse --short HEAD`. It looks like not all versions of git are equal and the short commit-ish as written has a variable length depending on the version. My code then checks for a git version by looking if the length matches :(

When I run `git rev-parse --short HEAD | wc -c` it returns `10`. What does yours return?

korotkov13:
8

hairmare:
*rolleyes*, I didn't know that `git rev-parse --short HEAD` is that fragile 😒 

You can override the VERSION as a workaround with `echo '3.0.0-alpha' > VERSION` and it should stop failing.

korotkov13:
reinstalling with -fiap not changed situation with "OOPS".
Will try to reinstall everything and changing VERSION file.

hairmare:
re-running the installer after changing the file should probably also work.

korotkov13:
already tried. not helped.

hairmare:
Robb just landed #115 so master should not show the sad tape anymore.

PR #106, opened 2017-03-20
Hideable Scheduled Shows tab

hairmare:
Makes the first tab remoeable and the tracks page be full width after removing the last tab. You can press the "Dashboard" link to open it back up again.

### DIFF
--- a/airtime_mvc/application/views/scripts/showbuilder/index.phtml
+++ b/airtime_mvc/application/views/scripts/showbuilder/index.phtml
@@ -23,9 +23,10 @@
 <div id="show_builder" class="sb-content content-pane wide-panel">
     <div class="panel-header">
         <ul class="nav nav-tabs">
-            <li id="schedule-tab" role="presentation" class="active">
+            <li id="schedule-tab" data-tab-id="0" role="presentation" class="active">
                 <a href="javascript:void(0)" class="tab-name" title="<?php echo _("Scheduled Shows") ?>">
                     <?php echo _("Scheduled Shows") ?>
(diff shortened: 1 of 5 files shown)


Robbt:
It works but I think that the UI could a little bit improved.  From what I saw there was only 3 dots to the right of scheduled shows. I think that it might be better to have an arrow or some other indicator. I also noticed that the area I could click on was somewhat limited.

hairmare:
It should be an X on the tab like the other tabs have. The three dots happens when the text doesn't fit. Maybe I missed some CSS. I'll have a closer look at it.

hairmare:
This is what it should like:

screen shot 2017-03-23 at 11 24 45

If the text is clipped a bit and showing dots your browser might still have a cached version of the css changed by this. I had to do a lot of <kbd>Cmd-Shift-R</kbd> reloading to hack on the css.

Robbt:
Ok, in the dashboard.css I needed to change the max-width to 160px vs. 140px to get it to show.
I figure this is probably something with Firefox vs. Chrome.

I can't do the push right now but here is the section that needs to change from my testing.

.tab-name {
    float: left;
    max-width: 160px;
    overflow: hidden;
    white-space: nowrap;
    text-overflow: ellipsis;
}

hairmare:
I'll give both a try and apply it.

hairmare:
I wasn't able to reproduce it, maybe my Firefox Aurora install is too hipster for it. I did the change anyway since a larger `max-width` shouldn't hurt anyway.

PR #109, opened 2017-03-21
Fixing dependence upon hard-coded password in AutoPlaylist

### LINKED ISSUE #103: Problem: Changing the admin's password causes the auto playlist function to fail.
I found bug. If I change admin's password, the autoplaylist stops working and refers to an access permission error.

PHP Fatal error: Call to a member function canSchedule () on a non-object in /usr/share/airtime/php/airtime_mvc/application/models/Scheduler.php on line 142 , referer: http://mysite.domain/schedule
Found a link to errors in Apache2 logs

zendpho.log shows

[Logging.php:166 - loggingShutdownCallback()] - Call to a member function canSchedule() on a non-object at /usr/share/airtime/php/airtime_mvc/application/models/Scheduler.php:142
array (
'type' => 1,
'message' => 'Call to a member function canSchedule() on a non-object',
'file' => '/usr/share/airtime/php/airtime_mvc/application/models/Scheduler.php',
'line' => 142,

After changing password back to admin, it's working again

created new admin user and it works

Robbt:
The add playlistoshow function had a parameter that wasn't being utilized to allow it to tell the validation to not check the user permissions. The passing of this parameter was added back to the addPlaylistToShow function and this removed the dependence upon initializing a user object with the hardcoded admin password. This fixes #103

### DIFF
--- a/airtime_mvc/application/common/AutoPlaylistManager.php
+++ b/airtime_mvc/application/common/AutoPlaylistManager.php
@@ -18,53 +18,26 @@ class AutoPlaylistManager {
         return empty($lastPolled) || (microtime(true) > $lastPolled + self::$_AUTOPLAYLIST_POLL_INTERVAL_SECONDS);
     }
 
-    /*
-     * This function is copied from the TestUser class and is used to instantiate a user so that
-     * the Scheduler model can be utilized by buildAutoPlaylist.
-     * Not sure if this is the best strategy but it works.
-    */    
-    public static function loginUser()
-    {
-        $authAdapter = Application_Model_Auth::getAuthAdapter();
-
-        //pass to the adapter the submitted username and password
-        $authAdapter->setIdentity('admin')
-                    ->setCredential('admin');
-
-        $auth = Zend_Auth::getInstance();
-        $result = $auth->authenticate($authAdapter);
-        if ($result->isValid()) {
-            //all info about this user from the login table omit only the password
-            $userInfo = $authAdapter->getResultRowObject(null, 'password');
-
-            //the default storage is a session with namespace Zend_Auth
-            $authStorage = $auth->getStorage();
-            $authStorage->write($userInfo);
-        }
-    }
-     
      /**
      * Find all shows with autoplaylists who have yet to have their playlists built and added to the schedule
      *
      */
     public static function buildAutoPlaylist() {
-        // Starting a session so that the User can be created
-	Zend_Session::start();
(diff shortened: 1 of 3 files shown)


airtime_mvc/application/common/AutoPlaylistManager.php
```diff
@@ -18,53 +18,26 @@ public static function hasAutoPlaylistPollIntervalPassed() {
...
      */
     public static function buildAutoPlaylist() {
-        // Starting a session so that the User can be created
-	Zend_Session::start();
-        static::loginUser();
-	Logging::info("Checking to run Auto Playlist");
-        $autoPlaylists = static::_upcomingAutoPlaylistShows();
-       foreach ($autoPlaylists as $autoplaylist) {
+            // Starting a session
+            Zend_Session::start();
```
hairmare: I don't think the session is needed at all now that you don't login in a dummy user.

airtime_mvc/application/common/AutoPlaylistManager.php
```diff
@@ -18,53 +18,26 @@ public static function hasAutoPlaylistPollIntervalPassed() {
...
             Logging::info("Scheduling $playlistid");
             // call the addPlaylist to show function and don't check for user permission to avoid call to non-existant user object
             $si->addPlaylistToShow($playlistid, false);
             $si->setAutoPlaylistBuilt(true);
-
             }
-        Application_Model_Preference::setAutoPlaylistPollLock(microtime(true));
-	Zend_Session::stop();
+            Application_Model_Preference::setAutoPlaylistPollLock(microtime(true));
+	        Zend_Session::stop();
```
hairmare: Same here, no need to stop it if it isn't started in the first place.

Robbt: My testing seems to indicate that the session is still necessary. When I comment out the start and stop session I get an authentication error again.

hairmare:
It looks like [this](https://github.com/Robbt/libretime/blob/934cad44b5c6e5ebd838a48d8b0674bc7913ae31/airtime_mvc/application/models/Scheduler.php#L43) is the only reason It needs a login. I'll give making that optional a go.

hairmare:
I figured out how to get rid of the session. Let me know if it still works for you and I'll merge it.

Robbt:
I tested it and it is working so I think we are ready to commit. Thanks for the help on the fix.

PR #110, opened 2017-03-21
ftp-upload-hook: Get rid of the SaaS stuff

HaJoHe:
I've just modifed the script, to not use anymore the SaaS config


hairmare:
Sweet, thank you very much!

Issue #42, opened 2017-03-06
Reinstate record from line in from Airtime 2.5.x


comiconomenclaturist:
A lot of the code seems to still exist, including recorder.py and some of the PHP functions.

hairmare:
I think [2.5.x:airtime_mvc/application/views/scripts/schedule/add-show-form.phtml#L33-L37](https://github.com/LibreTime/libretime/blob/archive/airtime/2.5.x/airtime_mvc/application/views/scripts/schedule/add-show-form.phtml#L33-L37) is all that's missing.

comiconomenclaturist:
Yup I found that bit of code for entering shows to be recorded, but I don't think the ShowRecorder function in recorder.py is being called. At least no ecasound process starts!

hairmare:
Ok, i removed "Fixes" annotation from the comment on #43. If this should close when we merge that, please reopen it if we haven't found the issue.

Do you know at what version the recorder last worked? It seems to still get started [in pypo](https://github.com/LibreTime/libretime/blob/276a69aa1bef527260efc170d7caab7d04b4ca13/python_apps/pypo/pypo/__main__.py#L298-L300).

hairmare:
I'm at the point where it calls ecasound on my vagrant box and then fails miserably. Most likely it is some issue with file permissions or a missing folder somewhere. Also, after adding a new recording i t takes up to an hour for the recorder to pick up on it so keep that in mind if you want to test this.

hairmare:
A quick update on this... After biting my teeth out on how to get the recording from ecasound uploaded to the API for inclusion in the library through analyzer (without loosing all the metadata concerning that it's a pre-recording). 

We should probably rewrite the code so that the file entry in the db gets pre-populated so we can pass its id to pypo-recorder when requesting the recording. pypo-recorder would then be able to just PUT an update to the file resource, that already contains the needed metadata. pypo-recorder would need to add proper id3 tags as well (I need to re-check if it does anything smart yet).

On the MVC side of things the models need to be adapted to store these file resources in a fashion similar to how podcast episodes and watched files are handled. It also makes sense to clean up the ui integration points a bit and keep the recordings nicely split from the main library in a fashion similar to other things like podcasts and streams.

Sorry about the braindump but I just realized that I need somewhere to put this. I'm currently focussing on a largish js refactor but I hope I'll find some time to look into this.

GroundTruthNZ:
[...]

PR #121, opened 2017-03-24
Fix recording through ecasound

hairmare:
This is part 2 of fixing ecasound recordings from line-in. Part 1 restored the User-Interface, part 2 takes care of getting to the point where ecasound gets started, records something and uploads it through rest when done. Part 3 will take care of making sure that the recorded file is mapped to the show and not just stored as a new track.

I refactored api_clients to not use urllib2 for posting multipart data since I was loosing my sanity over it and requests seems to have a modern approach to doing this compared to what api_clients was previously doing.

This is a large part of #42 but there is still some more re-implementing to do. I'm PRing this now since I already spent way too much time to get it to this stage.

### DIFF
--- a/airtime_mvc/application/configs/conf.php
+++ b/airtime_mvc/application/configs/conf.php
@@ -73,10 +73,6 @@ class Config {
 
         $CC_CONFIG['apiKey'] = array($values['general']['api_key']);
         
-        if (defined('APPLICATION_ENV') && APPLICATION_ENV == "development"){
-            $CC_CONFIG['apiKey'][] = "";
-        }
-
         $CC_CONFIG['soundcloud-connection-retries'] = $values['soundcloud']['connection_retries'];
         $CC_CONFIG['soundcloud-connection-wait'] = $values['soundcloud']['time_between_retries'];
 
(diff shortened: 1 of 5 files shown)


Robbt:
Just a quick question, is it possible for this to record shows that are broadcast via Show Source login and/or the scheduled shows ? Recording via line-in would require the automation system be located at the studio whereas our system is likely going to be located at our transmitter site to minimize potential outages from network connectivity.  I haven't spent enough time looking into this yet but the ability to record everything that is broadcast into a archive would be more useful than only via line-in.

hairmare:
It only records from line-in with ecasound. As such it only covers some very narrow use-cases. For instance you can use LibreTime for rebroadcasting if you hook up an output from your studio directly to libretime. I see re-implementing this as part of backwards compatibility for legacy upstream users.

I would like stream recording and recording for what pypo does as well.

Our archive at rabe is a different system that only uses libretime to get infos on the show schedule. We have separate boxes that do the recordings and grab the audio signal before too much on air processing is done. We're currently working on making those truly redundant.

PR #122, opened 2017-03-24
Restore datatables-library from server on login

### LINKED ISSUE #69: Problem: The UI doesn't retain customizations by user from prior logins.
Often times, navigating the track listing requires modifying the columns displayed and sort order. This is retained per session but whenever a user logs in again they need to set this up again. In practice this can be very time consuming and annoying, the alternative I'm proposing is for these settings to be stored per user and thus maintained between logins. This is the first step in improving the usability of the track UI.

Basically the column and sort order of the track UI will need to be saved somewhere in the database besides the session cookie.

hairmare:
When localStorage is empty we try getting the data from the server. This is a bit hacky since the ui uses a very old version of datatables.

Fixes #69

### DIFF
--- a/airtime_mvc/application/controllers/UsersettingsController.php
+++ b/airtime_mvc/application/controllers/UsersettingsController.php
@@ -47,7 +47,9 @@ class UsersettingsController extends Zend_Controller_Action
     {
         $data = Application_Model_Preference::getCurrentLibraryTableSetting();
         if (!is_null($data)) {
-            $this->view->settings = $data;
+            $this->_helper->json($data);
+        } else {
+            $this->_helper->json(false);
         }
     }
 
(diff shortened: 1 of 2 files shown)


Robbt:
Awesome, this works. One thing I noticed only when we did this was that the Playlists and Smart Block interfaces show columns for meta-data that we don't allow users to add for these sort of items, for instance Playlists don't have a Genre so if you add a genre column then this will always be empty. That led me to think, maybe we should add a Genre field to playlists and/or smart blocks as this could be a easy way for people to categorize them. Just a thought, perhaps I'll open a feature request for this. Awesome work on this PR, I think this will make it a lot easier for people working with the library to not do redundant customizations every time they log in.

PR #131, opened 2017-03-28
Do not bundle liquidsoap library

hairmare:
I think the code was bundled into legacy upstream before it was
readily available in liquidsoap proper. These snippets are now
maintained at https://github.com/savonet/liquidsoap/tree/master/scripts.

On all distros I checked these scripts are installed to /usr/bin/liquidsoap
by the corresponding package. Liquidsoap loads them from there by
default.

By not bundling this we will be ready to profit from new features and
bugfixes in liquidsoap next that is expected to land on opam soon. We'll
need to figure out how to get debian and others updated when that
happens.


bburton:
My guess is that SourceFabric built their own versions of Liquidsoap so they could be assured that certain packages required by Airtime were compiled in by default. In addition, this would ensure Airtime was tested against a known version of Liquidsoap.

However, going forward, Savonet is releasing via OPAM only. The issue is the effort to build packages for all the distros is a large enough distraction that they won't continue to maintain the distro packages for future releases. But, if someone steps up to maintain the distro packages, they would welcome that.

As Liquidsoap provides critical functionality to LibreTime, allowing any version of Liquidsoap to be used can result in issues with LibreTime not working as expected. 

Once LibreTime 3.0.0 is released, then new releases have to be backwards-compatible to some extent. The problem with relying on distro packages for Liquidsoap is they may be fine for the 3.0.0 release but could very well be a dead end when it comes to upgrading Liquidsoap to a newer version. So in the end, it may be preferable to support installing Liquidsoap via OPAM for the initial release.

However, when looking at the [OPAM install page](http://opam.ocaml.org/doc/Install.html), it seems the current versions of OPAM aren't available via the different distro's packaging systems. There is a pre-built Linux binary.

So, what's the best way to be able to install a current version of Liquidsoap for all the supported platforms?

Robbt:
It looks like packages are available for OPAM in most of the supported distros, the question is whether they will be able to install the latest version of liquidsoap. I think that it might be worth trying to install liquidsoap via OPAM and see if it works from the start. We are already using pip & composer so installing packages via OPAM isn't that different, especially if it provides us with the latest version of liquidsoap (and the latest version works).

hairmare:
I absolutely hate how we are using pip. In fact my CentOS install only uses pip for development and in production everything is nicely packaged as RPMs. Stuff like pip and opam makes installing stuff sooo much harder when you are on a machine that does not have a direct connection to the internet.

If I was even slightly a debian guy I would really consider taking over the maintainership. I'm assuming legacy upstream built their own liquidsoap packages and just they didn't get around to contributing them. If I where to search for the reasons behind this I would look for them in the same place as for why `saas-dev` existed and was not merged to master in the first place.

The version being installed on debian is so old that most of it (like fdk-aac) has long [broken](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=841871). In the long run I'd rather target latest and be able to report bugs against a current version instead of an old packaged but broken version. In most distros an admin can freeze the packages as needed should the latest not work.

I also feel that having to grok both pip and opam to know what you are installing is a strain and thus OS packages are a good thing. Even more so given the large parts of the user base that is not Linux native. I mean seriously, to install 1 piece of software you need to learn 2 additional package managers to what your os already has. 👀 🤦‍♂️ 

For the most part these are also the reasons I made a tarball `libretime-3.0.0-alpha.tar.gz` on the release. It already contains the php vendor parts and saves a potential user from having to learn another library management tool. It's also installable without getting your firewall guy to open up internet access on a machine that does not need it.

Robbt:
Perhaps we should bundle the pip packages in the tarball following these instructions - http://stackoverflow.com/questions/11091623/python-packages-offline-installation

hairmare:
This PR does not touch either pip nor the way liquidsoap gets installed. We can bundle pip packages should the need arise. At least on CentOS there were almost no packages that where not already available as an RPM.

Robbt:
Yeah I just need to test this on Ubuntu and Debian and will commit it. Are there any issues with debian packages being out of date for liquidsoap that might cause issues here ? I know there is the out of date version of silan.

hairmare:
I did a diff on the liq files installed by the package vs. the ones I am removing here and there does not seem to be much difference. I don't believe this causes any issues with outdated liquidsoap. The best case is that it will work better because the libs are always in sync with what was installed.

Old liquidsoap and old silan might have issues that I am not aware of. Those wouldn't be introduced by this though.

PR #136, opened 2017-03-31
FreeIPA Auth Adaptor for LibreTime

hairmare:
Allow delegating user authentication to FreeIPA rather than having it be checked against the database.

With this feature I've ported everything I need to switch to LibreTime in my infrastructure. I did a ubuntu re-install to check if I don't break anything and have also been testing this on CentOS against a real FreeIPA server.

I've already started cleaning up the [v3.0.0.alpha.1 Milestone](https://github.com/LibreTime/libretime/milestone/2) and plan on doing another release soonish (ie. over the next couple of days unless lots of new bugs show up).

### DIFF
--- a/airtime_mvc/application/auth/adapters/FreeIpa.php
+++ b/airtime_mvc/application/auth/adapters/FreeIpa.php
@@ -1,0 +1,117 @@
+<?php
+/**
+ * Auth adaptor for FreeIPA
+ */
+
+class LibreTime_Auth_Adaptor_FreeIpa implements Zend_Auth_Adapter_Interface {
+    /**
+     * @var string
+     */
+    private $username;
+    /**
+     * @var string
+     */
+    private $password;
+    /**
+     * @var Application_Model_User
+     */
+    private $user;
+
+    /**
+     * username from form
+     *
+     * @return self
+     */
+    function setIdentity($username) {
+        $this->username = $username;
+        return $this;
+    }
+
+    /**
+     * password from form
+     *
+     * This is ignored by FreeIPA but needs to get passed for completeness
+     *
+     * @return self
+     */
+    function setCredential($password) {
+        $this->password = $password;
+        return $this;
+    }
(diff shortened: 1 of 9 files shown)


Robbt:
Ok, so I tested this out and I got a WSOD because there was a new ldap index required in airtime.conf

The installation process will create this but I think we might want to figure out a way to automatically add this via an upgrade script so that when people upgrade to a new alpha they don't end up with broken installs.

It probably isn't a big deal at this point but I think it might become an issue when we have more users so we might as well think about it now and see if there are any easy solutions that come to mind.

Here's the error - Uncaught exception 'ErrorException' with message 'Undefined index: ldap' in /vagrant/airtime_mvc/application/configs/conf.php:101\nStack trace:\n#0 /vagrant/airtime_mvc/application/configs/conf.php(101):

hairmare:
Yeah, I know the error (and the point you raise is valid). I'm not quite sure on how to proceed with new such values in such cases. This is the same issue we were facing after I refactored the soundcloud stuff. Once we reach 3.0.0 this kind of changes should either be automated or warrant a 4.x release as per SemVer.

Robbt:
Well I can commit this if we don't have an easy fix for the modified configuration. I'm sure we could hack something together but I'm trying to figure out what the best practices in this regard are.

hairmare:
IMHO the best practices would be to use `Zend_Config_Ini` and not have a `Config` class that looks like it's mostly a case of the "not-invented-here" antipattern.

I added a check for the ldap section to the code only loads it if it's there. While it's not the nicest solution it does make updating without changing the config not lead to a WSOD.

Robbt:
Sounds good. It does seem like it would make sense to switch over to Zend_Config_Ini, but not sure how high a priority it is at this point.

hairmare:
Shouldn't be that much of a prio considering we also need so switch whats happening in python...

PR #139, opened 2017-04-01
Report on airtime-celery service in status

### LINKED ISSUE #138: Celery service is missing from /?config and /systemstatus
Problem: The airtime-celery service is in the setup but missing from /?config and /systemstatus.

Solution: Add checks for celery in those locations

hairmare:
Adds celery checks to /?config and /systemstatus for completeness' sake.

Fixes #138


Robbt:
I tried just merging this commit onto my debian install gave me an index unhappy tape - Undefined index: celery.
`
2017-04-01T16:02:38+00:00 ERR (3): localhost [ErrorController.php:27 - errorAction()] - #0 /vagrant/airtime_mvc/application/views/scripts/systemstatus/index.phtml(16): exception_error_handler(8, 'Undefined index...', '/vagrant/airtim...', 16, Array)
`
Then I did a complete reinstall and the setup config celery check appears to not be working for Debian.
If I run pgrep -f -u www-data airtime-celery I get nothing whereas if I run ps aux | grep 'airtime-celery' I get

```
celery   29180  8.9  5.5 120684 56448 ?        S    17:55   0:01 /usr/bin/python -m celery worker --config=celeryconfig --app=airtime-celery.tasks:celery --time-limit=300 --loglevel=INFO --concurrency=1 --logfile=/var/log/airtime/airtime-celery.log --pidfile=/var/run/celery/airtime-celery.pid --hostname=airtime-celery@jessie
celery   29187  0.0  4.9 119508 50416 ?        S    17:55   0:00 /usr/bin/python -m celery worker --config=celeryconfig --app=airtime-celery.tasks:celery --time-limit=300 --loglevel=INFO --concurrency=1 --logfile=/var/log/airtime/airtime-celery.log --pidfile=/var/run/celery/airtime-celery.pid --hostname=airtime-celery@jessie
vagrant  29215  0.0  0.2  30404  2228 pts/0    S+   17:55   0:00 grep airtime-celery

```

I think the issue is that debian runs celery as the celery user vs. www-data. 
I get the pid when I run it as pgrep -f -u celery airtime-celery so perhaps we need to change the checkCeleryService to not check for the -u www-data and it should work as intended.

hairmare:
#meh, I'll look into fixing this... Not that it's really critical.

Robbt:
Hmm I tried this on my production server and even though the pgrep -f -u celery airtime-celery command shows pids the status page is showing a fail on the check.

hairmare:
I'll need to retest this on a Ubuntu or Debian box so it really works.

hairmare:
It turns out that the apache user does not have the permissions to call `posix_kill()` on the celery process. The `posix_kill()` call is used to see if the process is running. I'll have to think about this some more to figure it out.

Robbt:
Why don't we just see if 'pgrep -f -u celery airtime-celery' returns a non-blank value ? I did some simple testing and when the airtime-celery service was running it returns 2 pids and if it is stopped it returns nothing. Not sure what the posix_kill is needed for in this case.

hairmare:
Sounds like a plan, I was considering rewriting it to leverage more of systemd at some point. On CentOS there is also SELinux to consider.

Robbt:
Yeah I am still resisting switching over fully to systemd. It also seems like there is no good easy to read overview material for adjusting to it, but I also don't think exclusively supporting Devuan OS and init.d is an option either.

Cysioland:
Curious, what does Celery do?

hairmare:
"Allow edits from maintainers." is active on the branch, I'm not sure what's up with that.

I never followed through with it since it won't work on my setup where celery is running on a dedicated vm. As far as I can remember I also got caught up in systemd and dbus while looking into ways to refactor the whole status page in a way that doesn't rely on the webserver needing its privileges escalated to the point where it can signal other processes.

airtime_mvc/build/airtime-setup/load.php
```diff
@@ -154,7 +154,7 @@ function checkLiquidsoapService() {
 function checkCeleryService() {
     exec("pgrep -f -u celery airtime-celery", $out, $status);
     if (array_key_exists(0, $out) && $status == 0) {
-        return posix_kill(rtrim($out[0]), 0);
+        return 1;
```
hairmare: Shouldn't this be `return true`? It looks like the `return $status == 0` below is returning a bool and the `1` here is likely getting coerced into a bool by the caller.

hairmare:
I'm perfectly fine with merging this as is, I do want to give it a closer look as it seems that the GitHub merge algo isn't showing just your changes but heaps more on top. I'll look into maybe rebasing/squashing some of it to get a more precise commit log and diff here.

airtime_mvc/build/airtime-setup/load.php
Robbt: Yeah it should probably be true because the 1 is just coerced into a bool.

Robbt:
Yeah, I'm not sure why all of the commits showed up, it probably has to do with me updating my base to match the current master and for some reason all of them being tagged in this merge. If it's something I did incorrectly let me know, I figured it was just weirdness with github but its very possible that I did something weird to cause this.

hairmare:
[This looks](https://github.com/LibreTime/libretime/compare/master...radiorabe:feature/celery-status?expand=1) like everything will be proper after the merge... I'm assuming the UI just isn't rendering 329d214b25802f85e7e3e2d23dbce086913ca104 in a transparent way.

PR #142, opened 2017-04-03
Remove hardcoded include of pervasives.liq

### LINKED ISSUE #140: Debian 8.7 - airtime-liquidsoap status / apache config error
Hey, i discovered libretime 2 days go ( so happy to found a fork of airtime :) ). But i have a problem to install it. 

I install it with `./install` like the docs. But after installation i got 2 different issue : 

**With digital ocean server (dply) :** 
That worked only the first time, i have retry it later ( new install of debian ) and i got a problem on the last page of settings. 

Every service are detected up except of airtime-liquidsoap. I retried to restart it manually ( and same for all others services ) but same issue. If i execute `service airtime-liquidsoap status` the service is up... 

capture du 2017-04-02 23-15-31

**With OVH VPS ( VPS 1 ) :** It's the same issue but sometimes i got the default page of apache. 

All my tests are try with a clear install of debian.

These servers are on debian 8.7 with the 3.16.39-1 kernel on KVM.

Thx to read my issue ;)

Robbt:
There was an attempt to include library/pervasives.liq which was causing errors after #131 was merged. I think that this library is automatically included so I removed the include and it appears to be working. Hopefully this fixes #140

### DIFF
--- a/python_apps/pypo/liquidsoap/ls_script.liq
+++ b/python_apps/pypo/liquidsoap/ls_script.liq
@@ -5,7 +5,6 @@ set("server.telnet", true)
 set("server.telnet.port", 1234)
 # set("init.daemon.pidfile.path", "/var/run/airtime/airtime-liquidsoap.pid")
 
-%include "library/pervasives.liq"
 
 #Dynamic source list
 #dyn_sources = ref []


Robbt:
Also to the hard-coded location of pervasive.liq in utils/airtime-test-stream.py which I think is causing that script to fail.

hairmare:
Sorry about that, I seem to have missed that. I'm currently reading through pervasives and the files it is including. It looks like nothing from pervasives was really used. There are some funcitons that might come in handy at a later point and we can re-add them then. I'll also do a PR for `utils/airtime-test-stream.py` which was news to me.

PR #143, opened 2017-04-04
Do not explicity load pervasives.liq

hairmare:
As per the liquidsoap docs the file is loaded per default anyhow http://savonet.sourceforge.net/doc-svn/script_loading.html.

I think the output.shoutcast might be the only place where we really use it though. Testing this against shoutcast would be nice, but I don't have one and am not sure how relevant it still is.

Sorry again for breaking liquidsoap with #131 😞

### DIFF
--- a/utils/airtime-test-stream.py
+++ b/utils/airtime-test-stream.py
@@ -98,7 +98,7 @@ try:
     if stream_type == "icecast":
         command = "%s 'output.icecast(%%vorbis, host = \"%s\", port = %s, user= \"%s\", password = \"%s\", mount=\"%s\", sine())'" % (liquidsoap_exe, host, port, user, password, mount)
     else:
-        command = "%s /usr/lib/airtime/pypo/bin/liquidsoap_scripts/library/pervasives.liq 'output.shoutcast(%%mp3, host=\"%s\", port = %s, user= \"%s\", password = \"%s\", sine())'" \
+        command = "%s 'output.shoutcast(%%mp3, host=\"%s\", port = %s, user= \"%s\", password = \"%s\", sine())'" \
         % (liquidsoap_exe, host, port, user, password)
 
     if not verbose:
--- a/utils/airtime-test-stream.py
+++ b/utils/airtime-test-stream.py
@@ -30,6 +30,8 @@ def printUsage():
 
 def find_liquidsoap_binary():
     """
+    With libretime 3.0 we are no longer depending upon airtime-liquidsoap but using the built in liquidsoap
+    rather than a bundled version of it. So this function no longer needs to be used.
     Starting with Airtime 2.0, we don't know the exact location of the Liquidsoap
     binary because it may have been installed through a debian package. Let's find
     the location of this binary.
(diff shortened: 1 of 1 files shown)


bburton:
The local pervasives.liq file is still loaded by [utils/airtime-test-stream.py](https://github.com/LibreTime/libretime/blob/master/utils/airtime-test-stream.py#L101).

bburton:
It's possible to confirm `output.shoutcast` is available via built-in pervasives by running the command: `liquidsoap -h output.shoutcast`. If it's available, online help will be displayed, otherwise, the results are, _Plugin not found!_

Robbt:
When I try to run it even with the modified command I get.
There was an error with your stream configuration. Please review your configuration and run this program again. Use the -h option for help

I looked into it and the issue is with calling airtime-liquidsoap which doesn't recognize the options etc. I tried it with just plain liquidsoap and it worked. I'll try to push my changes to your branch. We may want to test or remove all of the airtime-liquidsoap references in the tests.

hairmare:
Given a4244595e547bce049507afe9db3b715c76a2c92 I think this has been broken for quite a while. At least I don't remember seeing airtime-liquidsoap used as liquidsoap binary anywhere (it's just a hacky wrapper afaik). I'll do some additional cleanup on this and test it some more so we can get this merged.

hairmare:
I switched it up some more to check for `liquidsoap` on the path some more and listened to a test tone successfully.

bburton:
Although the assumption is to use the liquidsoap packaged with each distro, in some cases there may be some missing dependency such as Opus or AAC+ (fdkaac) which is going to require users with those requirements to build their own liquidsoap binary with the required dependencies. Since the LibreTime installer installs the distro liquidsoap, any custom built version has to be installed elsewhere such as /usr/local/bin resulting in two liquidsoap binaries in the PATH.

The problem with relying on the PATH is it differs depending on how a user has their shell configured and different still in the init scripts when the starting up as daemon. The result is there could be mysterious issues that are hard to diagnose if the wrong liquidsoap is started. The most reliable method would be to explicitly define the path to liquidsoap in /etc/airtime/airtime.conf then pypo and the init scripts would reference the specified path to ensure the correct liquidsoap is started.

This is why in some ways, retaining the airtime-liquidsoap wrapper is a good thing because it could ensure the correct liquidsoap is run when called by the init scripts instead of having to teach the init scripts how to parse /etc/airtime/airtime.conf (not that it's very difficult).

Robbt:
I haven't commited this PR yet because I think more research needs to be done into how airtime-liquidsoap works and whether we want to continue using it before we factor out the code.

One thing I noticed in my testing from the CLI is that if I ran sudo -u www-data airtime-liquidsoap the ps command would show airtime-liquidsoap running with the config file but if I try to pass the config file when running airtime-liquidsoap as a parameter I had the same error I was receiving while testing this.

I haven't spent the time really investigating how airtime-liquidsoap works and where it pulls the config file from, but that would be a next step if we want to keep it as a wrapper so that we can determine why it isn't working with the tests.

hairmare:
We certainly want to keep the wrapper as it is an important part of pypo and does [most of the config file magic](https://github.com/LibreTime/libretime/blob/f1d6fdea39ddee6a7b5e487d16f54cdc0a36fbc8/python_apps/pypo/liquidsoap/__main__.py). AFAIK The way this test script calls it by passing a liquidsoap snippet on the cli doesn't mesh with how the wrapper is implemented.

Robbt:
Ok, that makes sense, for the purpose of this test I do think that using the generic liquidsoap command should be sufficient.  

I was attempting to add the same find liquidsoap binary to the soundcard test but it needs some additional testing and my git push is failing for some reason so I will do that as a separate PR once I can test it on real hardware (as it doesn't want to work in the virtualbox due to lack of sound hardware I think).

hairmare:
Sound hardware should actually be there in the vm, at least I had it working in ubuntu over pulseaudio at one point. I need to do some real tests with a proper interface to be sure about if/how all of the audio hw stuff works.

bburton:
Yes, a generic _liquidsoap_ command must be called in _utils/airtime-test-soundcard.py_ as the _airtime-liquidsoap_ command is specifically limited to only running the ls_script.liq Pypo script and cannot be used to run arbitrary liquidsoap commands and options.

PR #149, opened 2017-04-09
Use json_encode if we need to return js and not json

### LINKED ISSUE #145: Debian 8.7 - Calender not Working
Hello Online the Calendar not working.
LibreTime - Calendar

hairmare:
Turns out the zf1 jsonHelper whas doing some weird stuff to the content-type header under the hood.

Fixes #145

### DIFF
--- a/airtime_mvc/application/controllers/LocaleController.php
+++ b/airtime_mvc/application/controllers/LocaleController.php
@@ -438,7 +438,7 @@ final class LocaleController extends Zend_Controller_Action
         );
         $this->view->layout()->disableLayout();
         $this->_helper->viewRenderer->setNoRender(true);
-        header("Content-type: text/javascript");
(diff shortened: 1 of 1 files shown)


Robbt:
Just doing a simple grep there is also a this->_helper->json->encodeJson in the ApiController.php file in the returnJsonOrJsonp function. Do we want to change that as well.
It's unclear whether the json change actually had an impact on the bug in #145 or if it was just an issue with CloudFlare. We can go ahead and make this change though if it is better form, but we should probably change both places if it matters.

hairmare:
As long as we return json it should be ok with `encodeJson()`. This really only concerns when the returned output is not being requested through xhr but rather added directly to the page.

Since CORS is now configurable e might be able to completely get rid of JSONP as it is rather legacy.

JohnnyC1951:
Interesting, this may have fixed one of my reported issues re weekly widget

PR #150, opened 2017-04-09
Install curl on all platforms

hairmare:
As per the https://github.com/LibreTime/libretime/issues/145#issuecomment-292744215 it looks like we always need to install this.

### DIFF
--- a/installer/lib/requirements-debian-wheezy.apt
+++ b/installer/lib/requirements-debian-wheezy.apt
@@ -34,6 +34,7 @@ patch
 
 icecast2
 
+curl
 php5-curl
 mpg123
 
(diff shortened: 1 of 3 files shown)


korotkov13:
Maybe this will help to someone who uses openVZ vps with only ipv4.
Before installing libretime I installing theese packages:
`curl git`
after i downloaded and installed latest erlang for RabbitMQ (without it i had a problems with rabbitMQ)

> wget https://packages.erlang-solutions.com/erlang/esl-erlang/FLAVOUR_1_general/esl-erlang_19.3-1~ubuntu~trusty_amd64.deb

after i creating and adding this `NODENAME=rabbit@localhost` in /etc/rabbitmq/rabbitmq-env.conf
and only after it installation finishes without any issues and works fine on ubuntu 14.04

hairmare:
@korotkov13 `git` should only be needed if you are not installing from a tarball so I expect that need to go away once we have a stable release.

JohnnyC1951:
Excellent!

PR #151, opened 2017-04-09
Alsa-ubuntu.deb Update

Lapotor:
In the name was outdata so i changed it


hairmare:
I'll give this a spin before merging it.

Issue #141, opened 2017-04-03
Preserve line endings when repository is cloned


bburton:
## Issue

When checking out the repository under Windows using the official Git version for Windows, it tries to be helpful by converting line endings to the Windows native CR+LF for all text files. Normally, this might not be an issue but when provisioning a VM with Vagrant, it breaks the install as Bash can't execute scripts with CR+LF line endings.

## Expected behavior

Git cloning under Windows preserves line endings of relevant scripts and files regardless of a user's local Git configuration so it's possible to provision a VM with Vagrant, install and run LibreTime in place without issue.

## Possible solution 1

Modify repository configuration:

`git config core.autocrlf input`

However, I'm not sure if this configuration change will persist so that it works when cloning remotely.

## Possible solution 2

Create a .gitattributes file in the root of the repository that is configured to preserve all line endings for any scripts or files required to provision a VM, install and execute LibreTime.  In this case, it appears each text file type that needs to preserve the LF line ending would have to be identified as "text eol=lf"

In this case, it can be helpful to identify every unique file extension in the repository. The following command line does this and also lists every file with no extension when run from the root folder of the Git workspace:

`git ls-tree -r HEAD --name-only | perl -ne 'print $1 if m/\.?([^.\/]+)$/' | sort -u`

Attached is a proposed .gitattributes file.
[gitattributes.txt](https://github.com/LibreTime/libretime/files/891680/gitattributes.txt)

## Possible solution 3

Don't make any changes other than to document the workaround below.

## References

[GitHub Help - Dealing with line endings](https://help.github.com/articles/dealing-with-line-endings/#platform-linux)

## Workaround

Before cloning the repository, change the global setting for `core.autocrlf` to `input`. This will preserve the line endings when files are checked out into the workspace after cloning. However, any new text files must be created with LF line endings as the CR+LF will not be converted to LF on commit:

`git config --global core.autocrlf input`

Then perform the clone.

hairmare:
I haven't had to use Windows in ages so I can't say if this isn't something special that every windows git user need to be aware of.

Solution 1 is only local to the repo and would need changing for each new clone. I feel the workaround is a bit much for someone who wants to get into LibreTime and it also kinda pollutes the global git config. 

As long as the gitattributes change doesn't have any side-effects on Unix platforms I'd be happy to merge a PR containing your documented fix for windows. Does it really need all the files to be listed? Shouldn't a double star glob (`**`) catch them all?

hairmare:
[...]

PR #153, opened 2017-04-09
Create .gitattributes to ensure proper cloning with Windows Git

bburton:
Address issue #141 so Git cloning under Windows preserves line endings of relevant scripts and files regardless of a user's local Git configuration so it's possible to provision a VM with Vagrant, install and run LibreTime in place without issue.


hairmare:
Thanks!

PR #157, opened 2017-04-11
Ubuntu 16.04 installer update

Lapotor:
I think this is a good step i updated the installer that Ubuntu 16.04 user can use it without any ~~extras except that the need to preinstall **Upstart**.
Which is very simple:~~
<s>
```
apt-get install upstart-sysv -y
update-initramfs -u
reboot
```
_[https://wiki.ubuntu.com/SystemdForUpstartUsers](https://wiki.ubuntu.com/SystemdForUpstartUsers)_
</s>
Than they easily can install it on Ubuntu Xenial.
If there is an error please report.
If you want to test this you can clone it:
```
git clone -b Ubuntu-16.04-Update https://github.com/Lapotor/libretime.git
```
I hope that will help a bit.

---
#### EDIT:
Updated the code so no need for upstart will work fine here.


install
```diff
@@ -645,10 +670,13 @@ else
...
 # Enable Apache modules
-if [ "$dist" != "centos" ]; then
+if [ "$dist" != "centos" ] && [ "$code" != "xenial" ]; then
     loudCmd "a2enmod rewrite php5"
+elif [ "$dist" = "ubuntu" ] && [ "$code" = "xenial" ]; then
+	loudCmd "a2enmod rewrite"
+	loudCmd "a2enmod php7.0"
 fi
-
+exit
```
hairmare: It looks like the code is bailing out on this line. I'm assuming this is some left over debug code.

install
```diff
@@ -676,7 +676,6 @@ elif [ "$dist" = "ubuntu" ] && [ "$code" = "xenial" ]; then
 	loudCmd "a2enmod rewrite"
 	loudCmd "a2enmod php7.0"
 fi
-exit
 loud "\n-----------------------------------------------------"
```
Lapotor: Ok sould be removed now.

hairmare: Thanks, I'll give it a try before merging...

Lapotor: Sure

hairmare:
I've been trying to figure out how to run this against the ubuntu-xenial vagrant box merged in #159 and haven't been able to figure out a sane way to automate the upstart install.

It looks like ubuntu switched to systemd so it might be simpler to just use that instead. I'll do some more testing and report back.

Lapotor:
Yeah will update it with an systemd cmd but I need to test it before

hairmare:
I'm pretty sure that the CentOS specific systemd code should just work with Ubuntu as well.

Lapotor:
will test

bburton:
In looking at the Systemd vs. Upstart issue, all of the most recent OS releases that LibreTime would support, - Debian Jessie, Ubuntu Xenial and CentOS 7 use Systemd out of the box. The maintainers of these distros have made the decision to move forward using Systemd so that's what LibreTime (or any other application) should support. 

Having said that, it's perfectly within the right of any sysadmin to revert to Upstart on systems he or she maintains. LibreTime could potentially support Upstart to some degree as an alternative (or not) on these systems but there are a number of implications to doing this such as what happens if LibreTime is installed and then the system is reverted to Upstart.

This brings me to the install script - after reviewing, it appears to make a number of assumptions and does not confirm many of those assumptions such as Systemd vs. Upstart and certain paths, etc. I suppose this is at least in part because it used to only support older versions of Debian and Ubuntu which are very similar but now the scope has increased quite a bit. More specifically, the install script should work a little more like configure and probe the system for key features then specifically test for said features instead of making so many assumptions based on the distro and release.

hairmare:
@bburton It used to support some older Debian and Ubuntu versions as well and already had some checks that didn't really make sense at the time either. Parts of it even some cleanup recently.

Sniffing the proper setup and doing the right thing is probably the most sane thing to do. I see refactoring needed on this script once we get proper deb and rpm packages built again. The [docs look like](http://libretime.org/manual/automated-installation/#install-airtime) there used to be an install script that was run by dpkg on debian. It might make sense to evolve this into something similar but with an install process more like [the easy-setup docs show](http://libretime.org/manual/easy-setup/).

That being said, I'll merge this PR when it's done just so we can get started with really testing xenial. I see some value in that. Once we have proper packages I'm hoping that the script installer will mostly be used for dev work as with the vagrant boxes or refactored as I outlined above.

bburton:
@hairmare Agreed. Have implemented OS and release detection along with systemd vs. Upstart detection but it should probably be implemented on top of or in conjunction with this branch as it could affect many of the tests.

Lapotor:
When i get my VM up again i will test out a bit with detection.
I think we should detect:
- Ubuntu Trusty
- Ubuntu Xenial
- Debian Wheezy
- Debian Jessie
- CentOS 7
Any other OS which should be detected ?

rallep71:
Perhaps still arch, but I think that will be too much effort, since is constantly changing, RR....

MfG LU

bburton:
About to submit a PR to support Wheezy in the Vagrantfile and based on the bento boxes for Debian.

Lapotor:
@bburton may you can help me to make the support for normal support on install will open a new PR for that ...

Lapotor:
The only error i noticed is that i can't upload files:
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;It saied uploaded but it isn't. May you can help there

hairmare:
I gave this a try against the bento box from #171 and I was able to upload files as expected.

Thanks for your contribution!

Issue #162, opened 2017-04-15
Stream label formatting only works for scheduled tracks, not for master or show sources


bburton:
## Background

In the web UI, the _Stream label_ setting can be configured in _Settings, Streams,_ to be one of three formatting options:
1. Artist - Title
2. Show - Artist - Title
3. Station name - Show name

## Issue

The _Stream label_ setting works as expected for scheduled tracks that are properly tagged with Artist and Title. However, the formatting is completely ignored under the following circumstances:
1. No scheduled tracks available or offline status - whatever is specified for the _Off Air Metadata_ setting is used as-is with no formatting.
2. Show DJ Source - title (if specified) used as-is with no formatting.
3. Master DJ Source - title (if specified) used as-is with no formatting.

## Expected Behavior

Given the _Off Air Metadata_ setting is configured as _LibreTime - offline_:
1. The _Artist - Title_ formatting method should show _LibreTime - offline_. This works as expected.
2. The _Show - Artist - Title_ formatting method should show: _Show name - LibreTime - offline_ but instead shows the same as 1 above.
3. The _Station name - Show name_ formatting method should show: _Station name - Show name_ but instead, shows the same as 1 above.

## Diagnosis

After reviewing the Pypo liquidsoap scripts ls_script.liq and ls_lib.liq, it's clear that the [append_title function defined in ls_lib.liq](https://github.com/LibreTime/libretime/blob/master/python_apps/pypo/liquidsoap/ls_lib.liq#L26-L43) implements the Stream label formatting in liquidsoap depending on what option was selected in the web UI. However, the call to this function takes place way too early in the stream configuration process in [ls_script.liq on line 111](https://github.com/LibreTime/libretime/blob/master/python_apps/pypo/liquidsoap/ls_script.liq#L111) so it can only be applied to scheduled tracks and not in any of the other scenarios.

## Proposed Fix

Move the call to _rewrite_metadata_ much farther down in the script to [about line 333](https://github.com/LibreTime/libretime/blob/master/python_apps/pypo/liquidsoap/ls_script.liq#L313). In addition, add checks for blank Artist and/or Title to ensure the Stream label formatting looks reasonable under all conditions.

I've been running a patch that partially addresses these issues for about two years. Working on some improvements and testing thoroughly before submitting a PR.

As a separate issue, the default Off Air Metadata should be changed from _Airtime - offline_ to _LibreTime - offline_.

zklosko:
[...]

PR #164, opened 2017-04-15
Problem: legacy upstream name is still in defaultdata

hairmare:
This shows up in places like *Off Air Metadata* as mentioned in #162 and others like the default stream setup.

Solution: Replace legacy upstream name with "LibreTime" in defaultdata.sql

I believe the file is not autogenerated so just changing it directly is the way to go.


bburton:
In [ls_script.liq line 182](https://github.com/LibreTime/libretime/blob/master/python_apps/pypo/liquidsoap/ls_script.liq#L182), it has "Airtime - offline" defined. Normally, this value should never be used as the value is pulled from the database and saved in liquidsoap.cfg each time the airtime-liquidsoap script is run. But it still should be updated. If you don't update it in this PR, I can as part of #162.

hairmare:
I updated it in here since it's an easy fix.

PR #201, opened 2017-05-16
Fix release script call

hairmare:
## Problem
The [script was logging](https://travis-ci.org/LibreTime/libretime/jobs/224026779) "Script failed with status 127". It seems that this is due to release.sh not being executable and not being referenced with a path starting with ./. As a workaround I built and uploaded the full release tarball for 3.0.0-alpha.1 manually.

## Solution
See https://github.com/travis-ci/travis-ci/issues/5538 for various ways on how to fix this. I hope committing the `chmod +x` attribute to git and adding `./` works.


hairmare:
👯‍♂️ Yay, this worked 👯 Releasing tarballs is now pretty easy 👍 

The "bundled-deps" release tarball was uploaded from travis pretty soon after I used the GitHub webinterface to tag the release 😄

Issue #215, opened 2017-06-02
Setup stalling on RabbitMQ


kernelpanic3:
Hi,

Trying to upgrade an Airtime 2.5 install to a freshly-pulled clone of 3.0.0-alpha1 (or whatever the current version is) on a server I own. I managed to get things up and running and get to the second step of the setup page, but when I hit Next it simply stalls with a waiting indicator that never goes away. I've confirmed that RabbitMQ is running where it expects it, and the passwords line up. Am I missing something or did I just find a bug?

Thanks for your work - I'd been following the (lack of) progress of Airtime, and I was very glad to hear about this fork tonight!

Robbt:
Hmm, what Linux distro are you using ?
Is there anything in the logs at /var/log/airtime/zend-php.log related to the failure ?

kernelpanic3:
I'm using Ubuntu 16.04, and that logfile doesn't even exist in that location...

Robby262:
Is Zend installed?

kernelpanic3:
Whoops, it actually wasn't (I never needed it for 2.5...). Installed `zend-framework` and `zend-framework-bin` from the repos. Still no dice though.

Robby262:
Zend Apache Module?

kernelpanic3:
I'll be perfectly honest, no clue there. Airtime 2.5 worked before, so I'd assume so...

Robby262:
If you can, I would start again with a fresh ubuntu trusty 14 install. It will all go a lot easier. ;)

kernelpanic3:
I really wish I could, but this system is already running a few other things...

Robby262:
LT doesn't seem to like PHP 7 much for a start. Backporting to 5.x is possible but then other stuff might stop.

Robby262:
Is this a dedicated server? or a small vps?

kernelpanic3:
A phpinfo() shows me that Apache is using 5.6...
![zend-php](https://cloud.githubusercontent.com/assets/7623828/26748654/83446ea8-47cd-11e7-9a99-d7c666a7ae1f.png)

Robby262:
Is it a vps or dedicated?

kernelpanic3:
> This is a physical machine I own myself, so dedicated for all intents and purposes.

Robby262:
Cool, excellent! So you can install a Virtual Machine system and run airtime in a Trusty bubble' :)

Robby262:
[...]

PR #218, opened 2017-06-05
Bump phpamqp

hairmare:
During composer install there are a couple of abandoned messages. This PR addresses the following one of them.

```
Package videlalvaro/php-amqplib is abandoned, you should avoid using it. Use php-amqplib/php-amqplib instead.
```
The change here is mostly in name and only contains some fixes concerning error handling and connection closing.

There is an upcoming 2.7 version of the lib that will need additional testing which is why I'd like to get this update into 3.0.0-alpha.2. It looks small enough and has worked on all distros I tried recently. The amqplib project still seems to support php 5.3 from the looks of it.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,7 +13,8 @@
         "massivescale/celery-php": "2.0.*@dev",
         "simplepie/simplepie": "dev-master",
         "zendframework/zendframework1": "^1.12",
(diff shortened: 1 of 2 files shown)


hairmare:
This might also help with #215

Issue #107, opened 2017-03-21
Contributing: Russian translation


yarik2720:
Hi! I'd like to contribute for Russian translation of Libretime

hairmare:
As you may have noticed we [don't have an interface](http://libretime.org/translating/)/#78 for that yet.

I figured out how to regenerate the base strings and could have something ready for use with [poedit](https://poedit.net/) in a reasonable timeframe. I'm assuming that you will not be the only russian translator so it probably makes more sense to prioritize zanata as per #78.

hairmare:
I'm updating the po files in #108 so they contain the new strings. automation and zanata to be done later...

hairmare:
I set up a zanata for LibreTime and tested it by round-tripping some de-DE strings. It may be found at https://translate.zanata.org/iteration/view/libretime/master.

It is based on the strings from #108. 

Merging back to github is not automated yet. So you'll need to let me know when you are done so I can pull the new strings from zanata. I'll try pulling them at somewhat regular intervals until this is automated but feel free to let me know if you want me to manually trigger a merge.

hairmare:
I looks like I need to know your zanata username to add you to the project. You can get one by going to https://translate.zanata.org/.

I'll set you up as the language admin if that's ok with you.

korotkov13:
@hairmare I got the same account as here

hairmare:
@korotkov13 You are now a Translation Maintainer, Russian Translator, and Reviewer.

This means you can also translate other languages and add translators as well. Thanks for contributing.

Let me know if you need any help or would like for me to pull the translations back into LibreTime.

yarik2720:
My username yarik2720

hairmare:
@yarik2720 I gave you the same permissions as Yuriy

korotkov13:
almost 79% ready)

korotkov13:
Most are translated and approved.
I will continue later if no one will help until then.

yarik2720:
99% ready. Some fuzzy strings remain

korotkov13:
@yarik2720 Great job!
I edited a little bit and approved.
I think we can close this thread and adding translation to master.

yarik2720:
Agreed

hairmare:
You will have to restart apache after updating the `.mo` files. That's not convenient but it's that way php-gettext works.

hairmare:
[...]

PR #251, opened 2017-07-19
Spanish missing translations and improvements

xabispacebiker:
Added missing translations and language improvements.


xabispacebiker:
There are some fixes that need to be applied before sending the pull request. Some strings are literally translated and when applied they break the interface. Those strings need to be replaced with shorter alternatives.
In the other hand, there are some PHP and JS files that need to be modified too, as they do not include the translation functions, therefor certain strings are not translated.

hairmare:
Yeah, translating can be hard given the ui constraints. We already did some tweaking of the design and added some missing strings for Russian (#107).

I'm sorry that I haven't updated the [docs](http://libretime.org/translating/) yet, but there is actually a zanata instance for translating LibreTime. It's at https://translate.zanata.org/iteration/view/libretime/master.

Anyway, once Spanish is ready we can either push pofiles from git to zanata or pull the other way around.

xabispacebiker:
> Anyway, once Spanish is ready we can either push pofiles from git to zanata or pull the other way around.

Great ! Because I did all the job already in a forked libretime instance 😅  .. I will have it ready soon, just need to accommodate those long strings and I will send the pull request.

Issue #233, opened 2017-06-14
Logs don't rotate


Robbt:
At least on Ubuntu 14.04 the various /var/log/airtime logs don't rotate and thus we end up with files that continue to grow rather than truncated .1 and .2 or zipped files. This could be an issue over time and is probably something we should resolve.

Robbt:
So the logrotate.d has entries for airtime-liquidsoap and airtime-php. And there is an entry in cron.daily for logrotate but I'm not sure this is working. It is possible that my production box that I upgraded just needed these entries set-up.

Also since we enabled logging for airtime-analyzer and airtime-playout we will also need to add logrotate.d entries for them.

paddatrapper:
Fixed in #987

PR #261, opened 2017-08-06
[WIP] adding and fixing log rotation

Robbt:
This is a fix in progress for #233 
The syntax for restarting the processes after postrotate needs to be fixed. 
the postrotate for the airtime-liquidsoap logrotate uses a pidfile that doesn't exist and I'm not sure the best way to restart the processes considering we want this to work in upstart and/or init.d and systemd.

### DIFF
--- a/python_apps/airtime_analyzer/airtime-analyzer.logrotate
+++ b/python_apps/airtime_analyzer/airtime-analyzer.logrotate
@@ -1,0 +1,11 @@
+/var/log/airtime/airtime_analyzer.log {
+  compress
+  rotate 10
+  size 1000k
+  missingok
+  notifempty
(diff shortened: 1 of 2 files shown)


hairmare:
Shouldn't `logging.handlers.RotatingFileHandler` from #260 already handle these? I think the existing `*.logrotate` files where there for services like liquidsoap that don't rotate on their own.

Robbt:
Yeah I'm not sure. I haven't really dealt with logging very much. So perhaps this is not needed.

Robbt:
Ok, so I'm going to close this for now since I think that the logging.handlers.RotatingFileHandler should do the trick for these files.

Issue #267, opened 2017-08-07
Feature Request: Document methods and classes using phpDocumentor and phpdoc comments


Robbt:
Since the majority of our codebase was written by people who are no longer active contributors I thought it might be helpful for us to document the various methods and classes so that they are easier to understand.

I looked around and found phpDocumentor as a way to create API documentation and I ran it on my repo and I'm creating a PR for it. It might be overkill as the actual documentation generated is 364MB in the /docs/api

To actually run it I had to do it outside of the vagrant VM because it has issues with file locking on NFS. 
After I ran composer install and composer update I then ran `./vendor/bin/phpdoc -d ./airtime_mvc/ -t ./docs/api/` to create the documentation. 

Some of the boilerplate Propel ORM classes code should probably be excluded and for it to be anything other than a twisted dump of classes and method listings we'd actually need to start using the protocol to write at least one liners and possibly descriptions for the various classes and methods.

Not sure if we want to go down this route but I decided to play around with it and figured it wouldn't hurt to explore the possibility as a way of rectifying some of the lack of knowledge about the existing codebase that I have.

PR #268, opened 2017-08-07
[WIP] - API docs w/ phpdoc at docs/api

Robbt:
This is a serious work in progress and currently consists of far too much boiler plate code dumps from phpdocumentor see issue #267 for discussion.


hairmare:
btw, we can treat the results of phpdocumentor as build artefact and upload them through the travis based libretime.org build chain (if you think they are worth it).

Robbt:
That sounds like a good idea.

I'm not sure how useful they are at this point. If we wanted to take the time to document all new features it might make them worthwhile eventually. We could also add documentation to existing features as we revisit and/or refactor them. Whether to use the tagging system and follow the format of the PhpDocumentor seems like a project decision that could go along with coding standards. 

While looking through the sourcefabric wiki I found the Php Coding standards document that we might want to import at some point if for no other reason that it forms the basis for how most of the code was in theory styled up to this point.

Robbt:
I'm going to close this for now as I don't think we are interested in investing the effort it would take to redocument the codebase and including these files makes manually searching through the code provide a lot of duplicate results that don't generally contain any more information.

PR #269, opened 2017-08-08
[WIP] Adding Block tests for smartblock refactoring

Robbt:
This is most likely the beginning of writing thorough tests for Blocks specifically focused on Smartblocks criteria etc, since I want to refactor them to include relative times I want to ensure that nothing else breaks in the mean time.


Robbt:
I'm going to merge these tests into #704 and add additional test cases specific to that PR to try and do a bit of test driven development. So I'm going to close this PR for now.

Issue #13, opened 2017-02-25
Make Smart Blocks capable of filtering by relative dates


Robbt:
In reference to this discussion of a popularly requested feature from the Airtime forum - https://forum.sourcefabric.org/discussion/16424/favourite-smart-block-criteria?post#Form_Body - 

It would be nice if LibreTime Smart Blocks were capable of filtering their criteria based upon relative dates vs. only specific date criteria. 

This would also help with the AutoPlaylist code.

korotkov13:
I want to offer one more option in smart blocks.
Add to sort tracks point: rarely played.
![screen shot 2017-03-21 at 20 22 57](https://cloud.githubusercontent.com/assets/25418594/24161151/adfbbf26-0e74-11e7-9277-26c74aa2889e.png)
I understand that you need statistics from the database for the playback of tracks and put a priority in the formation of a playlist in a smart block to those tracks that have played for a long time.

hairmare:
@korotkov13 I think you can partially achieve this with the "Last Played" criteria:

<img width="559" alt="screen shot 2017-03-21 at 19 27 21" src="https://cloud.githubusercontent.com/assets/116588/24163984/8b3bc0ec-0e6c-11e7-8a92-0e567cb199af.png">

korotkov13:
Oh my! Thanks!

Robbt:
So I'm looking into how to approach this and I thought I'd take notes here so I can check them later and anyone else can chime in.

Right now the smart block criteria is stored either as a string or a number and so it filters the date column to only be a numeric date this is in the SmartBlockCriteria.php file I believe. Whereas I filterByDbMtime in BaseCcBlockQuery.php might already take relative date strings (if properly formatted.)

Robbt:
Ok, so I think that if we just change the validation on this [form](https://github.com/LibreTime/libretime/blob/master/airtime_mvc/application/forms/SmartBlockCriteria.php#L568) to accept [PHP relative time formats](http://php.net/manual/en/datetime.formats.relative.php) formats in addition to actual dates it will work with the propel ORM. The question now is there a library or code already written to validate these I find it hard that there isn't something off the shelf but my research thus far hasn't found anything.

Robbt:
[...]

PR #274, opened 2017-08-17
Relative date filtering for smart blocks.

Robbt:
So I finally have a working prototype of this, but more needs to be done.

Basically it adds 3 modifiers to the 3 time based criteria, uploaded, modified and last played.
So you can select "before", "after", or "between" and these will pass the relative date field directly to the smartblock so that if you input "3 days ago", it will parse this using http://php.net/manual/en/datetime.formats.relative.php to 3 days prior to the time you are scheduling the smart block.

This is a work in progress because it currently does no validation and we discussed in issue #13 that it should be more international friendly possibly by using a dropdown menu to create the relative date.


korotkov13:
That's what I waited for!)
I tried to change these 4 files, and I got these new criterias. And I thikk that it's working.
I changed Last played - before and Playing with different dates (2-8 days ago) and duration is changing like it should. 
Will try it soon on working server. Thanks again! And it's very pity that i just saw this.

xabispacebiker:
Awesome !! +1

korotkov13:
I found one interesting thing.
I opened smart block after about few hours after adding "Last played" -3 days ago.
and now it looks like it only saving the date and time that was 3 days ago. and writes as a static date.
Maybe I'm wrong, and need more time to check it. But it looks little bit strange.

Robbt:
With regards to what it displays. I am not sure why it is translating it into a static date. If I remember correctly the database was actually storing it as the relative date text. Although I did see it displayed as a static date as well. I think we agreed when the issue #13 was first discussed that it would be more user friendly to international users to provide a drop-down selection box for the relative date stuff and then store it in ISO 8601 format. 

This will involve a fair bit of javascript and jquery mods to get working which I haven't gotten around to messing with it. I'd also think that being able to write a relative date might be easier for some people as well.

korotkov13:
I waited for this function so long, so It is a big win, that it's working this way.
Thanks again!

xabispacebiker:
If I am right, to display the drop-down selection we can follow the same coding as when Programming a **linked** show rebroadcast.

korotkov13:
I had one issue today.
I could't add any track to the empty show. Tried to add from smartcrates, and from playlists and even single tracks, it didn't added and showed this massage "The show has been previously updated!"

I opened zendphp.log and found this error. 
SQLSTATE[40001]: Serialization failure: 7 ERROR:  could not serialize access due to read/write dependencies among transactions
DETAIL:  Reason code: Canceled on identification as a pivot, during write.
HINT:  The transaction might succeed if retried.

After restarting server everything works.

[zendphp.log](https://github.com/LibreTime/libretime/files/1368389/zendphp.log)

hairmare:
@korotkov13 There is an issue with deadlocking somewhere, there are actually heaps of lockfile based guards against this kind of problem but it seem the actual reason for the deadlocks never got figured out/solved. Restarting apache works since it makes mod_php close all PostgreSQL connections, thus closing any unfinished transactions.

Robbt:
I started working on this again and realized why I walked away when trying to implement the drop_down select modification. 

First the jQuery code is messy, this was an obstacle I overcame.
Next basically the idea is if you select 'before, after or between' it shows the input box followed by a dropdown that allows you to select from minutes, days, months or years labeled Select Unit of Time and followed by the words ago.

This in theory will be translatable and thus easier to internationalize than requiring people to type in relative dates in English. The other benefit is that it is far easier to validate the data when you restrict what can be entered. 

The challenge and design decision comes in terms of how to save and parse this data in the database.
Do we modify the database to add an additional field to store the datetime_select value for both the main value and extra (in the case of range) or do we add a parser to the PHP code in SmartBlockCriteria.php that interprets the value and chooses the appropriate select.

I'm leaning towards the parser although it will probably be annoying to code and further complicate the already complex codebase because it won't unnecessarily complicate our database.

I think initially I'm still going to store the value in the database as parseable relative plain text as this seems easier to me than converting back and forth between ISO 8601. Although I may change my mind as I actually implement it. Anyone else feel free to chime in. 

The lazy part of me wants to just use text-based relative dates but I think the right decision is to guide the user in the right direction vs. requiring them to 'guess' or look up PHP documentation in order to choose a time.

Robbt:
Ok, I have done the majority of the work for this. It just needs testing. I've replaced the text-based input with a drop down when the relative date modifiers are selected. I just need to add some validation so people don't enter strings or whatever into the text field and also make sure that the smartblocks are being parsed correctly by LibreTime.

Robbt:
I think that this is ready for review at this point but at the same time there might still be some edge case bugs with regards to the jQuery form manipulation. I have tested it somewhat thoroughly and it appears to be working at this point without introducing additional errors. 

I should also probably update the documentation but the existing smart block documentation is so lacking that it probably could benefit from a more thorough write up in the first place.

Feedback and testing is welcomed. I don't know for sure that this will be compatible with any text input only based relative smartblocks unless you used the format "number" "time unit" "ago" which is what this stores in the database and parses. The time units are minutes, hours, days, weeks, months and years. All times are based upon the past and thus use ago to filter.

korotkov13:
Last ones is not working. 
Earlier versions of files works.

Robbt:
Hmm, ok, well the latest version should include an additional drop-down when you select Last Played that allows you to choose the unit of time, for instance in this case you would choose days from the drop down. Either there is an issue with the jquery not working, or you need to remove this specific "before" modifier and re-add it and it will display the drop-down for the unit time selection. The "ago" is automatically appended and not required.

Robbt:
Ok as I suspected there are still some issues with this in terms of when the relative unit dropdown shows up or disappears. Right now it appears to incorrectly display or hide when an error dialog is presented due to incorrect input. I'll be

Robbt:
The issue appears to be when reloading the smart-block-criteria.phtml after the $form-isValid check fails in the PlaylistController it doesn't keep the form attribute settings that were previously established. No jquery code is called to "fix" the display and for some reason this causes the datetime dropdown to reappear and/or hide w/o regard to what the dropdown values are. Not sure the easiest way to fix this. 

Is there a way to automatically trigger jquery code from a phtml template after display. That would most likely fix the form. Otherwise I'll be looking into why the display settings are changed after form validation.

Robbt:
Ok, I moved the logic as to whether to display a relative date drop-down to the view template rather than the model and this appears to fix the error mentioned above. I still want do a little bit more testing and remove any superfluous code left over in the model related to validation.

korotkov13:
I can test it but only in few days.

Robbt:
At this point there is at least one bug related to the drop-down handling and that is if you switch from before or after to between, it first drops the setting it had for unit of time selected and then if the form validation fails the 2nd "extra" unit of time select box is disabled and thus will cause a PHP error when you save it the second time. I will try to get this fixed before you get around to testing it.

In some ways the between for two relative times is confusing and I think it might make sense to validate it so that there is never a from value that is less than the to value because there will never be any matches. If anyone has any suggestions for how it could be more cognitively friendly let me know. 

Perhaps when between is selected it should prepend before "x" time and after "y" time vs. between "unit of time" ago to "unit of time" ago. But coding that with the spaghetti nightmware of jquery and php isn't something I'm really enthused about. I'm looking forward to the idea of rewriting the entire UI in something like vue.js

Robbt:
Essentially the remaining issues again have to do with an incongruence between the values in the database and those of the form.

Specifically the SmartBlockCriteria - startForm function uses this line of code
$modifierTest = (string)$storedCrit["crit"][$criteriaKeys[$i]][$j]["modifier"];
to determine if the user has selected before, after or between and thus shows and enables the appropriate drop-down.

This works when a form is pulled from the database, but if for instance a user previous had a between 5 weeks ago and 3 weeks ago selected and then deletes the row. The dropdown fields although hidden by jquery will still be considered enabled by the form validator. Which causes an issue when attempting to prevalidate the form.

This is also what causes for instance the form to cause the 2nd dropdown field to be disabled when switching from before to between and then not selecting both fields with appropriate data.

I was able to work around the previous bugs where the forms would show up by moving the validation of the display to the smart-block-criteria.phtml but that was only affecting whether they were displayed not whether they were enabled or disabled. And even that code was a little bit of a hack.

I'd like this feature to actually work well and be robust to bad input from users but I'm not sure how to accomplish that exactly with this particular implementation.

Robbt:
So what I'm leaning towards at this point due to the difficulty associated with validating the between and also how it doesn't seem intuitive conceptually is the complete removal of the "between" x and y for relative dates and simply including before or after. I think this will satisfy most conditions that people require and I'm not imagining a lot of need for the between functionality. 

Know that I'm thinking about it, it could also be more intuitively added by adding an additional modifier w/o needing to add "between". For instance if you wanted tracks that where last played before two weeks ago you could add that modifer and then an additional modifer if you wanted to also play tracks that were last played after 4 weeks ago. This would bring up tracks that were last played in the 2 week period from 4-2 weeks ago w/o needing to create a between filter.

I'm open to suggestions from other people here but that is the direction I'm thinking of heading in terms of tightening up this code and making it easier to validate etc.

Robbt:
Ok, I realized that my "fixes" to resolve the display of the datetime select by performing the logic in the template have unintended consequences on the form validation. So I need to do further research on Zend Form and how it is being utilized in conjunction with the jquery to come up with a fix for these issues.

Basically I need to revisit and thoroughly test the jquery display and removal of elements so there are no unintended consequences.

Also I need to determine why model for the form doesn't pull the actual form data after there is an error with validation. Instead it appears to contain what was saved in the database vs. what is displayed to the user. If I can figure this out my original code to display the appropriate "datetime select" etc, will work w/o modification.

Robbt:
Another bug I discovered that is present in the original codebase has to do with the removal of additional modifiers and left-over elements on the array that cause the validation function to crash.

This would only occur if you had two in the range modifiers and you deleted one and then tried to save it. The data element still contains the "extra" field and that alone in the form. 

This occurs more frequently for the datetime_select which either isn't disabled properly by the jquery delete (or I'm running into the same issue with the form not being the same).

I've pinpointed exactly where the discrepancy between the two forms is at least with my code but I'm not 100% sure how to fix it yet.

Basically startForm pulls the form data from the database. It enables or disables the form elements based upon the saved criteria. If you try to save it, the playlist controller saveAction runs startForm on the obj_id but startForm pulls the information from the database to determine what form elements to enable or disable. Next the form attempts to validate the $params which are the actual content of the form. If it fails it reloads the form using the data from the form but it determines which elements are enabled or disabled based upon the contents of the database. 

One idea I had was to create a new function called reValidate that would disable the date_dropdown_select based upon the current start form, but if you actually look at the way the form is built it is kind of a huge array of 28 different elements and additional ones added if you add multiple modifiers and it is no longer as easy to parse as the database. So it seems like asking for trouble.

This is probably better than the idea of enabling or disabling elements in the view template as that breaks the notion of model - vs. view and also didn't seem to actually disable the elements. Tricky stuff. Perhaps I'm just ignorant of the way Zend_Form is supposed to do this sort of thing or perhaps the original developers just did it a convoluted way out of sync with best practices. Either way I'm trying to figure out something here but it is becoming a massive time sink.

Robbt:
Ok, so I think I have figured this out. Using the preValidation function was the proper way to handle this. It enabled me to enable and disable elements.

I was also able to fix the issue related to deleting rows that I mentioned above.

There are still possibly a few cosmetic issues related to the jquery and I'd like to test and confirm that multiple crtieria work correctly together via adding additional modifiers etc as relates to the relative datetime.

But I think this is ready for testing and far better than it has been. It could probably be committed at this point if other people are able to confirm it works for them.

hairmare:
I gave this some testing and it seems to works for the cases I tested. I agree that it's ready for master.

I did some testing and it has some smallish merge conflicts with #370, it's only comment/whitespace stuff though. I guess I'll finish testing #370 and then push the resolved work to get both of the branches properly onto master.

korotkov13:
just tried to test, but it shows error 
I just replaced 3 files

and when tried to generate smart block with "last played - 2 days ago" it shows this error

Maybe I made something wrong?

Robbt:
There are a few more than that. It might be tricky to do everything on a file by file basis but at the same time the github UI also included all of the subsequent PR's. So here are the files that should be different from master before the PR was merged if you want to try to just replace certain files.

airtime_mvc/application/controllers/LocaleController.php
airtime_mvc/application/controllers/PlaylistController.php
airtime_mvc/application/forms/SmartBlockCriteria.php
airtime_mvc/application/models/Block.php
airtime_mvc/application/models/Show.php
airtime_mvc/application/models/ShowInstance.php
airtime_mvc/application/views/scripts/form/smart-block-criteria.phtml
airtime_mvc/application/views/scripts/playlist/update.phtml
airtime_mvc/public/css/styles.css
airtime_mvc/public/js/airtime/playlist/smart_blockbuilder.js

This PR required touching a lot of different parts of the UI and system models to implement. Let me know if replacing these files works.

korotkov13:
It's working at least now)
will check it longer and write here if something will go wrong.
Thanks again!

korotkov13:
I have one question.
Everything is working but only with songs.
If I upload new song, it's not adding in smartblock, because it's never been played.
Maybe If we add parameter, that will identify tracks that never been played.
example in "last played" column.

It can not generate with "is 0" or empty parameter. But I thought it would be great, if it worked.
PS: or maybe it is already working and I do not understand how?

Robbt:
Yeah you identified an issue with the logic, let's figure out the best way to fix it.

Issue #112, opened 2017-03-22
feature-request: add cloning function in smart blocks/playlists


korotkov13:
when I click on the Smart block, pops up only 2 options: preview and remove. It would be nice to add a feature copy/cloning. 
For example I have a smart block with certain parameters and duration 1-hour, I need the same block, but the duration of 30 minutes. Now to make it, I need to manually re-fill all parameters.

Robbt:
Great idea, this should make the creation of these blocks less tedious.

korotkov13:
I found in last release, that you added "Duplicate playlist" function in Playlists.
Can't wait to try it with smart blocks  :raised_hands:

Robbt:
This has been added in a PR. It doesn't copy the track contents of static smartblocks, only the criteria. Let me know if you think it would be useful to copy the tracks for static smartblocks as well.

Robbt:
This has been committed.

korotkov13:
just saw about it. will check it today in a few hours.
Thanks!!!

korotkov13:
@Robbt I upgraded updated version, but it is not work. 
It shows in context menu, but nothing happens when i click "duplicate"

korotkov13:
I will try to check it on clean system.

Robbt:
Ok, weird. Can you check the /var/log/airtime/zendphp.log and see if you notice any errors.

PR #275, opened 2017-08-17
Duplicate Smartblock functionality added to menu context

Robbt:
This adds the functionality requested in #112 so that there is a duplicate smartblock option.
If you click the 3 dot context menu for a smartblock you will see the duplicate smartblock option added.

It works and copies over the criteria and type but it doesn't copy the block track contents for pregenerated static smartblocks. Since the primary purpose will be for dynamic smartblocks duplicating the existing contents didn't seem important and new content can be easily generated.

### DIFF
--- a/airtime_mvc/application/controllers/LibraryController.php
+++ b/airtime_mvc/application/controllers/LibraryController.php
@@ -322,9 +322,27 @@ class LibraryController extends Zend_Controller_Action
         $newBl = new Application_Model_Block();
         $newBl->setCreator(Application_Model_User::getCurrentUser()->getId());
         $newBl->setDescription($originalBl->getDescription());
-
-        Logging::info($originalBl->getCriteria());
-        //$newBl->saveSmartBlockCriteria();
+        if ($originalBl->isStatic()) {
+            $newBl->saveType('static');
+        }
+        else {
+            $newBl->saveType('dynamic');
+        }
+        // the issue here is that the format that getCriteria provides is different from the format the saveCriteria
+        // expects due to the useage of startForm. So we either need to write new code that simply copies the database
+        // or figure out a way to instantiate a form inside of here and save it without modifying it.
+        //$newBlForm = new Application_Form_SmartBlockCriteria;
+        //$newBlForm->startForm($id);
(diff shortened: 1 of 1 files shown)


airtime_mvc/application/controllers/LibraryController.php
```diff
@@ -155,6 +156,7 @@ public function contextMenuAction()
                 $menu["duplicate"] = array("name" => _("Duplicate Playlist"), "icon" => "edit", "url" => $baseUrl."library/duplicate");
             } elseif ($type === 'block') {
                 $obj = new Application_Model_Block($id);
+                $menu["duplicate"] = array("name" => _("Duplicate Smartblock"), "icon" => "edit", "url" => $baseUrl."library/duplicate-block");
```
hairmare: Yay, more dead code revived 👍

airtime_mvc/application/controllers/LibraryController.php
```diff
@@ -280,6 +282,7 @@ public function deleteAction()
     public function duplicateAction(){
         $params = $this->getRequest()->getParams();
         $id = $params['id'];
+        Logging::info($params);
```
hairmare: Should this be `Logging::debug`? If `Logging::debug` is even a thing, I didn't check...

We need to clean up the logs anyhow so I won't consider this a blocker for this PR 😉

airtime_mvc/application/controllers/LibraryController.php
```diff
@@ -308,6 +311,42 @@ public function duplicateAction(){
...
+        if ($originalBl->isStatic()) {
+            $newBl->saveType('static');
+        }
+        else {
+            $newBl->saveType('dynamic');
+        }
+        // the issue here is that the format that getCriteria provides is different from the format the saveCriteria
+        // expects due to the useage of startForm. So we either need to write new code that simply copies the database
+        // or figure out a way to instantiate a form inside of here and save it without modifying it.
+        //$newBlForm = new Application_Form_SmartBlockCriteria;
```
hairmare: I think we would need to wrap `Application_Form_SmartBlockCriteria` in a proper form since it is based on `Zend_Form_SubForm` to be able to use the form.

PR #285, opened 2017-08-31
Hungarian translations

miles75:


hairmare:
Awesome! Thanks for contributing ❤️ 

I'll have to check that this also gets pushed to [zanata](http://zanata.org/) (writing docs for using zanata as main translation source has been on by list for what feels like ages, we used it with ru_RU already).

PR #308, opened 2017-09-27
Translation fixes + Spanish airtime.po

xabispacebiker:
Fixed some strings that were not translated
Fixed spanish airtime.po


hairmare:
Thanks!

Issue #307, opened 2017-09-26
[pypofile] could not copy from rest media to pypo cache (liquidsoap 1.3.1)


xabispacebiker:
As I explained earlier, I followed @hairmare instructions to integrate liquidsoap 1.3.1 into Libretime https://github.com/LibreTime/libretime/issues/192. 

OS: Debian 8 Jessie
Libretime 3.0.0-alpha.2 + Liquidsoap 1.3.1 implementation

Everything seem to work in a first place but then it started to stream silence, although with a scheduled show filled with content. 

**Edit 1:**
It's weird because some files play and some others don't (regardless the format).

**Edit 2:**
Apparently one file plays the next does not. Even when the file just played successfully a while ago: I create a new show. Add some wav, ogg, mp3 tracks. The show starts to play, after the first track the mp3 streams silence on second track, when it finishes with it, the stream starts playing the next track successfully and so on.. It's not 1:1 but more or less.

This is the pypo.log complaining about file not becoming ready and permission denied error:

2017-09-26 22:53:20,537 [pypoliquidsoap] [WARNI]  File /var/tmp/airtime/pypo/cache/scheduler/1760.mp3 did not become ready in less than 5 seconds. Skipping...
2017-09-26 22:53:20,537 [pypoliqqueue] [INFO ]  waiting 82.462166s until next scheduled item
2017-09-26 22:56:30,874 [pypofetch] [INFO ]  Queue timeout. Fetching schedule manually
2017-09-26 22:56:31,155 [pypofile] [INFO ]  copying from http:///:8080//rest/media/1761 to local cache /var/tmp/airtime/pypo/cache/scheduler/1761.mp3
2017-09-26 22:56:31,157 [pypoliquidsoap] [INFO ]  Need to add items to Liquidsoap *now*: set([637])
2017-09-26 22:56:31,158 [pypofile] [INFO ]  http://mydomain.com:8080/rest/media/1761/download
2017-09-26 22:56:31,159 [pypofile] [ERROR]  Could not copy from http://mydomain.com:8080//rest/media/1761 to /var/tmp/airtime/pypo/cache/scheduler/1761.mp3
2017-09-26 22:56:31,164 [pypofile] [ERROR]  [Errno 13] Permiso denegado: u'/var/tmp/airtime/pypo/cache/scheduler/1761.mp3'
[...]

PR #309, opened 2017-09-28
Pypo permission fix

xabispacebiker:
Fix for issue https://github.com/LibreTime/libretime/issues/307

Additionally corrected some related bugs

### DIFF
--- a/airtime_mvc/application/common/HTTPHelper.php
+++ b/airtime_mvc/application/common/HTTPHelper.php
@@ -37,6 +37,9 @@ class Application_Common_HTTPHelper
         if ($baseDir[0] != "/") {
             $baseDir = "/" . $baseDir;
         }
+        if ($baseDir[strlen($baseDir)-1] != "/") {
+            $baseDir = $baseDir . "/";
+        }
 
         $scheme = "http";
         if ($secured && !empty($_SERVER['HTTPS']) && $_SERVER['HTTPS'] !== 'off') {
--- a/airtime_mvc/application/common/HTTPHelper.php
+++ b/airtime_mvc/application/common/HTTPHelper.php
@@ -37,7 +37,7 @@ class Application_Common_HTTPHelper
         if ($baseDir[0] != "/") {
             $baseDir = "/" . $baseDir;
         }
-        if ($baseDir[strlen($baseDir)-1] != "/") {
+        if (substr($baseDir, -1) != "/") {
             $baseDir = $baseDir . "/";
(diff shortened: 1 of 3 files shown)


airtime_mvc/application/common/HTTPHelper.php
```diff
@@ -37,6 +37,9 @@ public static function getStationUrl($secured = true)
         if ($baseDir[0] != "/") {
             $baseDir = "/" . $baseDir;
         }
+        if ($baseDir[strlen($baseDir)-1] != "/") {
```
hairmare: You can use `substr($baseDir, -1)` to get the strings last char in a more straight forward fashion.

python_apps/pypo/pypo/pypofile.py
```diff
@@ -84,7 +84,7 @@ def copy_file(self, media_item):
                         handle.write(chunk)
 
                 #make file world readable
-                os.chmod(dst, stat.S_IRUSR | stat.S_IRGRP | stat.S_IROTH)
+                os.chmod(dst, stat.S_IRUSR | stat.S_IWUSR | stat.S_IRGRP | stat.S_IROTH)
```
hairmare: Awesome catch!

airtime_mvc/application/common/HTTPHelper.php
xabispacebiker: Great, I updated the code.. I guess not but.. should I send the request again?

hairmare:
Thanks for the patch!

xabispacebiker:
@hairmare , sorry but I checked the merge and realized there are missing changes which are very important. I am running right now with this pypo file and it seems to resolve all my issues with liquidsoap + it does not produce so many logs as it did before, which is a good point too. 

Mostly it was happening because I run libretime over the port https at 443.

I guess it was a fault from my side, as I made my head a mess with netbean's github thing... anyway, I have updated the full pypofile.py with the corrected version: https://github.com/xabispacebiker/libretime/blob/pypo-permission-fix/python_apps/pypo/pypo/pypofile.py

My apologize

hairmare:
[These](https://github.com/LibreTime/libretime/compare/master...xabispacebiker:pypo-permission-fix) seem to make sense to me. There are some small whitespace issues that might make sense to clean up (not all line are aligned properly on the left side). Could you address those and then make a new PR from the branch? I'd like to get these changes merged before I tag 3.0.0-alpha.3.

PR #324, opened 2017-10-07
Add a script to reprocess silan on all files. 

jmtrivial:
This script is useful after an update of the buggy silan:
https://forum.sourcefabric.org/discussion/17273/bugfix-silan-packages-for-ubuntu-14-04-lts-039trusty039-now-available

### DIFF
--- a/utils/airtime-re-silan
+++ b/utils/airtime-re-silan
@@ -1,0 +1,75 @@
+#!/usr/bin/python
+
+
+# usage: sudo -u postgres ./airtime-re-silan
+
+
+import os
+import sys
+import psycopg2
+import subprocess
+import json
+import datetime
+
+# loading config file
+srv_root = "/srv/airtime/stor/"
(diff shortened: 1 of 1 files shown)


hairmare:
Isn't this very similar to this one:

https://github.com/LibreTime/libretime/blob/44116bcc93b1fcbed1971a29df68ca9791748362/utils/airtime-silan#L51-L66

Are you sure it doesn't duplicate any features? The one already in utils/airtime-silan seems to be based off of the forum post with added config reading and it uses the api.

Maybe we should add an additional flag to force it to redo all the files rather than using the dedicated api endpoint "get_files_without_silan_value" 

https://github.com/LibreTime/libretime/blob/44116bcc93b1fcbed1971a29df68ca9791748362/utils/airtime-silan#L43

I haven't tested the script from the utils folder so I'm not sure if it works at all.

PR #334, opened 2017-10-15
Revert pypofile try block move

hairmare:
Config file getting is allowed to fail. I didn't realize that it would not just use the default instead of excepting and getting caught. Sorry for that. 

I broke this with my changes in #329 this reverts them and blocks 2.0.0-alpha.3. Sorry again about that.

---
### Workaround

Add `protocol = http` or `protocol = https` to the general section of your airtime.conf.

### DIFF
--- a/python_apps/pypo/pypo/pypofile.py
+++ b/python_apps/pypo/pypo/pypofile.py
@@ -61,11 +61,11 @@ class PypoFile(Thread):
             self.logger.info("copying from %s to local cache %s" % (src, dst))
 
             CONFIG_SECTION = "general"
+            username = self._config.get(CONFIG_SECTION, 'api_key')
+            baseurl = self._config.get(CONFIG_SECTION, 'base_url')
+            port = self._config.get(CONFIG_SECTION, 'base_port', 80)
+            protocol = self._config.get(CONFIG_SECTION, 'protocol', str(("http", "https")[int(port) == 443]))
             try:
-                username = self._config.get(CONFIG_SECTION, 'api_key')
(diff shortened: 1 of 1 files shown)


hairmare:
Refactored port and protocol defaults to be bare `ConfigParser` compatible in ace7e493f4346d5e6c17fc9f4311f88086daf819. De-WIP-ing again.

xabispacebiker:
Nice one! then.. I do not have to update the code I submitted..right?

hairmare:
Yupp, this contains the fix for when the config file doesn't have the protocol option.

Robbt:
So will this require an protocol option in the config file to enable https from now on ? If so have we added this to the documentation yet and should we alter the install script in some way to enable this configuration option during setup ?

hairmare:
Master requires the `protocol` block since #329. This makes it optional again by falling back to the previous behavior if the option doesn't exist.

Robbt:
I'm testing this right now. I didn't test the previous one and just merged it after code review. I figure this does work. I understand the change makes it so this works without a config file change, but my question is whether we should add something to the documentation or installer to make the config file change for people who want to use SSL.

hairmare:
If you use port 443 it should basically do the right thing without the need to specify the protocol.

I plan on adding complete docs on the whole TLS situation once we've reached the state where we can be confident that LibreTime actually works with TLS-only. I think we are getting closer to this state as this is the last issue I know about that really blocks making TLS work for all cases. 

I'd like to do some more testing since there may still be bugs lurking.

PR #351, opened 2017-11-10
Backup script

celianvdb:
Just have to set config, chmod +x and it works ;)


Robbt:
Ok, I don't see any problems with this, it is a useful contribution.

 It would be even more useful if there was a page on the documentation for it. 

Since it isn't 100% fool automatic. You have to manually modify the backup directory and it relies upon the default values for sound folder and the database. 

But considering how much revision and updating our docs need in general it might not very noticeable.

celiandu61:
I can auto, lot of things, you right, i will do that :) But do you know where the path of the media folder is stored ? I didn't found it in /etc/airtime/airtime.conf :thinking:

celiandu61:
Doc coming tomorrow ;)

utils/libretime-backup.sh
```diff
@@ -1,9 +1,21 @@
...
+        else
+                ## User arg as config
+                backup_folder=$1
+fi
+
+
+airtime_conf_path=/etc/airtime/airtime.conf
+sound_folder=/srv/airtime/stor/
+
+psdl_db=$(grep dbname ${airtime_conf_path} | awk '{print $3;}' )
```
hairmare: You could also do `$(awk '/dbname/ {print $3}')` instead of piping grep to awk. awk is very much regex based under the hood anyway. This works just as fine though.

celiandu61: Can you give me a complete example ? I don't understand 🤔

hairmare:
```bash
psdl_db=$(awk '/dbname/ {print $3}' ${airtime_conf_path})
```

Robbt:
So this has been sitting as a PR for quite some time and @celianvdb doesn't appear to be modifying the script as per the code review. So I'm trying to clean up the PR. 

While the code might not be perfect if it works and is basically a CLI script that only advanced users will use I would suggest we add it to the utils folder and merge it even if it isn't ideal because if it works for people and/or they can learn from it then its probably going to help more than hurt.

In reviewing the current util folder there are probably already a lot of scripts in there that might not work, so we should add a README disclaimer or go through and test and remove old scripts that no longer apply to libretime such as the airtime-nuke.sh & phone-home-stat. Actually I will open an issue for this.

Robbt:
So I'm going to merge this as is, and then we can suggest improvements and/or add documentation. Overall I think it is a handy utility script and until we come up with a good standard setup of style and documentation guidelines for util contributions I think this script works reasonably well.

PR #354, opened 2017-11-13
Added debian-stretch to vagrant setup

### LINKED ISSUE #353: Add Ubuntu-Stretch To vagrant setup
Our vagrant setup isn't configured to support debian stretch yet. Considering this is the current release for Debian then we should add this configuration to our options.

Robbt:
This fixes #353 and shouldn't cause any issues. Other than adding debian-stretch, it also changes the default suggestion to use debian-stretch vs. debian wheezy when someone types vagrant up debian.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -56,12 +56,16 @@ Vagrant.configure("2") do |config|
     os.vm.box = "bento/debian-8.7"
     provision_libretime(os, "debian.sh", installer_args)
   end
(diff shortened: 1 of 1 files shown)


hairmare:
Awesome. It's too bad that the bento boxes don't support KVM out of the box, that would make life for folks that switched to Fedora as their main desktop as i recently did way easier (VirtualBox ist just too slow on most platforms).

Issue #288, opened 2017-09-06
RabbitMq settings are not saved if you return to the form and statuspage crashes


Robbt:
I was performing a LibreTime install. Clicked through the rabbitmq page and then went back and changed the password. The new password was not saved by the installer and the systemstatus page crashes with the following error.

```
2017-09-06T16:11:23+00:00 ERR (3): x.x.x.x [ErrorController.php:54 - errorAction()] - An internal application error has occurred.: exception 'PhpAmqpLib\Exception\AMQPProtocolConnectionException' with message 'ACCESS_REFUSED - Login was refused using authentication mechanism AMQPLAIN. For details see the broker logfile.' in /usr/share/airtime/php/vendor/php-amqplib/php-amqplib/PhpAmqpLib/Connection/AbstractConnection.php:661
Stack trace:
#0 [internal function]: PhpAmqpLib\Connection\AbstractConnection->connection_close(Object(PhpAmqpLib\Wire\AMQPReader))
#1 /usr/share/airtime/php/vendor/php-amqplib/php-amqplib/PhpAmqpLib/Channel/AbstractChannel.php(199): call_user_func(Array, Object(PhpAmqpLib\Wire\AMQPReader))
#2 /usr/share/airtime/php/vendor/php-amqplib/php-amqplib/PhpAmqpLib/Channel/AbstractChannel.php(341): PhpAmqpLib\Channel\AbstractChannel->dispatch('10,50', '??oACCESS_REFUS...', NULL)
#3 /usr/share/airtime/php/vendor/php-amqplib/php-amqplib/PhpAmqpLib/Connection/AbstractConnection.php(213): PhpAmqpLib\Channel\AbstractChannel->wait(Array)
#4 /usr/share/airtime/php/vendor/php-amqplib/php-amqplib/PhpAmqpLib/Connection/AbstractConnection.php(180): PhpAmqpLib\Connection\AbstractConnection->connect()
#5 /usr/share/airtime/php/vendor/php-amqplib/php-amqplib/PhpAmqpLib/Connection/AMQPStreamConnection.php(60): PhpAmqpLib\Connection\AbstractConnection->__construct('airtime', 'airtime', '/airtime', false, 'AMQPLAIN', NULL, 'en_US', Object(PhpAmqpLib\Wire\IO\StreamIO), 0)
[...]

PR #355, opened 2017-11-13
Made rabbitmq settings write to tempfile upon check to fix install bug

### LINKED ISSUE #330: Web installer doesn't update RabbitMQ and postgresql custom passwords in airtime.conf
Upon initial install and deleting the /etc/airtime/airtime.conf, the web installer does not update the RabbitMQ password in /etc/airtime/airtime.conf.

Steps to duplicate:  On step 2 of web installer:
If initial install, run the cli command then...
`sudo rabbitmqctl change_password <username> <newpassword>`
If initial or after deleting airtime.conf for existing instance:
Then select advanced in step 2/5 on gui and enter new password.
Complete gui setup and the page will go to: 
http://yourpage.domain/?config
the page will not display.

Looking at airtime.conf, the default password is still airtime.
```
[rabbitmq]
host = 127.0.0.1
port = 5672
user = airtime
password = airtime
vhost = /airtime
```
Upon changing the password= value to what I set above in cli, ?config works and services look good.

I assume the issue is between the gui and updating the airtime.conf file since as soon as you make the change manually on airtime.conf things work.

For sure this applies to Ubuntu 16.04.  Not sure if it applies to other OS and not sure if other non-default parameters are applying in this or other pages.

PS: Love this app and lots of thanks to the hours people have put into this!

Robbt:
This fixes #330 and #288 by having the rabbit-mq setup function write to the temp file when it checks the rabbitMQ settings.

### DIFF
--- a/airtime_mvc/build/airtime-setup/forms/rabbitmq-settings.php
+++ b/airtime_mvc/build/airtime-setup/forms/rabbitmq-settings.php
@@ -10,7 +10,7 @@
     <h3 class="form-title">RabbitMQ Settings</h3>
     <span id="helpBlock" class="help-block help-message"></span>
     <p>
-        RabbitMQ is an AMQP-based messaging system used by Airtime. You should only edit these settings
+        RabbitMQ is an AMQP-based messaging system used by Libretime. You should only edit these settings
         if you have changed the defaults since running the installer, or if you've opted to install RabbitMQ manually.
     </p>
     <p>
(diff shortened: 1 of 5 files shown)


hairmare:
Sweet, I've stared at the setup code without finding this solution way too long.

Robbt:
Yeah, the issue was actually a typo on the database side of things. A variable was supposed to be $sections was $settings, at least for the db part of it.

Issue #356, opened 2017-11-13
Database password dependent upon hardcoded value


Robbt:
Ok, so in trying to fix #330 I came to the realization that our application is dependent upon database value set in "airtime-mvc/application/configs/airtime-conf-production.php". This value doesn't appear to come from the config setup and I was only able to get my application to work with a custom postgresql password when I overrode the database value on this page.

It came with this helpful comment 
`  2 // This file generated by Propel 1.5.2 convert-conf target /* THIS FILE IS NOT MEANT FOR 
   3 // from XML runtime conf file /home/james/src/airtime/airtime_mvc/build/runtime-conf.xml        PLEASE EDIT THE FOLLOWING TO CHANGE YOUR CONFIG:                                                  
  4* LIBRETIME_CONF_DIR/airtime.conf                                                                   
  5 /* The original name of this file is airtime-conf.php but since we need to make custom changes    │ */                                                                                   
6  * to it I've renamed it so that our changes aren't removed everytime we regenerate a database schema. require_once __DIR__ . '/constants.php';                                                             
  7  * our custom changes requires the database parameters to be loaded from /etc/airtime/airtime.con so class Config {
  8  * that the user can customize these.

Robbt:
So there is also a config/airtime-conf.php which appears to be generated by the install program but it  contains hardcoded airtime passwords rather than those entered during the web install.

` Propel::init(CONFIG_PATH . 'airtime-conf-production.php');  `

So this is going to require figuring out an alternative way to clean this up and have it load the settings from a install file to get it working out of the box. In case anyone wants to manually change this they can change those files in the meantime to use a database password other than airtime. Luckily postgres shouldn't be accessible to anyone w/o shell access to a server by default.

Robbt:
Also found that there exists documentation from "airtime" about this that references a now deleted CLI tool airtime-update-db-settings see that is set-up to update the file by generating the propel file from /usr/share/airtime/build/runtime-conf.xml "automatically" see http://libretime.org/manual/host-configuration/

Issue #360, opened 2017-11-16
Debian 9 - Install Fail (Step 4 of 5)


amavarick:
Step 4 of 5

Stays on this screen indefinitely:

![debian install freezing](https://user-images.githubusercontent.com/20918832/32916186-47ff0af2-cae9-11e7-933d-323f83cf24b3.png)

**Here are my exact install steps**:
```
apt install -y sudo nano locales
sudo apt-get update && sudo apt -y upgrade
sudo apt-get -y autoremove && sudo apt-get -y autoclean
#sudo reboot
sudo dpkg-reconfigure locales
sudo update-locale LANGUAGE="en_US.UTF-8"
sudo reboot
############# End Server Pre-Reqs #############################
#Disable IPv6 in /etc/hosts
#sudo sed -e 's/^::1/#::1/g' -i /etc/hosts
sudo sed -e 's/^fe/#fe/g' -i /etc/hosts
sudo sed -e 's/^ff/#ff/g' -i /etc/hosts
#Set SSHD to only listen to IPv4
sudo sed -i 's/^.*ListenAddress 0.0.0.0/ListenAddress 0.0.0.0/' /etc/ssh/sshd_config
sudo sed -i 's/^.*ListenAddress ::/#ListenAddress ::/' /etc/ssh/sshd_config
sudo sed -i 's/^.*Protocol 2/Protocol 2/' /etc/ssh/sshd_config
sudo sed -i 's/^.*ListenAddress ::/#ListenAddress ::/' /etc/ssh/sshd_config
sudo service sshd restart
#End Set SSHD to only listsen to IPv4
#Install automatic Updates
sudo apt -y install unattended-upgrades
#Install Libretime.org
#lsb_release -a
sudo apt -y install git && sudo git clone https://github.com/libretime/libretime.git
cd /root/libretime
./install -iap
```

Upon completing the steps above, I browse to the URL and walk through the setup wizard.  The setup wizard freezes on step 4/5.  After waiting hours to see if the screen changed, I did the following commands from CLI:
```
sudo service airtime-playout start
sudo service airtime-liquidsoap start
sudo service airtime_analyzer start
sudo service airtime-celery start
```
The configuration checklist appears to look good:
![configurationchecklist](https://user-images.githubusercontent.com/20918832/32916619-ade79518-caea-11e7-86e1-961d2a886062.png)

But after logging in, the settings are blank:
[...]

PR #361, opened 2017-11-17
Fix - repoint to airtime-production-conf.php to resolve hardcoded values and broken install

### LINKED ISSUE #330: Web installer doesn't update RabbitMQ and postgresql custom passwords in airtime.conf

Robbt:
This is part of the requirement of hardcoded database values tracked on issue #356.
My change to fix #330 actually caused #360 because it wasn't properly writing the media directory and thus causing 

Error: Call to a member function getDirectory() on null in /vagrant/airtime_mvc/application/models/MusicDir.php:34 

Which results in a UI with no values whatsoever.

Changing this back fixes it, but it still probably is still dependent upon a hardcoded database password to work.

So we need to either rewrite the hardcoded calls to the php based propel inits or something else.

### DIFF
--- a/airtime_mvc/public/setup/media-setup.php
+++ b/airtime_mvc/public/setup/media-setup.php
@@ -104,7 +104,7 @@ class MediaSetup extends Setup {
     function setupMusicDirectory() {
         try {
             $_SERVER['AIRTIME_CONF'] = AIRTIME_CONF_TEMP_PATH;
-            Propel::init(AIRTIME_CONF_TEMP_PATH);
+            Propel::init(CONFIG_PATH . "airtime-conf-production.php");
             $con = Propel::getConnection();
         } catch(Exception $e) {
             self::$message = "Failed to insert media folder; database isn't configured properly!";


Robbt:
Ok so I just tested this and I think that my change for this file was completely broken and we should undo it. Also since we now have /etc/airtime/airtime.conf with the changed DB values the airtime-production-conf.php is working and reading properly during install. I think this even fixes #356 as I was able to run the install with a altered database password and get into LibreTime and have it work without the errors I was confronted with previously.

So while this change should go through, someone else should test to confirm that it fixes #356

PR #362, opened 2017-11-19
Update Calendar Manual

paddatrapper:
Part of a drive to update the manual to reflect the current status of LibreTime. This updates the Calendar section of the manual.


Robbt:
This looks great. Thanks for taking the time to do this. The only question I have is about the removal of Cancel Current Show. On my version of LibreTime this is still the dialog box that pops up and I don't think we currently are supporting the recording of shows at this point.

paddatrapper:
I just checked mine and it has the same. The show I was testing must have had recording enabled (though no recorded file was produced, not sure if that was because of the lack of a physical line in or if it isn't set-up/supported). I'll fix that section later today

paddatrapper:
Fixed

hairmare:
Awesome guys! Nice work!

Issue #363, opened 2017-11-22
Feature request: Preview for dynamic smartblock


Robbt:
The problem: There is no way to validate or otherwise preview the results of a dynamic smartblock without setting the type to static, clicking generate, validating the results and then changing it back to dynamic.

What would be nice: The generate button could change to a preview button when viewing a dynamic smartblock and clicking it would generate a preview of a possible result of creating the smartblock.

PR #301, opened 2017-09-20
Initial Zanata integration

hairmare:
Most of this is me manually pushing and pulling the translations to and from zanata as well as updating the docs so they have a clear pointer to zanata.org. Feel free to nitpick.

For the time being I'll do PRs on the zanata changes before each release (as is included in this one with b0211ed6c818e95b6eabc1b659d4e97e816b632c). In the long run I'll automate this with a zanata API key that is configured as a secret in travis. You can expect some PRs changing `.travis.yml` for the next couple of releases. Right now I consider getting the docs update from this pr more important even though I'd have liked to start automating it with this PR as well.

PR #364, opened 2017-11-22
Feature: Modified form to allow preview of tracks for dynamic smartblocks 

Robbt:
This commit basically consists of a relabeling of the Generate button to Preview for dynamic smartblocks and the addition of a message above the tracks indicating that they are only a preview. I disabled some of the javascript that disabled the "Generate" button for non-static smartblocks. 

The way in which it is somewhat hacky is that it stores the tracklist preview in the database as if it were a static smartblock, but it still dynamically generates the contents when scheduled. This was an easier means of meeting the goal than writing code to create a separate preview action that would do the same thing but discard the values.

One possible improvement would be to further alter the CSS on the "preview" tracks to gray them out or otherwise show that they are only a preview, but I think the text label should suffice, but I might try to improve the styling of it.

I'm also not 100% sure that the modifications are easy to translate but I did try to utilize the _ in the php and I wrapped the javascript html change as such .html($.i18n._("Generate")) so I'm hoping that is the correct way to make the inner contents easy to translate.

This goes with the newly created issue #363


hairmare:
Again, sorry about taking so long to get reviewing these.

This one is great and it worked well during my testing.

The js parts need adding to the LocaleController manually since the gettext parts don't pick up on JS translations.

https://github.com/LibreTime/libretime/blob/0327a750ad6a1b3a96c650cf344de058a76a1bac/airtime_mvc/application/controllers/LocaleController.php#L21-L28

The linked method only actually gets used so the strings get injected into the gettext files.

The actual gettext files also need updating, but you can skip that part since I've been managing those parts in the yet to be merged branch from #301. I really should work on updating/finalizing that to get translations up and running through zanata.

Robbt:
Ok, I've added the appropriate javascript strings to the LocaleController, so this should be ready for merging.

PR #369, opened 2017-11-25
Add CORS handling to podcast feed

rfb:
I'm building an archive page for our station that will pull libretime's podcast feed. I needed to add CORS support to the file, and noticed libretime already handles CORS responses for other resources.

### DIFF
--- a/airtime_mvc/application/controllers/FeedsController.php
+++ b/airtime_mvc/application/controllers/FeedsController.php
@@ -7,14 +7,17 @@ class FeedsController extends Zend_Controller_Action
         $this->view->layout()->disableLayout();
         $this->_helper->viewRenderer->setNoRender(true);
 
+        $request = $this->getRequest();
+        $response = $this->getResponse();
+
         if ((Application_Model_Preference::getStationPodcastPrivacy()
-                && $this->getRequest()->getParam("sharing_token") != Application_Model_Preference::getStationPodcastDownloadKey())
+                && $request->getParam("sharing_token") != Application_Model_Preference::getStationPodcastDownloadKey())
                 && !RestAuth::verifyAuth(true, false, $this)) {
-            $this->getResponse()
-                ->setHttpResponseCode(401);
+            $response->setHttpResponseCode(401);
             return;
         }
 
+        CORSHelper::enableCrossOriginRequests($request, $response);
 
         $rssData = Application_Service_PodcastService::createStationRssFeed();
 


rfb:
Fantastic, thank you for the quick response.

rfb:
@hairmare hold off on merging this. `CORSHelper::enableCrossOriginRequests` explodes if the origin header is null (`Origin: null`). I imagine podcasting hosts and clients don't often set the Origin! I'll take another look. 

Where would I write tests for this in Libretime?

rfb:
Ugh, Libretime has some odd CORS behavior. 

I believe the correct behavior if the request origin is not recognized, is for the server to return the resource without a 'Access-Control-Allow-Origin` header. 

Instead, [Libretime is setup to throw a 403](https://github.com/LibreTime/libretime/blob/f17a3cb5bb5415bf59d9535acceb62f75ff40ae8/airtime_mvc/application/common/CORSHelper.php#L11-L16) if the origin isn't recognized.

Robbt:
Hmm, the testing coverage situation is pretty minimal, but there is a basic test suite set-up with PhpUnit. The instructions for how to run the tests are on the [website](http://libretime.org/testing/). I'm not sure if it will work with what you are aiming for but that is what has been set-up thus far.

hairmare:
The php tests are in `airtime_mvc/tests` and they are PHPUnit based. As Robb hinted, they don't cover an awful lot. If you plan on adding test for the CORSHelper you will probably have to create new folders and a test-class for it since no one has implemented any before.

I haven't seen an `Origin: null` header as of yet. Currently the code correctly ignores `Origin: ` but doesn't like `null`. I was rather surprised to find that [the W3C TR on CORS](https://www.w3.org/TR/cors/#access-control-allow-origin-response-header) lists the string `null` as special case.

We should probably allow requests with the string `null` in their Origin header at the [same place](https://github.com/LibreTime/libretime/blob/f17a3cb5bb5415bf59d9535acceb62f75ff40ae8/airtime_mvc/application/common/CORSHelper.php#L11) where `$origin != ""` is checked in `CORSHelper::enableCrossOriginRequests`.

rfb:
@hairmare ok, I'll add that then. thanks!

hairmare:
I'd propose merging this since it is unrelated to the `$origin == null` bug. @rfb Does that sound right to you?

Robbt:
Is this ready to merge ?

hairmare:
I'm merging this since the changes it makes are sensible and a step in the right direction.

We should probably open an issue for `Origin: null` as it won't get supported by this PR but some clients might still send it as a valid Origin header.

Issue #366, opened 2017-11-22
Problem: Smartblocks no longer show number of files that meet the criteria


Robbt:
This is a minor issue but one that makes it more difficult for a human to validate their smartblock form is correct while they are writing it.

Basically I've traced it down to commit b6f33f60 and the refactoring of the smartblock where the $formisvalid parameter is no longer being sent and createUpdateResonse is used instead of createFullResponse w/ the isValid parameter being passed.

Robbt:
Ok so the problem here is that the means through which this would be updated is bypassed by the refactored way of doing things. The files meeting the criteria are part of the smart-block-criteria.phtml file and this is no longer generated touched after the initial load. 

This view is created by the function startForm in SmartBlockCriteria.php but it always takes the default parameter $p_isValid = $false. Previously this whole view would of been regenerated everytime a change was made by the createFullResponse function but the createUpdateResponse function only touches the update.phtml view  which only shows the tracks in the playlist (or smartblock). 

So there isn't an obvious simple fix here and it will require additional refactoring to reimplement this feature. Currently the display code is contained inside of the smart-block-criteria.phtml and displayed based upon the $showPoolCount variable (which is only set to true if $p_isValid is passed to the startForm.

So that's where this stands at the moment. anyone who wants to dive in and provide suggestions is welcome to do so.

Robbt:
In the spirit of updating my issue queue before I walk away I have determined the easiest course of action would be to perform an update w/ the createUpdateResponse function by simply calling $this->view-criteriaMatch = $obj->getListofFilesMeetCriteria()['count'];
similar to the way the length is updated and other portions not directly rendered. The issue right now is that there is no field called criteriaMatch and no mechanism in place with the current setup to display or hide that field based upon valid criteria as was developed before. So that is the next step.

Robbt:
Running into the issue that I have no idea where the $this-view-variables are set. For instance $this->view-length isn't part of the update.phtml rendered view yet it is updated by the createUpdateResponse function. When I tried to create an arbitrary variable and get it to display and change upon update it doesn't show at all. 

I tried last night to enable zfdebug and had to delete the memory and time options from it because they are no longer included in zfdebug but it doesn't show any of the variables related to the actual playlist. At one point I tried to setup a simple debugger so I could walk through the variables etc but I'm not sure how to do that with LibreTime especially considering so much happens as AJAX calls etc.

[...]

PR #370, opened 2017-11-26
Moved display of smartblock files that meet criteria to update.phtml

Robbt:
Basically I recreated what originally existed in the smart-block-criteria.phtml view and put it at the top of the track listing so that it will display with update.phtml and be updated when the criteria is changed and saved.

I'm not 100% happy with how this changes the UI but it is definitely the easiest option at this point and avoids touching the javascript to try to update it. Any suggestions for UI improvements are encouraged.  It might make sense to color it with CSS to make it more distinct but otherwise I think that this reimplements the functionality requested in #366

### DIFF
--- a/airtime_mvc/application/views/scripts/form/smart-block-criteria.phtml
+++ b/airtime_mvc/application/views/scripts/form/smart-block-criteria.phtml
@@ -119,37 +119,6 @@
             
         <?php } ?>
         </dd>
-
-        <?php
-        // this will never run due to refactoring of playlist controller. it is replaced with code in update.phtml
-        if ($this->showPoolCount) { ?>
-            <div class='sp_text_font sp_text_font_bold'>
-                <span id='sp_pool_count' class='sp_text_font sp_text_font_bold'>
-                <?php
-                if ($this->poolCount > 1) {
-                    echo $this->poolCount;
-                ?>
-                    <?php echo _("files meet the criteria")?>
-                    </span>
-                    <span class='checked-icon sp-checked-icon' id='sp_pool_count_icon'></span>
-                <?php
-                } else if ($this->poolCount == 1) {
-                    echo $this->poolCount;
-                ?>
-                    <?php echo _("file meets the criteria")?>
-                    </span>
-                    <span class='checked-icon sp-checked-icon' id='sp_pool_count_icon'></span>
-                <?php
-                } else {
-                ?>
-                    0 <?php echo " "._("files meet the criteria")?>
-                    </span>
-                    <span class='sp-warning-icon' id='sp_pool_count_icon'></span>
-                <?php
-                }
-                ?>
-            </div>
-        <?php } ?>
     </dl>
 
 </form>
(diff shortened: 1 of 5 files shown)


airtime_mvc/application/views/scripts/form/smart-block-criteria.phtml
```diff
@@ -120,7 +120,9 @@
         <?php } ?>
         </dd>
 
-        <?php if ($this->showPoolCount) { ?>
+        <?php
+        // this will never run due to refactoring of playlist controller. it is replaced with code in update.phtml
```
hairmare: Can we remove this whole block then? Since `$this->showPoolCount` is long gone this is just dead code and we can get rid of the whole div without any side effects.

airtime_mvc/application/views/scripts/playlist/update.phtml
```diff
@@ -2,6 +2,35 @@
...
+                <span id='sp_pool_count' class='sp_text_font sp_text_font_bold'>
+                <?php
+                if ($this->poolCount > 1) {
+                echo $this->poolCount;
+                ?>
+                <?php echo _("files meet the criteria")?>
+                    </span>
+    <span class='checked-icon sp-checked-icon' id='sp_pool_count_icon'></span>
+    <?php
+    } else if ($this->poolCount == 1) {
```
hairmare: You might want to consider using [plural support in zf1](https://framework.zend.com/manual/1.11/en/zend.translate.plurals.html) for this. You should be able to use the adapter as follows:

```php
echo $this->getTranslator()->plural("file meets the criteria", "files meets the criteria", $this->poolCount);
```

This would also shift the burden of knowing how to handle the zero case to the translators (TIL that not all languages handle the zero case like English).

airtime_mvc/application/views/scripts/playlist/update.phtml
```diff
@@ -2,6 +2,35 @@
 $items = $this->contents;
 $isStaticSmartBlock = ($this->obj instanceof Application_Model_Block && $this->obj->isStatic());
 $isPlaylist = ($this->obj instanceof Application_Model_Playlist);
+if ($this->poolCount) { ?>
+<div class='sp_text_font sp_text_font_bold'>
```
hairmare: Some CSS might have gone missing, at least the text seems to have very low contrast. Most likely this got kicked during a redesign and never had it's style updated.

image

We might want to consider cleaning this up for more contrast. I won't hold off on merging over this though.

airtime_mvc/application/views/scripts/playlist/update.phtml
Robbt: Looking into this. Your suggestion didn't work as getTranslator doesn't exist as a function. I'm trying to figure out if I should create a new Translator object to pull in plurals support. The only mention of Zend_Translate I see is in the rest RouteController.php file at this point.

Robbt:
I did a commit with ngettext since the Zend_Translate object appeared to be missing. I also deleted some of the deadcode and fixed the logic so that it would display a pool count only on updates and the no files match message.

Issue #371, opened 2017-11-28
Problem: Shows set to be scheduled with Auto Playlist show as empty on the Calendar


Robbt:
Problem: Basically any show that has an automatic playlist is not differentiated in any way from a show that is completely empty. They both show the Red ! mark.

The solution: Add a new indicator that displays something like a green A to indicate an automatic playlist. It might even be possible to do a test run of the schedule to determine if the playlist will "fill" the show based upon projected length. This could change based upon the varying lengths of a dynamic files but it could also be feasible to indicate that a show will likely underrun and could be marked with a yellow A.

This is all done in javascript right now in the eventRender function in full-calendar-functions.js which retrieves its events from the &getFullCalendarEvents in Show.php

Code to follow.

Robbt:
Ok, so the way it determines if a show is filled is currently done via a function in the show_instances object called getIsFull that queries the database and compares the ends-starts of a show_instance vs. the time_filled. So it doesn't really make sense to create something similar for "couldBeFilled".

So the simple thing to do is just replace the ! with an A and change the color, but the issue here is that this doesn't necessarily indicate that a show won't require manual intervention to not have dead air. Also those signs can be a good indicator of failed playlists. So I'm thinking that we just add an A to the left of the ! so that there is an indication that there is an automatic playlist associated with the show and work on any projections later as we refine the tools (for instance by adding flexible smartblocks that automatically fill a show based upon length etc). We could add another indicator that a autoplaylist will repeat until full perhaps with a different color.

Robbt:
Probably going to need to write a new showInstance function that returns shows with an automatic playlist.

Robbt:
This was merged.

PR #202, opened 2017-05-16
[WIP] Run dos2unix on all markdown files in docs/

hairmare:
## Problem

The docs had mostly Windows CRLF line endings. I assume these were created with the CRLF line ending during my conversion of them for some reason. My mac being smarter as need be hid this from me perfectly and it only started breaking when I worked on the upgrade docs. Since the files where in the repo with CRLF line endings they always got mapped by git. This lead to the weird behaviour seen in #188 where all the lines always changed.

## Solution

Temporarily remove the `.gitattributes` file from the local working copy, run dos2unix on all affected markdown files and commit results. Re-add `.gitattributes`.

I created a branch with some minimal changes to verify that I can now edit a file without the line endings getting changed throughout the file: https://github.com/radiorabe/libretime/commit/ba2247b1f96ee95ecb340cb2cea1087625930c3e.

This PR only contains line end changes and no changes to the actual content of the markdown file.

PR #373, opened 2017-11-29
Fix/display autoplaylist on calendar

Robbt:
This addresses #371 by adding a A icon with green background to shows on the Calendar that have a automatic playlist. I think the icon could be improved but it is sufficient and working.


airtime_mvc/application/models/Show.php
```diff
@@ -1055,6 +1057,7 @@ public static function &getFullCalendarEvents($p_start, $p_end, $p_editable=fals
 
             $startsDT->setTimezone($displayTimezone);
             $endsDT->setTimezone($displayTimezone);
+            xdebug_break();
```
hairmare: left over debug code

airtime_mvc/application/models/ShowInstance.php
```diff
@@ -238,6 +238,7 @@ public function addPlaylistToShow($pl_id, $checkUserPerm = true)
         $id = $this->_showInstance->getDbId();
         $lastid = $this->getLastAudioItemId();
 //        Logging::info("The last id is $lastid");
+        xdebug_break();
```
hairmare: more left over debug code

airtime_mvc/application/models/ShowInstance.php
```diff
@@ -290,7 +291,9 @@ public function clearShow()
     }
 
     private function checkToDeleteShow($showId)
+
     {
+        xdebug_break();
```
hairmare: more loft over debug code

airtime_mvc/application/models/ShowInstance.php
```diff
@@ -556,6 +559,27 @@ public static function getIsFull($p_start, $p_end)
         return $isFilled;
     }
 
+    public static function getShowHasAutoplaylist($p_start, $p_end)
+    {
+        $sql = <<<SQL
+SELECT c.id, s.has_autoplaylist
```
hairmare: This looks like a fairly simple SQL statement, maybe the OR-Mapper can do this?

airtime_mvc/public/js/airtime/schedule/full-calendar-functions.js
```diff
@@ -214,7 +214,13 @@ function eventRender(event, element, view) {
                     $(element)
                     .find(".fc-event-time")
                     .before('<span class="small-icon linked"></span><span class="small-icon show-empty"></span>');
-                } else {
+                   // in theory a linked show shouldn't have an automatic playlist so adding this here
```
hairmare: seems sensible.

Robbt: So I put in most of the changes you requested. I didn't modify the JS code because it was working as intended, the comment was just pointing out the lack of a need for another branch where linked would also have an autoplaylist, but the two other branches were necessary.

Also the image is working locally but I'm not sure if there is a problem with git and binary files again. Ever since I merged some document change commits I have started running into issues like this 
warning: CRLF will be replaced by LF in airtime_mvc/public/css/images/icon_alert_cal_autoplaylist.png.

If there is an issue with the png then this is likely the case. 

Although it is basically seeing binary png and jpgs that have been updated as modified on my local git directory as we saw in #202 and I've needed to do my hard reset command referenced there to be able to do anything with my local repo. I tried updating git to via PPA to see if that makes a difference.

hairmare: I think the image issue might have something to do with the `.gitattributes` file. I'll have a look at the image locally and merge if it looks ok there.

hairmare: I just re-checked but somehow the file really does seem corrupted.

Here is what pngcheck is reporting:

```
pngcheck -v airtime_mvc/public/css/images/icon_alert_cal_autoplaylist.png
File: airtime_mvc/public/css/images/icon_alert_cal_autoplaylist.png (571 bytes)
  File is CORRUPTED.  It seems to have suffered DOS->Unix conversion.
ERRORS DETECTED in airtime_mvc/public/css/images/icon_alert_cal_autoplaylist.png
```

Robbt: Ok, I'll push a commit with the original file and make sure it doesn't get corrupted shortly.

hairmare:
I finally got to testing this, the php/js parts seem to be working but it looks like you missed pushing your changes to `airtime_mvc/public/css/styles.css`.

Robbt:
Ugh, just noticed that all of these other unrelated commits are showing up in here. Not sure if this will cause any problems. When testing this I've noticed some oddities in terms of whether the autoplaylist icon shows up or not that require further debugging. Let me know if I need to do this from a fresh branch or somehow otherwise clean up the various merges that got integrated into this branch.

hairmare:
Your merge commands seem to have tricked the GitHub ui, the commits shown have already landed in that form on master and there aren't any conflicts so git will be able to just do the right thing :)

Robbt:
I just realized that the conversion to an ORM based query isn't actually working at this point. So whether something shows up as an auto_playlist or not isn't based upon the code. I need to diagnose this further, but I'm thinking we should revert it if I can't get a fix in soon. I suppose it's not critical at this point because it is just a UI bug vs. something that will break the program, but I'm working to figure it out.

PR #393, opened 2018-01-02
Changed 'System->Streams' to 'Settings->Streams'

HammyHavoc:
The discrepancy in terms used definitely confused me to begin with. :- )


hairmare:
Can you remove the locale file changes from your changeset? They should be done by the translation tools and should be isolated in their own commit as such.

hairmare:
I'll have to clean up the translation stuff anyhow and this doesn't make more of a mess that there already is so I'm merging it before it has any merge-conflict issues :)

PR #395, opened 2018-01-02
Making documentation less confusing

HammyHavoc:


hairmare:
Thanks for all of these! 

If i recall correctly this was renamed everywhere except the docs during the SaaS fork of legacy upstream.

Issue #404, opened 2018-01-08
Problem: Clear button text has been replaced with Generate or Preview after preview PR


Robbt:
I just noticed this on my vagrant box, but after the PR was merged that allows preview for some reason the "Clear" button text shows either Preview of Generate.

I didn't see any obvious reason for this in the code, but I suspect it's an issue with Zend Form or jquery. I'll work on submitting a fix but wanted to document that it exists.

PR #405, opened 2018-01-08
modified jquery find for shuffle and generate buttons to not target clear button

Robbt:
The jquery activeTab.find included a variable that matched the css for the clear button. When I renamed the generate button this was also rewriting the HTML for the clear button thus overriding it's text.

By removing this from the identifier everything appears to be working with the text as it should be. I'm not sure why the jquery was written like that but based upon some testing the change doesn't alter the functionality.


Robbt:
This fixes #404

PR #408, opened 2018-01-16
Issue350b advsearch calendermode

jerry924:
This fix is the cleanest way I can find to resolve the issue introduced from the previous fix of this issue. Now both panels are clickable as expected and no other faults can be found with this change.


jerry924:
Fixed remaining issue from previous commit

Issue #350, opened 2017-11-10
Advanced Search not available in Calendar mode


jerry924:
Libretime version: 0dd0a443f

When I am in the calendar mode and I want to add tracks to a specific show I have selected from the calendar, there is no option to perform an advanced multifield search or time range search.

![image](https://user-images.githubusercontent.com/7436317/32639299-1197cbfa-c578-11e7-9ed2-5c605d2f7df7.png)

When I am in the Tracks view, this option exists.

![image](https://user-images.githubusercontent.com/7436317/32639334-462b33a2-c578-11e7-87bc-5410574b793b.png)

I really need the advanced search when I am trying to build my show.

hairmare:
Meh, you stumbled upon a bug in the datatables based UI parts of the frontend. I've made multiple attempts at restoring sanity in those code paths but have failed so far since the indivual parts of those pages are tightly coupled in a fashion that doesn't lead to an easy fix.

Under the hood the funtionality is there but refactoring the UI parts to do the right things seems rather elusive given the messy code underneath the calendar scheduling page.

jerry924:
The good news is the button appears to be there. I was playing around with the CSS editor in Chrome and I actually got it to show up. Now trying to reproduce it since I didn't notice which change made it appear. I think it may just be hidden below the input box, but now trying z-index isn't doing it.

Good news is at least in chrome dev tools I got it to show up at least once!

jerry924:
Ok, found this problem.

In CSS,  ```.panel-header{  z-index:1;  }```  is causing the Advanced Search carrot to be hidden.    

If you change ```z-index: auto;``` 

Then it shows up and functions perfectly fine.

![image](https://user-images.githubusercontent.com/7436317/32853403-0fac4c86-c9f0-11e7-8423-2c76cac703ff.png)

jerry924:
Pull Request #359 submitted

hairmare:
Great to see that you found the underlying issue. Thanks for the PR!

jerry924:
This fix had the unintended consequence of not allowing the Show Dialog to be closed in Dashboard mode.  The "X" here became unclickable.

The problem here is that in Dashboard mode and in Calendar mode, the same panel-header style is used for both, and the competing z-index setting will never allow both to work together. I have a transaction where I forked the panel-header into panel-header-calendar to be used for the Calendar-mode dialog.  

This is now in my environment, but I want to spend more time testing before doing a PR.

[...]

PR #409, opened 2018-01-17
Issue350c better advsearch fix

jerry924:
This is an improved fix for issue 350 that resolves the loss of closing the calendar panel while in Dashboard view.


hairmare:
Fixes #350

Issue #428, opened 2018-01-28
zendphp fails to register an IPv6 address for pypo and show-recorder.  


detobate:
From /var/log/airtime/zendphp.log:
Next PropelException: Unable to execute UPDATE statement [UPDATE cc_service_register SET ip=:p1 WHERE cc_service_register.name=:p2] [wrapped: SQLSTATE[22001]: String data, right truncated: 7 ERROR:  value too long for type character varying(18)] in /usr/share/airtime/php/vendor/propel/propel1/runtime/lib/util/BasePeer.php:450

ip column in cc_service_register too small for IPv6
Altering the column size to 45 chars allows it to register an IPv6 address and carry on its merry way.  Haven't found anything that has broken with it yet. 

airtime=> ALTER TABLE cc_service_register ALTER COLUMN ip TYPE character varying(45);
ALTER TABLE

airtime=> select * from cc_service_register;
     name      |            ip
---------------+--------------------------
 media-monitor | 5.196.xxx.yyy
 pypo          | 2001:41d0:[obfusacted]
 show-recorder | 2001:41d0:[obfuscated]
(3 rows)

This was a fresh install of Libretime for me, so I believe things were breaking as the entries had yet to be populated at all.  Existing installs might carry on working and mask this issue if they've already got IPv4 addresses in the table.

Unsure on the dependancies, but it might be worth looking to use psql's inet datatype instead of varchar?

hairmare:
There might be other issues with IPv6 that havent' shown up yet. I'm pretty certain that you need a really current liquidsoap version for IPv6 (at least if your stack is IPv6 native, but yours seems to be hybrid). The changes needed for using a current liquidsoap are linked/described in #192.

detobate:
Almost certainly, but this is one step in the right direction.
This host is indeed dual stacked, and everything seemed to go swimmingly after allowing IPv6 addresses to be registered in the db.

hairmare:
This is indeed a step in the right direction, I'm not sure if the current ORM in PHP actually supports the psql inet datatype so your change from #429 is probably the way to go.

PR #429, opened 2018-01-28
Update ip column size to allow for IPv6 addresses

detobate:
Allow for the use of IPv6 address as per [issue 428](https://github.com/LibreTime/libretime/issues/428)

45 chars are used instead of the expected 39 because of [IPv4-embedded](https://stackoverflow.com/a/1076755) addresses


hairmare:
I added the needed changes in #433, this will auto-close once that is merged.

PR #433, opened 2018-02-03
Fix/429 ipv6 schema updates

### LINKED ISSUE #428: zendphp fails to register an IPv6 address for pypo and show-recorder.

hairmare:
I regenerated the schemas and added migrations based on the work in #429.

Fixes #429 
Fixes #428

### DIFF
--- a/airtime_mvc/application/controllers/downgrade_sql/airtime_3.0.0-alpha.6/downgrade.sql
+++ b/airtime_mvc/application/controllers/downgrade_sql/airtime_3.0.0-alpha.6/downgrade.sql
@@ -1,0 +1,2 @@
+ALTER TABLE cc_service_register ALTER COLUMN ip TYPE character varying(18);
+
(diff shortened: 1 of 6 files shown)


detobate:
Thanks!  Sorry, I was going to try to get around to doing it this weekend also.

hairmare:
No worries, I had the code open anyway and figured it would be a quick fix.

Issue #311, opened 2017-09-28
ShoutCAST DNAS rejecting title updates


expert-geeks:
Libretime version: Current pull from git repo 29e7cc3
Ubuntu 16.04
ShoutCAST DNAS 2.5.5.732

I'm trying to connect to ShoutCAST DNAS and while the audio connects fine, the ID3 data is discarded. By checking the logs after a chance [forum remark](http://forums.winamp.com/showthread.php?t=381045), it seems that ShoutCAST is ignoring title updates generated by libretime. ShoutCAST DNAS apparently doesn't like titles that 1) start with punctuation or special characters, or 2) multiples of special characters in the title and stops DNAS updating the title.

When connecting, DNAS accepts the initial offline title (excepts taken from sc_serv.log);
`Title update [LibreTime - offline]`
but then rejects the track data, leaving the 'Libretime - offline message', in the DNAS ID3 cache.

When '_Stream Label:_ Artist - Title' is selected, Libretime generates the titles with a leading dash, and then ShoutCAST DNAS rejects the update the following way
`Title update rejected - value not allowed: - Artist - Track`

When '_Stream Label:_ Show - Artist - Title' is selected, Libretime generates the titles with two leading dashes, and then ShoutCAST DNAS rejects the update the following way

`Title update rejected - value not allowed: -  - Artist - Track`

When '_Stream Label:_ Station name - Show name'  is selected, Libretime generates the following;

`Title update [Station Name -]`

ShoutCAST DNAS accepts this, because it doesn't start with a dash.. neither does it contain any show info.. 

Icecast in the same setup clearly doesn't care about dashes and is proudly displaying it variously;

```
Stream Name: Station Name
Currently playing: - Artist - Title
Currently playing: -  - Artist - Title
Currently playing:Station Name - 
```

Has anyone else successfully connected to ShoutCAST with ID3/ICY data from libretime ?

Is it possible to stop libretime from passing ID3/ICY data to ShoutCAST DNAS (or anything else for that matter) that starts with dashes/special characters by stripping them out ?

If the ID3 field is blank or not passed, can we stop libretime appending a dash to the ID3 data ?

hairmare:
Hi @expert-geeks 

Welcome to the community!

I didn't know about ShoutCAST's limitiations when it comes to dashes in the metadata.

The current Stream Label selection thing is rather limited. We should probably open the feature up in a way that allows for a more configurable label.

[...]

PR #456, opened 2018-03-04
Stop ShoutCAST rejecting title updates

expert-geeks:
When Libretime streams a webstream, the ID3 data is put into a single title field, leaving the artist field blank. When processing the ID3 data, Libretime concatenates the blank 'artist' field with the title, leaving '- title'. ShoutCAST rejects title updates that begin with punctuation for stylistic reasons (thanks guys) and so effectively ignores all ID3 data coming with the stream, falling back to the 'Station off-air' message. This PR uses the title field alone if the artist field is blank, but both when the artist data is available e.g. when streaming an MP3.

### DIFF
--- a/python_apps/pypo/liquidsoap/ls_lib.liq
+++ b/python_apps/pypo/liquidsoap/ls_lib.liq
@@ -37,8 +37,12 @@ def append_title(m) =
       elsif !stream_metadata_type == 2 then
         [("title", "#{!station_name} - #{!show_name}"), ("mapped", "true")]
       else
-        [("title", "#{m['artist']} - #{m['title']}"), ("mapped", "true")]
-      end
+        if "#{m['artist']}" == "" then
+          [("title", "#{m['title']}"), ("mapped", "true")]
+        else
+          [("title", "#{m['artist']} - #{m['title']}"), ("mapped", "true")]
+        end
+     end
   end
 end
 


hairmare:
Thanks @expert-geeks

I'll do some smoke testing on a vagrant box before I merge this so it might take a couple of days until I find the time to test and merge this. It looks good at a quick glance.

hairmare:
Fixes #311

hairmare:
Thanks @expert-geeks 

I did some basic smoke testing on a vagrant box and everything works. I didn't check ShoutCAST since I don't have a server but it does look like this might make LibreTime usable with ShoutCAST.

expert-geeks:
Excellent. It's been working fine for us for the past 2 weeks or so. Glad to help and give back a bit to the project.

teris:
Shoutcast requires the legacy password in version 1.X (sc_serv). A connection with a multistream is only possible with ShoutCast 2.X.
However, a distinction is made between legacy password and the stream password.
In order to establish a connection, the selection of password and StreamID for Shoutcast 2.X is required
The password for the Shoutcast server is then defined as follows:

- Stream ID: 1
- Server Address: localhost
- Source Port: 8000
- Password: secret
- Legacy Password: secret:#1
- Protocol Mode: v2

- Stream ID: 2
- Server Address: localhost
- Source Port: 8000
- Password: completelysecret
- Legacy Password: completelysecret:#2
- Protocol Mode: v2

The developer wiki can be found here:
http://wiki.winamp.com/wiki/SHOUTcast_DNAS_Server_2

The stable servers I have summarized here once:
https://github.com/teris/sc_trans-sc_serv-files

Examples of PHP processing can be found in my Shoutcast Admincenter:
https://github.com/teris/Shoutcast-Interface-with-SC_Trans

Maybe I could give you a little help and prepare configurations for SV_SERV.
Unfortunately, I am not good at shell / bash....

Issue #271, opened 2017-08-10
Problem: certain podcasts fail to download files with "can't sync to an MPEG frame" error


adamburnsuk:
Hello everyone,

This is a bit of a feature request and a question at the same time.  I run an online podcast network.  We do live streaming using an icecast server, but are looking to add pre-recorded shows as well.  These shows are all podcasts and release shows on a daily and/or weekly basis.  I have created a java program that will automatically download the latest episode of each show and put in a specific folder.  I have seen an issue or two dealing with podcasts, so maybe I'm missing a feature that already exists in LibreTime, if so, apologies.

But, I would like to create a schedule in LibreTime and just replace the files every day with the newest version.  Based on my limited understanding (I'm new to LibreTime/Airtime) each new file gets it's own database entry, so I'm not sure I can just replace files.  When I try to do a smart list, it seems to select the files when I add them to the show and never get updated again.

Does anyone have an idea of how I might set up this automated solution?  

Thanks,
Adam

Robbt:
Hey Adam, not sure how I missed this issue. But there is a podcast subscription option added to LibreTime, you can just subscribe to a podcast and it will automatically download the newest one as well as create a playlist that contains a smartblock that will automatically update to the latest episode. 

Check out this wiki page for more details on how to set this up - https://github.com/LibreTime/libretime/wiki/Automatic-Playlists,-Smartblocks-and-Podcasts

I'm going to mark this closed for now but feel free to ask questions, there is also now a forum at discourse.libretime.org

adamburnsuk:
Hey Rob,

Thanks for the personal reply.  I've installed libretime and am trying out
the podcast feature.  It's FANTASTIC.  I do have a question though.  I also
posted this in discord.  I thought I'd email this direct to you in case you
have any tips.  I am a developer (automation specialist)...but mostly
familiar with javascript, java, and c#.  So, it is going to be a bit of a
slow process figuring out how to debug php/python code.

Anyway, I am trying to use the podcast functionality and I don't know much
about how to debug issues.  But some of the podcast feeds, when i try to
import episodes, it will spin, but the green check mark will never appear
and the track doesn't seem to be imported.  There are no errors thrown, so
I'm unsure how to troubleshoot this.

Other podcasts work fine.  The feeds validate, so I don't think it's an
issue with them.

Any help or even tips for how to go about debugging the podcast import
feature would be really appreciated.

Thanks,
Adam

Robbt:
[...]

PR #464, opened 2018-03-11
Update PodcastEpisodeService.php

JohnnyC1951:
Previously, this code imported the first enclosure from a podcast item and assumed that it was an audio podcast. However, some podcasts have image, file (transcripts) and even video enclosures, as well, which choked the import.
It has now been modified, with a simple test, to only import the audio enclosure and thus make more RSS feeds usable.

### DIFF
--- a/airtime_mvc/application/services/PodcastEpisodeService.php
+++ b/airtime_mvc/application/services/PodcastEpisodeService.php
@@ -392,10 +392,18 @@ class Application_Service_PodcastEpisodeService extends Application_Service_Thir
         $episodesArray = array();
         foreach ($rss->get_items() as $item) {
             /** @var SimplePie_Item $item */
-            // If the enclosure is empty or has not URL, this isn't a podcast episode (there's no audio data)
-            $enclosure = $item->get_enclosure();
-            $url = $enclosure instanceof SimplePie_Enclosure ? $enclosure->get_link() : $enclosure["link"];
-            if (empty($url)) { continue; }
+            // If the enclosure is empty or has not URL, this isn't a podcast episode (there's no audio data)     
+            // Added Checks for audio-only enclosures. JBC    
+            foreach ($item->get_enclosures() as $enclosure) {
+				$audio_podcast_found = false;
+				if (substr( $enclosure->get_type(), 0, 5 ) === 'audio') {
+					$url = $enclosure instanceof SimplePie_Enclosure ? $enclosure->get_link() : $enclosure["link"];
+					$audio_podcast_found = true;
+					break;
+				}		
+			}
+
+            if ($audio_podcast_found == false) { continue; }
             $itemId = $item->get_id();
             $ingested = in_array($itemId, $episodeIds) ? (empty($episodeFiles[$itemId]) ? -1 : 1) : 0;
             $file = $ingested > 0 && !empty($episodeFiles[$itemId]) ?
@@ -416,7 +424,8 @@ class Application_Service_PodcastEpisodeService extends Application_Service_Thir
                 "description" => htmlspecialchars($item->get_description()),
                 "pub_date" => $item->get_gmdate(),
                 "link" => $item->get_link(),
-                "enclosure" => $item->get_enclosure(),
+                // $enclosure now holds only an audio enclosure. JBC
+                "enclosure" => $enclosure,
                 "file" => $file
             ));
         }


adamburnsuk:
I would love to try this out if we can get the pull request approved and merged.  I'm having an issue with several podcasts not importing.  I'm not sure if that's the issue...but if it is, I'd love to get it resolved.

JohnnyC1951:
I gave up on that. Too many issues and not much support. So I made my own. Which was frowned upon, but works properly, at least here.

Robbt:
I'm not sure anyone is frowning upon your code, if it works that is good, but it would be good to be able to integrate it. I plan on sitting down and reviewing all of the PRs (including this) very soon so I can get back into my maintainer role.

JohnnyC1951:
I could paste it all here, but I think that would not be helpful, at all.
It could be summarized as: 'pointless making add-ins when what we need is core development.'
'OK Can I help?' 'No we don't like your coding style.'

JohnnyC1951:
I could paste it all here, but I think that would not be helpful, at all.
It could be summarized as: 'pointless making add-ins when what we need is core development.'
'OK Can I help?' 'No we don't like your coding
-----------------------------------
But we are not starting from there, are `we?`

JohnnyC1951:
I could paste it all here, but I think that would not be helpful, at all.
It could be summarized as: 'pointless making add-ins when what we need is core development.'
'OK Can I help?' 'No we don't like your codin

JohnnyC1951:
While I have your attention.. a tiny change to the radio page, The blurry green On Air pic is 514k big,
Yes half a megabyte. It should be like 45kb. Also the page load is really huge, including the zend framework, angular.js, lots more, to put up an single iframe, one log-in link and one station pic.

Robbt:
These are all good points. If someone reworks the radio page it'd be a good idea to take these into consideration. I'll be happy if I get around to just reviewing all of the PR's that I've been neglecting for the last few months. There is a ton of cruft in LibreTime.

JohnnyC1951:
Well, my suggestion is, if you need help ask around. There seem to be willing folks around

JohnnyC1951:
I will put in an informal PR with an optimized background image. That is very low hanging fruit.

Robbt:
Ok, so I'm reviewing this PR and I'm not sure what issue it is supposed to fix. I thought it might be related to #271 but that issue appears to happen on the python side as I still got the same error message.  

I tried it with https://pokeitwithastick.podbean.com/feed/ which was the podcast with an image that failed but it still fails to import.

Can you provide an example of a feed that didn't import before this patch but will import after it.

JohnnyC1951:
Hmm. :s From what I remember, It seems to happen when the FIRST enclosure is an image because it assumes that there is only  one enclosure and that enclosure is an .mp3. My own hom-ebrew importer uses the same fix.

Robbt:
ok, well it appears to fail when the 2nd enclosure is an image, do you have an example of a feed where the first enclosure is a image ?

JohnnyC1951:
I can load some up this weekend.. and see.. I am now using my own code for this, so my rss section in LT is presently empty

airtime_mvc/application/services/PodcastEpisodeService.php
```diff
@@ -392,10 +392,18 @@ public function _getImportedPodcastEpisodeArray($podcast, $episodes) {
...
         foreach ($rss->get_items() as $item) {
             /** @var SimplePie_Item $item */
-            // If the enclosure is empty or has not URL, this isn't a podcast episode (there's no audio data)
-            $enclosure = $item->get_enclosure();
-            $url = $enclosure instanceof SimplePie_Enclosure ? $enclosure->get_link() : $enclosure["link"];
-            if (empty($url)) { continue; }
+            // If the enclosure is empty or has not URL, this isn't a podcast episode (there's no audio data)     
+            // Added Checks for audio-only enclosures. JBC    
+            foreach ($item->get_enclosures() as $enclosure) {
+				$audio_podcast_found = false;
```
hairmare: It looks like this should be initialised outside of the scope of `foreach`.

airtime_mvc/application/services/PodcastEpisodeService.php
```diff
@@ -392,10 +392,18 @@ public function _getImportedPodcastEpisodeArray($podcast, $episodes) {
...
+				if (substr( $enclosure->get_type(), 0, 5 ) === 'audio') {
+					$url = $enclosure instanceof SimplePie_Enclosure ? $enclosure->get_link() : $enclosure["link"];
+					$audio_podcast_found = true;
+					break;
+				}		
+			}
+
+            if ($audio_podcast_found == false) { continue; }
```
hairmare: This check should use `===` since both sides of the check are boolean.

airtime_mvc/application/services/PodcastEpisodeService.php
```diff
@@ -392,10 +392,18 @@ public function _getImportedPodcastEpisodeArray($podcast, $episodes) {
...
-            $enclosure = $item->get_enclosure();
-            $url = $enclosure instanceof SimplePie_Enclosure ? $enclosure->get_link() : $enclosure["link"];
-            if (empty($url)) { continue; }
+            // If the enclosure is empty or has not URL, this isn't a podcast episode (there's no audio data)     
+            // Added Checks for audio-only enclosures. JBC    
+            foreach ($item->get_enclosures() as $enclosure) {
+				$audio_podcast_found = false;
+				if (substr( $enclosure->get_type(), 0, 5 ) === 'audio') {
+					$url = $enclosure instanceof SimplePie_Enclosure ? $enclosure->get_link() : $enclosure["link"];
+					$audio_podcast_found = true;
```
hairmare: The original code used to check `$url` for `empty()`. Maybe `$audio_podcast_foud = !empty($url)` and skipping the `break` depending on the results makes more sense here?

airtime_mvc/application/services/PodcastEpisodeService.php
```diff
@@ -416,7 +424,8 @@ public function _getImportedPodcastEpisodeArray($podcast, $episodes) {
                 "description" => htmlspecialchars($item->get_description()),
                 "pub_date" => $item->get_gmdate(),
                 "link" => $item->get_link(),
-                "enclosure" => $item->get_enclosure(),
+                // $enclosure now holds only an audio enclosure. JBC
+                "enclosure" => $enclosure,
```
hairmare: `$enclosure` might need initialising or a guard to ensure we never pass it as `null`.

JohnnyC1951:
This all seems over complicated. I just do foreach enclosure and break on an mp3

JohnnyC1951:
Hmm
I routinely see feeds with enclosures of Podcast Image, Podcast Media MP3, Podcast Media AAC and and Podcast Lyrics (often used for show script or reference stuff & links) - even pdfs
I handle that in my proxy so I don't care. Others might.

Robbt:
Yeah I was just curious if you could share any RSS feeds with this sort of content so we can test this PR. Thanks.

JohnnyC1951:
Will see if I can find one. Meanwhile, there is often confusion between RSS and Media RSS
RSS says you can have one media element but Media RSS says you can have as many as you want. 
[https://en.wikipedia.org/wiki/Media_RSS](https://en.wikipedia.org/wiki/Media_RSS)

Robbt:
Yeah I don't think anyone uses media RSS though and podcasts are all RSS 2.0 for the most part so they shouldn't have multiple enclosures unless they are improperly configured. So any samples so we can confirm this PR works would be good. Thanks.

JohnnyC1951:
Well, any I found would be Media RSS ones. :\ So I guess that is the question answered. Yahoo Feeds, Amazon and many more do use it. iTunes basically subverted it with their own extensions.

Robbt:
Well if this PR makes LibreTime support media RSS and that is currently needed then we can implement this. Just need some (media) RSS feeds to test it on.

JohnnyC1951:
I can't lay my hands on what I want right now,  but here is a messy feed with lots of different kinds of stuff in it. http://feeds.feedburner.com/MBCSermons

Robbt:
Ok, so I ran into an error caused by an image included as an enclosure on a feed on my stations site. I tried this code but it hid the episode completely from the Podcast episode view rather than allowing it to download the mp3. I'm going to revisit this.

Robbt:
So I got around to trying to figure out why this code wasn't working by doing my own derivative and I found that $item->getEnclosures() was only returning the first enclosure as per the bug reported here https://github.com/simplepie/simplepie/issues/585

So #675 isn't fixable unless we figure out why SimplePie is only returning one enclosure.

Robbt:
Yeah get_enclosures is in this PR, it is a way of reading and iterating over multiple enclosures. I was just trying to diagnose why this code wasn't working as described and this is based upon my testing why it ignored any enclosures beyond the first one. I'll see what I can pull together. I was working on a separate PR inspired by this one.

frecuencialibre:
replaced by #728

Issue #313, opened 2017-09-29
Libretime layout improvements


xabispacebiker:
I made a few (from my point of view) improvements to the interface.

![libretime-interface-mods](https://user-images.githubusercontent.com/1733884/31018327-f0a22e52-a52b-11e7-95a6-3deb46fb586b.png)

This screenshot shows the Host preferences page, who do not have access to the General Preferences.

https://github.com/xabispacebiker/libretime/tree/reconditioning-interface-personal-block

hairmare:
Awesome! These changes look great, moving the Logout button also solves the issue of the previous link disapearing on small screens in some locales. Do you plan on adding a UI for updating the logo or did I miss that looking at your branch?

xabispacebiker:
The logo is set in General Preferences.
![captura de pantalla 2017-09-30 a las 11 26 11](https://user-images.githubusercontent.com/1733884/31044571-309ad8a0-a5d2-11e7-9656-7b4f557becc4.png)

Robbt:
Hey @xabispacebiker do you want to do a PR for this ? I think it would be a good improvement for the LibreTime UI in general and the logo change would be a definite improvement. The only thing we might want to do is update the manual.

HammyHavoc:
I'd like to see this too. Though speaking of logos, would you be interested in a LibreTime logo for the project and to set as default to replace http://radio.voidancerecords.com/css/images/airtime_logo_med2.png? If so, let @bryandesrosiers know and he might be interested in helping. :- )

Cysioland:
@xabispacebiker Yeah, definitely a pull request would be nice

xabispacebiker:
@HammyHavoc , it would definitely be awesome if @bryandesrosiers could provide a default logo for the project. Hopefully, I can find some time this week to send the pull request.

Cysioland:
@xabispacebiker In absence of that, even a plain "LibreTime" text would be nice

xabispacebiker:
If we get the logo soon, then the wait is worth it.. but if it's gonna take long we might just add some default text, or just leave it empty until a station logo is configured in preferences.

PR #489, opened 2018-05-17
Reconditioning interface personal block

xabispacebiker:
This pullrequest makes some improvements to the interface as detailed in issue https://github.com/LibreTime/libretime/issues/313

It does not include a default logo as requested, but if anyone provides one then I can make the modifications to include it by default

### DIFF
--- a/airtime_mvc/application/views/scripts/partialviews/header.phtml
+++ b/airtime_mvc/application/views/scripts/partialviews/header.phtml
@@ -1,5 +1,7 @@
 <div id="master-panel">
-  <img class="logo" src="data:image/png;base64,<?php echo Application_Model_Preference::GetStationLogo(); ?>" />
+  <div class="logo-container">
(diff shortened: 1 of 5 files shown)


hairmare:
Hey @xabispacebiker 

Sorry for the late reply, I've been extremely busy lately.

I was really looking forward to this PR. I'll do some testing on it locally over the next couple of days before give it a merge...

I've also started prep for the next release so you can expect it to get released as soon as I'm through with RelEng.

Cheers
Lucas

airtime_mvc/application/views/scripts/partialviews/header.phtml
```diff
@@ -54,14 +56,6 @@
       <li class="time" id="time"></li>
       <li class="time-zone"><span id="time-zone"></span></li>
     </ul>
-    <div class="personal-block solo">
```
frecuencialibre: i think this caused https://github.com/LibreTime/libretime/issues/698

PR #501, opened 2018-07-31
Setup Reverse Proxy apache config.

Ark74:
I've been testing the reverse proxy setup in order to accomplish at easing a further fully working SSL config.
This PR is tested for Xenial and Trusty.

In case you'll like to see some other steps that I'd love to integrate on further development please check this [post-installation script](https://gist.github.com/Ark74/51e1ecb3da664473a1f8b1cf9d911f85) (also fully working with Xenial and Trusty) in order to have a working Reverse Proxy for Icecast section, and some other issues I've come across.

Cheers!


Robbt:
So I haven't tested this but is the basic idea that it sets up a radio1 radio2 and radio3 mp3 that can be accessed via SSL that point to the default stream names.

The only issue I see with this is that we currently allow people to customize the stream names and that would break this functionality.

The post-install script also seems good. Maybe we should just add most of this into the install script. Certainly the silan fix would be a good thing to automate. It might be good to get the debian files directly hosted on the libretime github and I also know for centos and some other distros @hairmare has created some packages using the opensuse distro platform. I didn't see them for trusty when I tested it out though so your packages might work for the Ubuntu distros as a default.

While we are talking about install script and you are working on icecast - somewhat related feature that we've noted but not worked on would be the ability to modify the icecast passwords from the web-based install and also have these passed onto the default values because I'm sure a lot of people end up with hackme because they don't know how to change it.

Ark74:
I've added the deb files to install supported versions of silan.
Does wheezy is still a supported platform?

Would it make sense to try to backport it?

Ark74:
I'm having issues uploading the correct binary for trusty on this repo :-/
Any how, I've come with a script to check and install/patch silan for *buntu/debian/centos7

Here it is:
https://gist.github.com/Ark74/7a86f28bd71cc9dad5fec9cb89a10d62

Cheers!

hairmare:
> I've been testing the reverse proxy setup in order to accomplish at easing a further fully working SSL config.

Icecast/Xiph does not recommend running icecast behind a proxy (ie. in [this medium post]( https://medium.com/@epirat/i-noticed-you-said-you-were-running-icecast-behind-a-reverse-proxy-nginx-which-is-strongly-8775db836c98) by one of the icecast devs).

Icecast can do SSL on it's own but it doesn't currently support reloading the TLS certificates on the fly without a restart. The [icecast-kh](https://karlheyes.github.io/) fork supports reloading certs on the fly and the upcoming icecast 2.5 release will support it as well. Reloading only becomes an issue if you want to use Let's Encrypt.

It looks like this kind of reverse proxy setup should work just fine, but i don't think it's a configuration that we can officially support (deferring to icecasts recommendations).

> Does wheezy is still a supported platform?

Nope, as per [the Debian project](https://wiki.debian.org/DebianWheezy) Debian Wheezy was EOLed 2018-06-01 when it stopped getting security patches. We got rid of "official" support at around the time Debian Stretch was released. It's more than likely that noone found time to remove Wheezy legacy code from the installer.

> I'm having issues uploading the correct binary for trusty on this repo :-/

We should not include distro specific binaries in our code repo.

> Any how, I've come with a script to check and install/patch silan for *buntu/debian/centos7 (can be tested on docker containers) 👍

Sweet script...

I just realized that a working silan is now in Debian stable: https://packages.debian.org/source/stable/silan is there any reason not to use that? I'll update the instructions in the release notes to reflect this when I do the next release.

Ark74:
> Icecast/Xiph does not recommend running icecast behind a proxy (ie. in this medium post by one of the icecast devs).
> 
> Icecast can do SSL on it's own but it doesn't currently support reloading the TLS certificates on the fly without a restart. The icecast-kh fork supports reloading certs on the fly and the upcoming icecast 2.5 release will support it as well. Reloading only becomes an issue if you want to use Let's Encrypt.

I'm quite new to Icecast, so I guess it would be worth to investigate more on the SSL capabilities.
Since the use of ssl on modern browsers is quite a requirement and preventing errors like Mixed content a must.

> Nope, as per the Debian project Debian Wheezy was EOLed 2018-06-01 when it stopped getting security patches. We got rid of "official" support at around the time Debian Stretch was released.

Cool, because there is no way to backport silan for it :laughing: 

> We should not include distro specific binaries in our code repo.
Oh well, the packages and scripts are already available if they are worth for something.

> Edit: Nope, still no 0.3.3 in stable.... It might be time to test the 0.3.3 in testing and make the installer ready for the next bump though.

As for Debian they have silan 0.4.0 on testing and for *buntu we have 0.3.0 'til Bionic (latest LTS) then 0.4.0.

I'll be closing this for now.
Cheers!

JohnnyC1951:
Just a quick note on https icecast proxy. This works with mountpoints defined in LT
Care must be taken else you always get the same IP for listeners in your stats.
This very simple Apache  site-available is working here, using a Certbot created cert. Just change your servername.
Sample working streams: 
[](https://ss.deprogrammedradio.com:8004/deprogrammedradio.mp3)
[](https://ss.deprogrammedradio.com:8004/deprogrammedradio.ogg)


`Listen 8004`
`<VirtualHost *:8004>`
`  ServerName ss.deprogrammedradio.com`
`  ServerAdmin hostmaster@domain.com`
`  Include /etc/letsencrypt/options-ssl-apache.conf`
`  SSLCertificateFile /etc/letsencrypt/live/ss.deprogrammedradio.com/cert.pem`
`  SSLCertificateKeyFile /etc/letsencrypt/live/ss.deprogrammedradio.com/privkey.pem`
`  SSLCertificateChainFile /etc/letsencrypt/live/ss.deprogrammedradio.com/chain.pem`
`  ProxyPreserveHost On`
`  ProxyPass / http://ss.deprogrammedradio.com:8000/`
`  ProxyPassReverse / http://ss.deprogrammedradio.com:8000/`
`  <Proxy *>`
`        Order deny,allow`
`        Allow from all`
`  </Proxy>`
`</VirtualHost>`

Issue #229, opened 2017-06-12
Problem: There is no easy way to "customize" the liquidsoap parts of LibreTime


hairmare:
Stations that use LibreTime may want to customize the liquidsoap layer to fit their exact needs. One example of this is replacing the default "silence" output with a fallback solution that can actually prevent dead air. Other use cases are doing some processing on the signal before it is sent to the encoder or updating stream metadata (artist, title, ...) in station specific ways.

As far as I can tell, we first talked about this in #194. https://github.com/LibreTime/libretime/issues/224#issuecomment-307750957 reminded me that it is still a thing.

## Proposed Solution

Make it easy to override the relevant parts of the provided liquidsoap scripts. I'm doing something similar in my badly named [dabplus-on-air-processing](https://github.com/radiorabe/dabplus-on-air-processing) repo. In that one you get to [choose the actual processor implementation](https://github.com/radiorabe/dabplus-on-air-processing/blob/1c467fae9da77b51df3755e5c77d8c81ad70ebf5/src/dabplus-on-air-processing.liq#L31-L44) in [config](https://github.com/radiorabe/dabplus-on-air-processing/blob/1c467fae9da77b51df3755e5c77d8c81ad70ebf5/dabplus-on-air-processing.conf#L6-L12). In that implementation the possible processors need to be hardcoded in in the liquidsoap script. Since `airtime-liquidsoap` generates a liquidsoap file anyway, doing something similar is more than trivial.

While this isn't the nicest architecture (it's hard to define a clear api since liquidsoap has no namespaces or anything non functional hacker might be used to) it can cover lot's of use cases. I should probably ask the liquidsoap mailing list for a recommendation on how else to solve this properly.

The downside of user changeable liquidsoap might be that it's hard to support (you never know if the error is in liquidsoap or the overrides). For that reason, I think we should implement this as an `airtime.conf` feature (like FreeIPA support). Power Users will appreciate it, others don't need to know it's there.

xabispacebiker:
+1

xabispacebiker:
Actually this is quite easy to implement. I have it done in a production server with airtime 2.5.2. I will try to find some time  to adapt it in the latest Libretime version and send a pull request.

xabispacebiker:
Silence in a FM radio station is bad, so bad, so let's discuss about this feature to make it real and usable for everyone. Let me tell you what we did and how we did it in the airtime installation running at 97irratia.info.

[...]

PR #505, opened 2018-08-27
Changes made for spaz radio [wip]

kenrestivo:
A year or two ago, someone from Libretime project asked if I would be able to share changes we made to support SPAZ Radio.

- A jukebox which plays random archive files when there are no scheduled shows
- Saves live shows and include metadata in the files saved
- Sends now playing information as JSON to our DJ Dashboard

Was finally able to whittle it down to a clean-ish patch, which is here. I don't think it is ready for application to other stations, but some of it might be useful to the project.

### DIFF
--- a/python_apps/pypo/liquidsoap/ls_lib.liq
+++ b/python_apps/pypo/liquidsoap/ls_lib.liq
@@ -1,3 +1,6 @@
+# coding: utf-8
+%include "utils.liq"
+
 def notify(m)
   command = "timeout --signal=KILL 45 pyponotify --media-id=#{m['schedule_table_id']} &"
   log(command)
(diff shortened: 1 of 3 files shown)


Robbt:
Cool. I think this is definitely useful. Let me look at it and see what approach towards merging makes the most sense. It might make sense to break the changes into individual PRs that each address the specific feature but this is a great starting point. Thanks for sharing.

Robbt:
So as @kenrestivo mentioned in his PR this isn't something we can really commit right now but instead a bunch of interesting modifications to the liquidsoap configuration that adds the features he mentioned above. So what we should do in order to add this is create an issue that documents the problem that the code will fix and then create a corresponding PR to go along with those specific changes rather than committing this. It looks like there are 3 different potential issues above.

This is also relevant to #229 as it would be ideal if we supported some kind of modular add-on to our liquidsoap config that would allow people to leverage the power of solutions like this without requiring hacking the liquidsoap config directly or adding specific hacks (such as the autoDJ) to every LT liquidsoap config.

Issue #70, opened 2017-03-12
Problem: Watched folders for automatic import/filesystem based import not included in analyzer.


Robbt:
I believe this functionality is missing from analyzer but the ability to bulk import tracks via FTP or the filesystem is pretty useful especially for people starting to use LibreTime from a different automation system. This was called the watched folders feature.

hairmare:
There is the [ftp upload hook](https://github.com/LibreTime/libretime/blob/a583d8fa1e9ceaa8d9a37a9043262a8452914869/python_apps/airtime_analyzer/tools/ftp-upload-hook.sh) that might replace parts of the watched folder feature and I also found `airtime-import` in utils which seems to be somewhat related.

hairmare:
I have the UI changes prepared on https://github.com/LibreTime/libretime/compare/master...radiorabe:feature/restore-media-folders-ui.

The calls to media-monitor that need to get changed to analyzer are on [airtime_mvc/application/models/MusicDir.php#L296](https://github.com/LibreTime/libretime/blob/cb4ad730fc9ccb342bb46a26c5074800957494e4/airtime_mvc/application/models/MusicDir.php#L296) and [#L383](https://github.com/LibreTime/libretime/blob/cb4ad730fc9ccb342bb46a26c5074800957494e4/airtime_mvc/application/models/MusicDir.php#L383).

I started looking at media-monitor but didn't get very far, may re-implementing this in analyzer greenfield style makes more sense.

samcarmex:
Just wondering if there's any update on this? Monitoring a network mounted folder is now impossible without the airtime-import function working. We store all our audio content on a separate server (same local network though) and simply 'import' the files into Airtime without having to copy them to a second folder. This can't be done with Libretime (AFAIK).

Are there any plans to implement a true watched folder functionality or will that not be a priority?

Robbt:
Work has been done under #111 but I'm not sure what state it is in. Either the original contributor @HaJoHe will need to finish the work or someone else will need to do whatever is left that needs to be done to bring this functionality back. I wouldn't say it's not a priority but it hasn't been on the top of my todo list at the moment.

HaJoHe:
Hi,
all works fine for me. So, if there are any modifications needed, I'll try to do it.

Robbt:
[...]

PR #514, opened 2018-09-09
[WIP] Feature: Libretime automatic import from filesystem directory

Robbt:
I still have a fair amount to do with this in terms of getting it cleaned up, but as an alternative implementation of a fix for #70 - this creates a python process that will monitor a folder in the filesystem currently /srv/airtime/stor/uploads and then import any file copied or moved into this folder. It also deletes the file immediately afterwards.

Here is some of my thinking behind it
Libretime_import 

to operate - needs a directory to watch
needs libretime host, port, basedir, api-key or could be done with user authentication for remote user setup

import folder location should be configurable during the libretime install

ideally wouldn't import duplicate files (would need way to look up existing file in database via track title and/or md5) or files that were being written (could be done by lsof) with pyinotify or watchdog

in theory would be cool if it could be ran remotely or even cross-platform

minimal code needed - 
pull config vars from /etc/airtime/airtime.conf
pull import_dir from database or parameter or airtime.conf

ways it could run:
scan every file in directory and subdirectories, import them via rest call and then (leave them intact, move them to new directory, or delete them)

watch for new files only and then import them and (move or delete them) or keep them intact

Issues with keeping them in place, don't want to re-import them again in the future (need duplicate checking and that will consume resources) 

Issues with moving them - need a folder to move them to that isn't scanned, takes up double space for every file imported, is of questionable value (just to show non-destructive status of the import)

Issues with deleting them - if someone chose the wrong directory ie a working and organized music directory it could be destroy the organization of their files - perhaps not giving them the option and just making it work with /stor/import by default would be a good safeguard, a setting to change this with warnings could be available under the admin settings page.

The copy based import makes sense as a one-off script that could be ran during the install or triggered via a request to the UI. (not sure best approach here as we want to prevent user mistakes with duplicate files)

The simplest approach for now is to simply setup a watched folder under /srv/stor/uploads and then ignore existing files and watch new files copied here and send them to the import process via the UI and delete them from this directory.


Robbt:
Ok, so anyone checking this out - at the moment it is working with the folder /srv/airtime/store/uploads - which will be created upon installation and will automatically import and delete any audio files and folders contained within it.

I still need to get it working as a daemon with upstart, init.d and systemd - I copied the code and modified what was in airtime_analyzer's install but it doesn't appear to start when I try to start it. I might get around to doing that later today as getting it to run as a standalone service is my goal.

It also is ignoring Ctrl-C and needs to be killed via killall libretime_import - which is probably an annoying result of using the libretime_analyzer as a model.

But if anyone wants to try this they could install it via python setup.py install from inside the directory and then run it with sudo -u www-data libretime_import - they will also need to make sure they add an import_dir = /srv/airtime/stor/uploads/ to /etc/airtime/airtime.conf under the general header and create that directory owned by www-data and it should work.

Robbt:
Ok, did a little bit more work on this and I think it should be ready to go in terms of installation etc.
I could add the option to configure the import directory to the media directory portion of the setup script, but I managed to get it working as a standalone service and it ready for testing.

xabispacebiker:
This looks very promising !! I will install it in my development environment and give it a try

Robbt:
So here are the todos revealed through testing

- The following TODOs are related to it not installing properly w/o the ./install script
- [ ] - The init-scripts are only installed through the airtime installer
- [ ] - It depends upon the ./install script to add required setting to /etc/airtime/airtime.conf (import_dir)
- [ ] - It needs the log created or else it crashes /var/log/libretime_import.log (done by installer)
- [ ] - it will crash if the /srv/airtime/stor/uploads directory isn't created before it is ran
- These are just in general TODOs
- [ ] - we agreed to use - vs _ so need to change everything to libretime-import vs. libretime_import
- [ ] - there needs to be documentation added to explain how this works
- [ ] - if you happen to start the program manually by running /usr/bin/libretime_import it ignores CTRL-C (needs to be killed externally)
- [ ] - FIX my last commit to not include everything that happened since I started doing this PR. (should have only been 1 simple file addition but evidently I made a mistake

I'll add any addition bugs revealed through training - thanks @frecuencialibre for helping me figure out the install situation wasn't working

frecuencialibre:
so we should test this one again once you're able to make progress on the todo lists above? 

as for cleanup, the option that occurs to me is to create a new branch (!! from the last stable commit!! eg `git branch feature/libretime_import2 7beadd05b004024d2513408391ac9f336a5ee7fe`) and then cherry pick the commits you do want.

hairmare:
@Robbt Don't fret about the cleanup... I'm not quite sure why this regularly happens on your PRs, maybe it's something to do with git pulling the master into your branch (I usually git rebase and force push). Could you share you workflow regarding these?

If you use the cherry picking route proposed by @frecuencialibre you can then force push the new branch onto an existing PR with something like `git push origin cherrypickbranch:oldbranch --force` to replace the PR contents.

I usually recommend using a ui like `gitk` to visualize complexer scenarios before pushing, but i've used it on PRs with this exact issue and haven't been able to grok why they exist.

Robbt:
Yeah honestly I typically just re-check out my branch and do a pull and then commit whatever and push it back and it seems to push the rest of the stuff along with it into my PR. I need to get a better understanding of how to use rebase and to avoid this but I'll try and do that on a practice repo. I think that if I reset my codebase to my last commit that might also keep things from travelling along (but I haven't tried that yet). Basically my git skills beyond the basics have all been developed while working on this project.

Robbt:
I also figured out that if you include the logs and systemd files in the data_files then this will make them installable without crashing upon creation. I looked and the existing python module setup.py have only systemd and upstart files and so probably would benefit from adding the systemd files as well. Our install ignores this part of it and does it manually as per the --ignore-init-script flag.

Robbt:
So I resolved most of the install issues. I didn't write anything to parse the config file and add the upload_dir section, instead I just set it up to use the default directory.

I also decided to make /srv/airtime/stor/uploads world writeable for a few reasons.
The first being I couldn't figure out an easy way to determine the apache user to chown it too that would be consistent between different linux distributions and second having it world writeable would make it easier for people to put files into it without needing to understand the way permissions work within linux.

On the other hand it does potentially open an avenue to abuse for people that are running this on a shared host or otherwise providing shell access to people who aren't directly connected to libretime but since running libretime on anything other than a dedicated host isn't recommended and the permissions and directory can be changed by someone who has basic knowledge of linux I don't think it is a big deal.

Also if we say kept the permissions owned by www-data this kind of requires either sudo access or some kind of chown to give a regular user access anyways so that didn't make too much sense.

The next logical step would be for a user to provide FTP access to this directory but I suspect that would be better handled through documentation and a tutorial than something we want to support out of the box. 

The only remaining outstanding issues are the startup scripts now installed by setup.py won't work if you leave it in /usr/local/bin so you need to run the ```sudo python setup.py install --install-scripts=/usr/bin ``` if you are manually installing it. Not sure if we should fix this or just add it to the documentation. Considering most people will be installing it via the ./install option in the first place. The work I did is more or less intended for people already running LibreTime who want to run setup.py manually to either test it or add it to their pre-existing LibreTime instance.

In addition I'd like to add some documentation before we merge this and I probably should change the command to be libretime-import vs. libretime_import which is a simple enough replace once I figure out the sed command or how to refactor it with my PyCharm IDE.

But in general I think this is actually ready for testing at this point.

Robbt:
Ok, I have added documentation and I am going to be removing the wip flag because I think everything is working. I'm going to hold off on the name change from libretime_import to libretime-import and we can do that as part of our refactoing of airtime_analyzer and renaming things to libretime-analyzer etc. 

But I'm willing to do that now if people think it's import. The docs could also probably use some more tweaking but I wanted to get something initially out. I also overwrote the link to the previous automatic file import section w/o removing that section from the site.

frecuencialibre:
again, monday i'll come back to this with a decent internet connection and will try to remove exterior wierd setup factors. thanks @Robbt !

Robbt:
Ok, hmm, you could try running ```sudo -u www-data libretime_import``` directly form the command line but be prewarned you won't be able to hit CTRL-C to kill the process and will need to open up another terminal to do this. This will at least show you if the process dies immediately and will also provide you with some idea. 

Also did you copy files into uploads prior to running the process ? This is what I suspect you might have done. The libretime_import service doesn't scan pre-existing files but actually watches the directory for new files and the imports them. The idea being it will always be a empty directory that just receives files and funnels them to the airtime_analyzer - I didn't bother to write code to go in and suck up files that already existed because in theory they shouldn't exist and this will also potentially prevent someone from setting an upload_dir thinking it works like a watched_folder and then having their original files and directories deleted and sucked into libretime.

frecuencialibre:
so i did a fresh install. 

here's the relevant part of the install script output:

```
/opt/libretime/python_apps/libretime_import
running install
running bdist_egg
running egg_info
writing requirements to libretime_import.egg-info/requires.txt
writing libretime_import.egg-info/PKG-INFO
writing top-level names to libretime_import.egg-info/top_level.txt
writing dependency_links to libretime_import.egg-info/dependency_links.txt
reading manifest file 'libretime_import.egg-info/SOURCES.txt'
writing manifest file 'libretime_import.egg-info/SOURCES.txt'
installing library code to build/bdist.linux-x86_64/egg
running install_lib
running build_py
copying libretime_import/folder_watcher.py -> build/lib.linux-x86_64-2.7/libretime_import
copying libretime_import/__init__.py -> build/lib.linux-x86_64-2.7/libretime_import
copying libretime_import/libretime_import.py -> build/lib.linux-x86_64-2.7/libretime_import
copying libretime_import/config_file.py -> build/lib.linux-x86_64-2.7/libretime_import
creating build/bdist.linux-x86_64/egg
creating build/bdist.linux-x86_64/egg/libretime_import
copying build/lib.linux-x86_64-2.7/libretime_import/folder_watcher.py -> build/bdist.linux-x86_64/egg/libretime_import
copying build/lib.linux-x86_64-2.7/libretime_import/__init__.py -> build/bdist.linux-x86_64/egg/libretime_import
copying build/lib.linux-x86_64-2.7/libretime_import/libretime_import.py -> build/bdist.linux-x86_64/egg/libretime_import
copying build/lib.linux-x86_64-2.7/libretime_import/config_file.py -> build/bdist.linux-x86_64/egg/libretime_import
byte-compiling build/bdist.linux-x86_64/egg/libretime_import/folder_watcher.py to folder_watcher.pyc
byte-compiling build/bdist.linux-x86_64/egg/libretime_import/__init__.py to __init__.pyc
byte-compiling build/bdist.linux-x86_64/egg/libretime_import/libretime_import.py to libretime_import.pyc
byte-compiling build/bdist.linux-x86_64/egg/libretime_import/config_file.py to config_file.pyc
creating build/bdist.linux-x86_64/egg/EGG-INFO
installing scripts to build/bdist.linux-x86_64/egg/EGG-INFO/scripts
running install_scripts
running build_scripts
copying and adjusting bin/libretime_import -> build/scripts-2.7
creating build/bdist.linux-x86_64/egg/EGG-INFO/scripts
copying build/scripts-2.7/libretime_import -> build/bdist.linux-x86_64/egg/EGG-INFO/scripts
changing mode of build/bdist.linux-x86_64/egg/EGG-INFO/scripts/libretime_import to 755
copying libretime_import.egg-info/PKG-INFO -> build/bdist.linux-x86_64/egg/EGG-INFO
copying libretime_import.egg-info/SOURCES.txt -> build/bdist.linux-x86_64/egg/EGG-INFO
copying libretime_import.egg-info/dependency_links.txt -> build/bdist.linux-x86_64/egg/EGG-INFO
copying libretime_import.egg-info/not-zip-safe -> build/bdist.linux-x86_64/egg/EGG-INFO
copying libretime_import.egg-info/requires.txt -> build/bdist.linux-x86_64/egg/EGG-INFO
copying libretime_import.egg-info/top_level.txt -> build/bdist.linux-x86_64/egg/EGG-INFO
creating 'dist/libretime_import-0.1-py2.7.egg' and adding 'build/bdist.linux-x86_64/egg' to it
removing 'build/bdist.linux-x86_64/egg' (and everything under it)
Processing libretime_import-0.1-py2.7.egg
creating /usr/local/lib/python2.7/dist-packages/libretime_import-0.1-py2.7.egg
Extracting libretime_import-0.1-py2.7.egg to /usr/local/lib/python2.7/dist-packages
Adding libretime-import 0.1 to easy-install.pth file
Installing libretime_import script to /usr/bin

Installed /usr/local/lib/python2.7/dist-packages/libretime_import-0.1-py2.7.egg
Processing dependencies for libretime-import==0.1
Searching for pyinotify==0.9.6
Best match: pyinotify 0.9.6
Processing pyinotify-0.9.6-py2.7.egg
pyinotify 0.9.6 is already the active version in easy-install.pth

Using /usr/local/lib/python2.7/dist-packages/pyinotify-0.9.6-py2.7.egg
Searching for requests==2.20.1
Best match: requests 2.20.1
Processing requests-2.20.1-py2.7.egg
requests 2.20.1 is already the active version in easy-install.pth

Using /usr/local/lib/python2.7/dist-packages/requests-2.20.1-py2.7.egg
Searching for python-daemon==1.6
Best match: python-daemon 1.6
Processing python_daemon-1.6-py2.7.egg
python-daemon 1.6 is already the active version in easy-install.pth

Using /usr/local/lib/python2.7/dist-packages/python_daemon-1.6-py2.7.egg
Searching for mock==2.0.0
Best match: mock 2.0.0
Processing mock-2.0.0-py2.7.egg
mock 2.0.0 is already the active version in easy-install.pth

Using /usr/local/lib/python2.7/dist-packages/mock-2.0.0-py2.7.egg
Searching for coverage==5.0a4
Best match: coverage 5.0a4
Processing coverage-5.0a4-py2.7-linux-x86_64.egg
coverage 5.0a4 is already the active version in easy-install.pth
Installing coverage2 script to /usr/bin
Installing coverage-2.7 script to /usr/bin
Installing coverage script to /usr/bin

Using /usr/local/lib/python2.7/dist-packages/coverage-5.0a4-py2.7-linux-x86_64.egg
Searching for nose==1.3.7
Best match: nose 1.3.7
Processing nose-1.3.7-py2.7.egg
nose 1.3.7 is already the active version in easy-install.pth
Installing nosetests script to /usr/bin
Installing nosetests-2.7 script to /usr/bin

Using /usr/local/lib/python2.7/dist-packages/nose-1.3.7-py2.7.egg
Searching for file-magic==0.4.0
Best match: file-magic 0.4.0
Processing file_magic-0.4.0-py2.7.egg
file-magic 0.4.0 is already the active version in easy-install.pth

Using /usr/local/lib/python2.7/dist-packages/file_magic-0.4.0-py2.7.egg
Searching for daemon==1.2
Best match: daemon 1.2
Processing daemon-1.2-py2.7.egg
daemon 1.2 is already the active version in easy-install.pth

Using /usr/local/lib/python2.7/dist-packages/daemon-1.2-py2.7.egg
Searching for urllib3==1.24.1
Best match: urllib3 1.24.1
Processing urllib3-1.24.1-py2.7.egg
urllib3 1.24.1 is already the active version in easy-install.pth

Using /usr/local/lib/python2.7/dist-packages/urllib3-1.24.1-py2.7.egg
Searching for idna==2.7
Best match: idna 2.7
Processing idna-2.7-py2.7.egg
idna 2.7 is already the active version in easy-install.pth

Using /usr/local/lib/python2.7/dist-packages/idna-2.7-py2.7.egg
Searching for chardet==3.0.4
Best match: chardet 3.0.4
Processing chardet-3.0.4-py2.7.egg
chardet 3.0.4 is already the active version in easy-install.pth
Installing chardetect script to /usr/bin

Using /usr/local/lib/python2.7/dist-packages/chardet-3.0.4-py2.7.egg
Searching for certifi==2018.10.15
Best match: certifi 2018.10.15
Processing certifi-2018.10.15-py2.7.egg
certifi 2018.10.15 is already the active version in easy-install.pth

Using /usr/local/lib/python2.7/dist-packages/certifi-2018.10.15-py2.7.egg
Searching for lockfile==0.12.2
Best match: lockfile 0.12.2
Processing lockfile-0.12.2-py2.7.egg
lockfile 0.12.2 is already the active version in easy-install.pth

Using /usr/local/lib/python2.7/dist-packages/lockfile-0.12.2-py2.7.egg
Searching for setuptools==40.6.2
Best match: setuptools 40.6.2
Adding setuptools 40.6.2 to easy-install.pth file
Installing easy_install script to /usr/bin
Installing easy_install-3.6 script to /usr/bin

Using /usr/local/lib/python2.7/dist-packages
Searching for six==1.10.0
Best match: six 1.10.0
Adding six 1.10.0 to easy-install.pth file

Using /usr/lib/python2.7/dist-packages
Searching for pbr==5.1.1
Best match: pbr 5.1.1
Processing pbr-5.1.1-py2.7.egg
pbr 5.1.1 is already the active version in easy-install.pth
Installing pbr script to /usr/bin

Using /usr/local/lib/python2.7/dist-packages/pbr-5.1.1-py2.7.egg
Searching for funcsigs==1.0.2
Best match: funcsigs 1.0.2
Processing funcsigs-1.0.2-py2.7.egg
funcsigs 1.0.2 is already the active version in easy-install.pth

Using /usr/local/lib/python2.7/dist-packages/funcsigs-1.0.2-py2.7.egg
Finished processing dependencies for libretime-import==0.1
Created symlink /etc/systemd/system/multi-user.target.wants/libretime_import.service,pointing to /etc/systemd/system/libretime_import.service.
```

at first the libretime_import service wasn't even in the list generated by `service --status-all`. so i tried: 

```
root@36c263b86fc2:/usr/share/airtime/php/python_apps/libretime_import# python setup.py install --install-scripts=/usr/bin/usr/share/airtime/php/python_apps/libretime_import
```

and then saw the service present in the list, but not running. then: 

```
root@36c263b86fc2:/usr/share/airtime/php/python_apps/libretime_import# initctl reload-configuration; sudo service libretime_import restart
Restarting libretime_import: Done.
root@36c263b86fc2:/usr/share/airtime/php/python_apps/libretime_import# service --status-all
 [ + ]  apache-htcacheclean
 [ + ]  apache2
 [ - ]  bootmisc.sh
 [ - ]  checkfs.sh
 [ - ]  checkroot-bootclean.sh
 [ - ]  checkroot.sh
 [ + ]  cron
 [ - ]  dbus
 [ - ]  hostname.sh
 [ ? ]  hwclock.sh
 [ - ]  icecast2
 [ - ]  killprocs
 [ + ]  libretime_import
[...]
```

i then copied a file into `/srv/airtime/stor/uploads`, **and it worked!** the file imported and was available in the LT web interface.

```
root@36c263b86fc2:/var/log/airtime# cat libretime_import.log 
2018-11-27 16:55:28,804 [folder_watcher] [INFO ]   Watching /srv/airtime/stor/uploads folder for new files...
2018-11-27 16:59:31,725 [folder_watcher] [INFO ]  This file was written to the import directory and will be uploaded: /srv/airtime/stor/uploads/ALR Elite/022 En pie de lucha_Olmeca.mp3
```

i'm not sure why the service didn't start immediately after the install. long-running services such as playout are managed in  @ned-kelly 's multicontainer instance using Supervisor. there's probably work that should be done in the docker install to also include this new `libretime_import` service, but i unfortunately won't have time to get into that until next month when i'm planning on trying again to get the multicontainer repo under the LibreTime user, and benefiting from improvements we're making upstream in this main LT repo. nonetheless hope this helps a little!

Robbt:
Ok, yeah the services won't automatically start with the install, just as we need to manually start the various other python services we should add libretime_import to the service.
I'll check and see if it was being added but I suspect that the second install was unnecessary as it should have been ran by the ./install and installed to /usr/bin/libretime_import and if it wasn't the systemd wouldn't have worked at all. But perhaps the install isn't copying the files correctly because we have been doing the --no-init-scripts and manually doing things I think as a result of a legacy of some SAAS workaround they did so perhaps we don't need to that anymore.

Thanks for testing, looks like there are a few more tweaks to make sure everything is done.
- [ ] Tell user to start libretime_import during web install
- [ ] Make sure install script creates the appropriate systemd files and copies libretime_import into appropriate directory

Robbt:
Ok, so I tested this on a fresh Ubuntu 18.04 system and it worked and the instructions to start it were already included. I did find a bug with the upstart config in the install and I fixed that. I think it is working.

frecuencialibre:
is the idea that libretime_import start by default and always run? if so could it be started automatically?

is it necessary to manually run `systemctl daemon-reload` and `initctl reload-configuration` and then restart the service? 

of course the simpler we can make users' experience the better.  making progress! and then the final step would be an OK from @hairmare ?

Robbt:
So testing this with @frecuencialibre and running into an issue regarding the permissions related to running the service as www-data. It works for individual files but seems to fail for directories with permission denied even if the folders and internal files are all chown to be owned by www-data and have 777.

Running it as a regular user that owns the files and uploads directory solves this problem. So I'm kind of stumped. Will test further.

This also highlights how permissions might be an issue if people don't know their way around a command prompt. 

Running it as root solves the problem but I think this would be a potential security issue that we wouldn't want to use in production. It would solve the permission issue though. What do other people think ?

Also the folders are left as relics after the import happens, so ideally it would go through and clean up the folders.

Robbt:
So I decided to make this service run as root because that is the most expedient fix to the issue I was running into and because it would prevent issues where people copy files or folders into the uploads directory without understanding the way permissions work in Linux.

The question is whether this is a security risk and if we should avoid it. I don't see a lot of attack vectors by default that this would add but it depends upon pynotify which depends upon inotify. So @hairmare you are usually pretty astute about system decision such as this, do you think it's acceptable to run this as root or should we try to figure out a way to avoid that ?

Robbt:
Ok, I  realized it still had some potential rough edges in terms of empty folders being left after import and so I added some code to delete them and also remove the watch so that we don't exceed the inode watch limit with really big imports. 

It could use some more testing but I think it is operational. I tested it under docker and was able to get it working.

Robbt:
Yeah the running it as www-data was that it couldn't read directories even if they were owned by www-data and given 777 for permission.

From a usability perspective it would be good if it could always read any folder or file copied into it. This can only happen with it running as root. If we run it as www-data then non-root users won't be able to copy folders into the directory. Unless they set them as 777 ahead of time. So we could do a permission check and see if we can read the file and then give an error for that. But if the idea is a filesystem based import directory that automatically ingests anything put into it I don't see a way to have it not require elevated permissions on the part of the user or the service. As far as I know about Linux filesystem permissions there is no way to say anything copied into this directory will be set writeable by a specific user. But those are the reasons I think as designed it needs to run as root. If we say wanted to integrate a special uploads user account that people could connect to via ssh that might be another option and we could run the service as this special user.

I do understand that running services as root is something that should generally be avoided. 

I guess integrating ftp might actually be a good idea too as if we are honest about this most users won't be comfortable using the CLI too copy files via ssh and so this is probably going to be used primarily by an admin and might even make more sense as a one off import script ala airtime-import without making it have a logical interface for non technical users.

ned-kelly:
> i'm not sure why the service didn't start immediately after the install. long-running services such as playout are managed in @ned-kelly 's multicontainer instance using Supervisor. there's probably work that should be done in the docker install to also include this new `libretime_import` service, but i unfortunately won't have time to get into that until next month when i'm planning on trying again to get the multicontainer repo under the LibreTime user, and benefiting from improvements we're making upstream in this main LT repo. nonetheless hope this helps a little!

Yep - this just needs to be added into Supervisor on the docker build - I've done some work on the docker build this week (such as building in Silan to 0.4.0, Liquidsoap, etc) - I might merge this feature into a branch of the docker build for anyone wanting to test until the permissions that was pointed out by hairmare is changed and this becomes one of the main services.

Robbt:
So as per my comment yesterday I'm going to recommend we hold off on introducing this as a new service into LibreTime until the use case can be better clarified and the service refined to meet that. 

It might also make more sense to simply roll the python code I wrote into airtime_analyzer to avoid creating a whole new service if it isn't necessary. and then making libretime-import the equivalent of airtime-import a CLI tool that works to import files and folders.

I can see this service being useful if it was tied together to a docker watched folder and automatically inserted any files that were copied into that but I'm not sure if this would work in terms of propogating the inotify events from host to docker instance. It won't work with a [windows host](https://github.com/docker/for-win/issues/56) - but that is a whole separate theoretical issue.

It would probably be better than nothing if we can get it working w/o using root and perhaps we could make it work with a user that can SFTP into the uploads directory and so a FTP client could also upload folders for automatic import. This would most likely be more useful than requiring the end user to have elevated privileges to '''sudo cp files /srv/airtime/stor/uploads''' - the other idea of setting up virtual FTP that automatically imports to a designated users files could also make sense but would require supporting a FTP server that authenticated by libretime_auth and modifying the code to import the owner based upon who "logged in".

frecuencialibre:
My vote is for this to be a one off command rather than a long-running service. why?
- a command could be run as root, thus getting us past permissions problems which can be very frustrating for folks without linux experience
- success/failure feedback could be shown to the user immediately right in their terminal. no more digging through multiple log files
- i agree that this is a function that will be used exclusively by admin. normal users can use the web ui. 
- i recently lost time because pynotify wasn't being triggered by `rsync`, which i only discovered much later by accident.

frecuencialibre:
it would be great to move files that fail to import to another folder, ideally preserving folder hierarchy.

Robbt:
So helping @frecuencialibre troubleshoot this we found out that it is failing for files that contain utf8 or non-ASCII characters in the filename. I found it erroring out on [this change](https://github.com/LibreTime/libretime/blob/634ac9c8463ad62860750ddd579356a19fcd98ab/airtime_mvc/application/modules/rest/controllers/MediaController.php#L132-L137) from #3 Zend_File_Transfer and printing 'invalid file'.

So not sure what the fix or issue is perhaps @hairmare has more insight since it says that this error shouldn't happen in the code.

Robbt:
Ok, so I fixed this by using urllib.pathname2url to wrap the pathname. So hopefully it'll also solve the issue @frecuencialibre had.

Robbt:
quick update, while this last "fix" did move things further along, ie bypassing the Zend_File_Transfer->isValid check - Mutagen borked when it received the filename and in addition the file names were urlencoded and the original UTF8 was lost. I spent a few hours trying to figure out a way to get PHP to reconvert or play nice with the urlEncoded files but no luck thus far.

Robbt:
Yeah, so this will work out of the box with files that are only ASCII filenames but UTF8 filenames require the urllib3 library to be installed from my PR. I merged it with my master and hopefully I can help steward it into the main urllib3 library.

Anyways anyone who wants to use it can do so `git clone https://github.com/Robbt/urllib3.git` into your python_apps folder and then run `sudo pythone_apps/urllib3/python3 setup install` and it'll install the newest version.

You probably need to rerun the setup for libretime_import by `sudo python3 setup.py install --install-scripts=/usr/bin` in the libretime_import directory.

But yeah, might end up refactoring this and/or just using curl for a CLI based app. And still need to troubleshoot the permission issues.

ned-kelly:
> I can see this service being useful if it was tied together to a docker watched folder and automatically inserted any files that were copied into that but I'm not sure if this would work in terms of propogating the inotify events from host to docker instance. It won't work with a windows host - but that is a whole separate theoretical issue.

Based on my experience you're gonna have issues for anyone running on Docker for Windows/Mac unfortunately inotify does not play that nice when working from within a docker container (based on experience) and tends to use a lot of resources on the host machine if you're got a reasonable folder size / file count ..

It may be better to make a command line import type script - This is how (Funkwhale)[https://funkwhale.audio/] Not a radio solution but more a self hosted Spotify solution do it - They are also using a lot of Python code behind the scenes.

Robbt:
Well yeah we are running into issues testing with @frecuencialibre and I think he is using Linux yet it seems to stop after a few hundred files are imported. My suspicion is that the pynotify can't keep up with the inotify results being sent as it just seems to lose them. So yeah this is evidently not the right way to do the bulk import of tracks. A command line tool that uses curl probably makes more sense. pynotify was used extensively in media-monitor though and perhaps could of been causing some of the instability issues that app experienced, it's been so long since we used it as a project that I don't know.

In general this would work as an alternative way to bulk import a reasonable # of tracks but not large libraries. And so I think it might be worth going back to the work I did trying to get #111 working with our current HEAD. It just needs a fair amount of cleaning up as it was more of a proof of concept that a finished module we'd want to ship. It built its own import stack that used parts of analyzer vs. modifying analyzer. It's kind of uncharted territory.

frecuencialibre:
> I think he is using Linux

yes. host running debian 9, multicontainer docker build with ubuntu 16.

> stop after a few hundred files are imported

yup. luckily the files that were _not_ imported remained in `/srv/airtime/stor/uploads/`, so i was able get past this short term pinch by looping a workaround dozens of times:
```
rsync -av /srv/airtime/stor/uploads/ /external-media/round2 --delete
rm -Rf /srv/airtime/stor/uploads/* && cp -R /external-media/round2/* /srv/airtime/stor/uploads
```

more notes for the record here:
- files with capitalized suffixes didn't import
- in order to use Robbt's upstream fix to urllib3 we had to do run the commands below. note that the `/opt...` paths work for ned's multicontainer build, but will probably need changing if used elsewhere. also the `killall` command at the end stops the process managed by Supervisor, and then it is automatically restarted. if not using supervisor you'll need to restart it manually.
```
cd /opt/libretime/python_apps/ && git clone https://github.com/Robbt/urllib3.git
pip uninstall urllib3
pip install -e /opt/libretime/python_apps/urllib3/
cd /opt/libretime/python_apps/urllib3/ && python3 setup.py install
rm -R /usr/local/lib/python3.5/dist-packages/urllib3-1.24.1-py3.5.egg/
killall libretime_import
```

hairmare:
>  A command line tool that uses curl probably makes more sense.

So... [this](https://github.com/LibreTime/libretime/blob/5b822afe14046157a8d9fbdff103eb383673e6be/python_apps/airtime_analyzer/tools/ftp-upload-hook.sh)? It doesn't do any inotify things but AFAICT you can run it on individual files and they get uploaded... Maybe something like this:
```bash
find /path/to/files -name '*.mp3' -exec python_apps/airtime_analyzer/tools/ftp-upload-hook.sh {} \;
```
It could even be possible to call this via [systemd.path](https://www.freedesktop.org/software/systemd/man/systemd.path.html).

Robbt:
As far as a command line tool, I think it would be helpful to write something structured like airtime-import which did more than just upload tracks. It included basic help etc and also made it possible to manage watched folders. 

As far as the extension goes, I had code checking to see if a file was a mp3 or a MP3 but perhaps the uppercase switch wasn't matching for some reason.

It's hard to diagnose exactly what was failing based upon the information but I'm suspecting that inotify is probably not the right solution for bulk-importing from the command line.

Robbt:
So I'm not sure what is salvageable from this module. It appears that there are some issues that would need to be resolved when "mass importing" tracks with this and the issue with the requests library and utf-8 would need to be fixed before this could be suggested as a filesystem based mass import. As it stands right now it might be useful connected with a FTP upload process as documented in http://libretime.org/manual/automated-file-import/

Robbt:
So my PR was finally accepted by urllib3 so this could actually work for utf8 files without requiring a forked copy of the upstream library.

kmahelona:
Kia ora. I've just started looking through these watch folder issues and alternatives in terms of bulk importing of media. In one of the comments above the "use case" was mentioned. We've been running our terrestrial radio 24/7 in the cloud on Airtime 2.5.1 for more than 3 years now (we playout locally using a raspberry pi and the airtime stream is a channel on our sound desk).

We use watch folders to upload news for every hour. We've made this "work" based on Airtime 2.5.1. Here's how it happens:
1. I have a script that downloads the news for each hour to a watch folder e.g. `/srv/airtime/watch_folder/news/`
2. That news is saved in a watch folder e.g. `1PM News.mp3` (I also write some metadata to it)
3. In the Airtime Database - every time I update that file with the same name - even through the audio changes (and therefore the md5) - Airtime sees it as the same file - just modified.
4. #3 above means that I have a playlist with the `1PM News.mp3` scheduled to play at the top of the hour - and every day at the top of the hour I have a refreshed news item playing.

This is critical for us - we also produce our own local news at play that at the half hour. One "gotcha" here is that *IF* the file is updated *AFTER* the show is scheduled (and this is always the case) and the duration of your audio file changes, you'll either cut your news short OR you'll play dead air. The news at the hour is always 6 minutes long so that's not a problem. Our news, however, is never the same length (for a number of reasons). I've even done something like speed up or down the audio by up to 5% to fit it to a fixed length.

So Airtime had its issues with "how do i play a news item every hour without requiring a human to manually upload and schedule it" but we were able to make that work. Airtime.pro ... didn't seem to have this feature (I could create a cron job on their cloud hosted services) so this is why we are still on 2.5.1.

So we definitely need the ability to automatically update a media item in the LibreTime database programatically. Watch folders is how we do that in 2.5.1, but if there's a better way with LibreTime 3.0 I'm happy to do that. Our organisation is also willing to pay for someone's time or do a one time large donation to the project to get this done because it's the only thing preventing us from upgrading to LibreTime - and the auto playlist scheduling in shows is something we desperately need.

I guess in the meantime I might install this branch of LibreTime and see if I can get something working and help by submitting any bugs.

kmahelona:
On a separate note @Robbt I've pulled in this branch into our fork because we want to see if we can get your upload folder feature implemented. I've ran into some python issues that I thought was worth sharing.

```bash
# command: ./install -f
# Detected distribution description: Ubuntu 18.04.4 LTS
Using /usr/local/lib/python2.7/dist-packages/lockfile-0.12.2-py2.7.egg
Searching for setuptools==44.0.0
Best match: setuptools 44.0.0
Adding setuptools 44.0.0 to easy-install.pth file
Installing easy_install script to /usr/bin
Installing easy_install-3.8 script to /usr/bin

Using /usr/local/lib/python2.7/dist-packages
Finished processing dependencies for airtime-analyzer==0.1
Removed /etc/systemd/system/multi-user.target.wants/airtime_analyzer.service.
Created symlink /etc/systemd/system/multi-user.target.wants/airtime_analyzer.service → /etc/systemd/system/airtime_analyzer.service.
  File "/home/ubuntu/libretime/python_apps/libretime_import/setup.py", line 1, in <module>
    from setuptools import setup
ModuleNotFoundError: No module named 'setuptools'
```

I thought I just needed to do something like `pip install setuptools` but that doesn't work. I suppose this is using some other python somewhere? Any help to fix this would be appreciated.

Mahalo.

paddatrapper:
Python 3 support for the rest of the apps is sitting in #951 and hasn't been merged yet, so the python 3 dependencies for this PR are probably not installed correctly

nathanfriend:
Is this feature useable?  I'm setting up a new LibreTime instance, have around 8000 media items to import from an AirTime server.  Don't fancy trying to do that through the web interface.

kmahelona:
Hi @nathanfriend i think you might have more luck with watch folders https://github.com/LibreTime/libretime/pull/908. It’s not quite the same as importing but I can sync thousands of files. It’s also a WIP so may require a few manual commands to get working.

nathanfriend:
Thanks @kmahelona I'll take a look.

zklosko:
Has anyone considered Linux's `inotifywait` binary? It keeps an eye on a directory and can be used to pipe file paths into a shell command. https://unix.stackexchange.com/questions/323901/how-to-use-inotifywait-to-watch-a-directory-for-creation-of-files-of-a-specific

Issue #545, opened 2018-10-12
Ubuntu Trusty 14.04.5 LTS EOL in April 2019


hairmare:
According to https://wiki.ubuntu.com/Releases Ubuntu Trusty 14.04.5 will be end of life by April 2019. This means that Trusty will stop receiving security patches in less than half a year. In effect we need to address this as soon as possible, 5 months is not a long time to switch to a new major OS release.

I'm proposing we do this as follows:
* [x] Add a warning about this deprecation to the 3.0.0-alpha.5 release notes
* [x] Mark Trusty as deprecated in docs
* [x] emit a warning in the installer #557
* [x] emit a warning during the `vagrant up ubuntu-trusty` command
* [x] Ensure that we support Ubuntu 18.04 Bionic Beaver enabling existing Trusty users to upgrade to a distro that results in another five years of support (until April 2023) rather than going for Xenial (which is EOL in April 2021)
* [x] Remove any trusty specific code after the deprecation notice has been published for a while
* [x] Replace easymp3gain-gtk with something still supported in Bionic (mp3gain was already replaced with python-rgain in legacy upstream)

I think I know that there are still a lot installations out in the wild based on Trusty. It needs to be abundantly clear that we cannot continue supporting those once Canonical stops releasing security patches for it. I haven't done much testing in Ubuntu, given that Debian 9 Stretch works fine I'd assume that a new version of Ubuntu shouldn't have much issues.

For the next Release there won't be too much we can do. For now there will be a warning in the release notes so people can get prepared. Starting with 3.0.0-alpha.5 there should also be (experimental) support for Bionic in the Vagrant set up and the installer. I'm assuming that most changes related to this will be in 3.0.0-alpha.6 to give users ample warning and some time. I'm not going to take the lead on anything related to Ubuntu support since I'm not a user of that case (I use Debian for development and CentOS on production).

frecuencialibre:
missing from the docs PRs i just made are updates regarding vagrant. assuming the default vagrant build will change as well. thanks Lucas!

hairmare:
The vagrant docs changes are in #546

PR #546, opened 2018-10-12
Add Ubuntu Bionic Beaver Vagrant box

hairmare:
This contains my preliminary work on Ubuntu Bionic Beaver support as per #545 and it creates the development environment needed to hack on Bionic support. It also marks Trusty as deprecated in Vagrant and the vagrant docs without (yet) touching on Trusty deprecation from the installers point of view.

It's marked as experimental in the docs since I didn't do enough testing to be confident that everything still works (ie. there are most likely some php 7.2 compat issues that are surfacing with this bump).

This also gets rid of the mp3gain or easymp3gain package since it was replaced with python-rgain during the switch to analyzer before our fork.


Robbt:
Cool. I was thinking this would be a good idea since 18.04 is a LTS and it would make sense for it to be a tested option for new installs.

PR #548, opened 2018-10-12
(docs) deprecate ubuntu 14

frecuencialibre:


docs/install.md
```diff
@@ -1,11 +1,16 @@
...
+Once you have downloaded and extracted the LibreTime repository, run the instalation script by navigating into the folder containing the LibreTime codebase, and run it's install script from the command line:
 
     :::bash
     sudo ./install
 
-Instalation in Debian 9 and other Linux distributions is possible, but multiple outstanding issues have yet to be resolved.
+By default, the installer will install and configure all dependencies.
+
+## Alternative OS installations
+Instalation in Debian 9 and other Linux distributions is possible, but multiple outstanding issues have yet to be resolved. Instalation on Ubuntu 14.04.5 LTS (Trusty Tahr) is also working, but deprecated due to the fact that this version will reach its official end of life in April 2019.
```
hairmare: Maybe we should consider removing the bits about Debian 9 at some point. I'm not aware of any issues that block Debian 9 installs, it's what I've been using for dev work since quite a while.

hairmare:
Thanks!

docs/install.md
frecuencialibre: i think there have been issues with importing tracks in debian, probably do to licensing wierdness. 

although i do work as a javascript dev and use debian every day, i'm trying to focus my libretime participation on getting this working easily for community radios. if focusing on a single distro (ubuntu) means that we can get it working bug free faster, then i'm not going to be a debian purist.

hairmare: Makes sense. I believe the issues with regards to importing are/where most likely centered around using a proper silan package.

One point that speaks for Debian is that silan and liquidsoap will finally be maintained again once Debian 10 (Buster) is released. Given that Ubuntu is downstream from Debian this is quite important. The liquidsoap bump in Debian will finally enable us to merge #192 and use a liquidsoap version that is actually maintained and supported by the savonet project. At the moment I'm ignoring the fact that #192 will be a backwards compatibility breaking change for almost all distros we currently support.

frecuencialibre: decoupled libretime 4.0.0 docker install on debian 10!!! lol

frecuencialibre: but seriously, very good points. and debian 10 freeze is in like january i think so it's not too far out

hairmare: Everything I know about already seems to have made it into testing, so Debian 10 will have what we need out of the box. I started documenting what we need on the wiki a while ago: https://github.com/LibreTime/libretime/wiki/Distro-Packages but it needs updating as it's most likely outdated by now.

hairmare: I think we can consider Ubuntu 18 and Debian 10 for beta releases, not sure about the final 3.0.0 though as it there is still quite some work needed for reaching the kind of stability I would expect from that.

Issue #560, opened 2018-10-14
Problem: How Automatic Playlists work is not intuitive with the current UI


Robbt:
The Automatic Playlist as currently set-up in the show configuration isn't easily understood by people just starting with LibreTime.

The problem is that one if a automatic playlist is added to a currently airing show (ie when someone is testing a new LT install) it won't do anything.

Additionally it isn't clear to users that an automatic playlist is scheduled an hour before the show airs or what a good example configuration for a "show based" automatic playlist is.

This was brought up in #544 

There are a few possible solutions, one simple one would be adding a tool tip that can further explain automatic playlists or modifying the code to have them schedule for new shows that are just created. In addition some of this could be handled in the documentation by providing more examples.

Another option is to add parameters to automatic playlists to allow the user to set at what time "before the show airs" they run and the possibilty to have multiple automatic playlists such as those that pre-pend the content rather than always appending the scheduled content if any. But it isn't clear whether complicating the setup would make it more usable or just confusing to end users.

JohnnyC1951:
Yet another option is simply add a paragraph in the Automatic Playlist section of the LT manual.

frecuencialibre:
Thanks @Robbt for sprinting yesterday. got a lot of balls rolling!!! 

Seems like we need a solution that is both feasible to code within our (as yet undefined) beta release timeframe, and also packs a decent user experience. 

I followed the steps in [the wiki](https://github.com/LibreTime/libretime/wiki/Automatic-Playlists,-Smartblocks-and-Podcasts), nice work! very cool. 

from my total ignorance, i'm guessing that background task scheduling is necessary for podcast downloading rather than anything specific to auto-playlists, correct? if so, would it be possible for automatic playlists and podcast downloading to be more independent from each other? Could you point me to code, or help me understand this point?

again, ignorant questions since i haven't looked at the code, but from just looking at the UI, i'd guess that i could add a podcast smartblocks to a show, and then click the "link" checkbox upon creating with all repeating instances. also ideally auto-playlists could also survive despite failure to download podcasts. just further reasons why IMO it'd be great for auto-playlist and podcast functionalities to be well separated.

hopefully this is at all helpful. 

also a sidenote: 
>  If this didn't happen go to Settings -> General table and ensure these options are both checked Podcast Album Override and Podcast Automatic Smartblock and Playlist.
i'd say these should just be enabled by default.

Robbt:
They are enabled by default I believe.
[...]

PR #591, opened 2018-11-08
Autoplaylist calendar fix

xabispacebiker:
This small fix displays a new icon in the autoplaylist schedules, also removes the "show empty" red icon to avoid confusion:

autoplaylist_calendar_fix


Robbt:
So this replaces the green A with the 3 horizontal lines and removes the red exclamation mark from shows that have an auto playlist ?

This makes sense to me.

frecuencialibre:
i got [up and running](http://github.com/frecuencialibre/libretime-docker-dev), tested this and can confirm it works as described. 

I agree that we should merge this now, but also see a few ways this PR could be improved:
- I associate the "hamburger icon" with a menu. The A stood for automatic, but that isn't translatable, and is [in flux](https://github.com/LibreTime/libretime/issues/560#issuecomment-438293763). I don't have any brilliant ideas, but perhaps a clock?
- The exclamation point is removed if a playlist is assigned. It has no relationship to whether the show is actually full. Ideally we would hook into the same system that is used for conventionally scheduled content, but I'd guess we have other pressing issues to fix first. 

Awesome! thanks @xabispacebiker !

Robbt:
So I agree that a clock would probably be better I found the [eva collection](https://akveo.github.io/eva-icons/#/) of open-source icons has this clock [icon](https://akveo.github.io/eva-icons/outline/png/128/clock-outline.png) which looks like it would work. Perhaps we could use that instead of the hamburger menu.

Also I think that if we are going to remove the RED ! then we should replace it with some kind of indicator so that people can see if a show was scheduled or not. It is also useful to have the orange partial scheduled icon visible, especially to easily gauge whether a autoloading playlist failed to fill the show for some reason. Perhaps we could change the icon though or even color-code the clock icon. Just thinking about it. I don't want to remove it all together but it could definitely make sense to change it.

frecuencialibre:
i like the clock you found. i think it's definitely an improvement over the menu icon. the red exclamation point never disappearing despite having scheduled an autoloading playlist is confusing to people at my station. 

color coding the icon is an interesting thought, and easily done with css classes if we add it as an inline svg, but i'm not sure people would understand it. the orange and red exclamation point icons are already present in the UI, and already have established meanings.

how complicated would it be to either 1. check whether or not the show is actually full using the system that is used for conventionally scheduled content, or 2. show a css class conditionally depending on whether or not the repeat checkbox is checked?

my vote: 
1. to replace the menu icon with the clock, keeping current conditional display logic. 
2. if either of the two options i ask about above are feasible, then use one to conditionally add the yellow exclamation point. 

@xabispacebiker thoughts?

frecuencialibre:
after concluding with @Robbt that there is no good way to gauge whether an autoloading playlist will fill a show, i went ahead and did the icon change without adding any extra logic. it's this same PR with a few very small fixes to @xabispacebiker 's work: https://github.com/LibreTime/libretime/pull/615

Issue #487, opened 2018-05-13
Blank page on login


vanzeelandj:
Ubuntu 16.4.04 on vmware dedicated host.
I have two near identical hosts. One - after apt-get upgrade (I think) - developed the following login problem logged below. Result is a blank page after logon instead of GUI. Havent updated the other it seems fine. I figure it's probably something fairly straightforward but I haven't worked it out yet.

Wondering if related to #410 
I'm imagining some code needs revision to suit a new version or I roll something back...

Initial install was from vanzeelandj/libretime which is same as current tree. I tried re-running install etc 

Tried re-install using 3.0.0-alpha.3 sources having read re: similar issue, but fault persists - event log result same with different version string i.e. 0c34d*** rather than 3.0.0.-alpha.3

2018-05-13T11:20:50+10:00 ERR (3): air01 [Logging.php:166 - loggingShutdownCallback()] - Uncaught TypeError: Argument 1 passed to Composer\Semver\Semver::satisfiedBy() must be of the type array, string given, called in /usr/share/airtime/php/airtime_mvc/application/views/helpers/VersionNotify.php on line 37 and defined in /usr/share/airtime/php/vendor/composer/semver/src/Semver.php:53
Stack trace:
#0 /usr/share/airtime/php/airtime_mvc/application/views/helpers/VersionNotify.php(37): Composer\Semver\Semver::satisfiedBy('3.0.0-alpha.3', '>=3-stable')
#1 /usr/share/airtime/php/vendor/zendframework/zendframework1/library/Zend/View/Abstract.php(349): Airtime_View_Helper_VersionNotify->versionNotify()
#2 /usr/share/airtime/php/airtime_mvc/application/layouts/scripts/layout.phtml(32): Zend_View_Abstract->__call('versionNotify', Array)
#3 /usr/share/airtime/php/vendor/zendframework/zendframework1/library/Zend/View.php(157): include('/usr/share/airt...')
#4 /usr/share/airtime/php/vendor/zendframework/zendframework1/library/Zend/View/Abstract.php(888): Zend_View->_run('/usr/share/airt...')
#5 /usr/share/ai at /usr/share/airtime/php/vendor/composer/semver/src/Semver.php:53
array (
  'type' => 1,
[...]

Issue #580, opened 2018-10-30
Problem: ZendFramework 1 is incompatible with Php 7.2 and thus Ubuntu 18.04


Robbt:
So, Php 7.2 introduced some inheritance tightening that causes ZF1 to break. See this [post](https://stackoverflow.com/questions/49072332/is-zend-framework-1-compatible-with-php-7-2). 

Ubuntu 18.04 uses Php 7.2 by default and so LT is no longer going to work with this without some workarounds.

One option would be to with the installer force Ubuntu 18.04 to use an older version of PHP, I think Php 7.1 should still work.

Another option would be to switch to a community fork of Zend 1 Framework that includes Php 7.2 fixes. But I don't think there really is one here is a [discussion](https://discourse.zendframework.com/t/zf1-community-maintained-official-fork/419/5) that doesn't end in an actual project and includes some reasons why this isn't a good idea. Someone did make individual [composer packages](https://github.com/zf1s) but it doesn't appear to be very well maintained and there doesn't seem to be a fix in zend_rest for the issue we are running into.

It should work on Php 7.0 which is supported until 3 Dec 2018. and PHP 7.1 is supported until 1 Dec 2019. The breakages all happen with Php 7.2

hairmare:
*sigh* 7.2 is going to be relevant really soon. We might be able to get away by switching to the https://github.com/magento/zf1 fork that seems to support 7.2 as far as I can tell. I haven't tried yet but i did some digging through available zf1 forks.

Robbt:
Yeah none of the forks looked promising from the bit of digging I did. They all rely upon an unsupported version of PHPUnit and have no tests for anything beyond Php 7.0. Also none of them appeared to modify the Rest server php file that was erroring when I tried the bionic vagrant box. There doesn't appear to be a real supported fork, so we might have to start work on migrating sooner than we like and just set 3.0 to work on older PHP versions and distros that support it. The discussion I linked to above was pretty relevant because it featured the people who had been maintaining Zend 1 before it was EOL and their perspective on the work required to keep it running with new versions of PHP.

hairmare:
I haven't really done a deep dive into this (yet) but I might have some reasons to do so for non-LibreTime related projects... For now it looks like we won't be supporting 7.2 in the foreseeable short term future.

Robbt:
[...]

PR #594, opened 2018-11-16
Fixing failing version test

jelko:
Wrapping version in Array() to comply with SemVer new method signature. This should fix issue #487. Tested on Ubuntu 18.04. with PHP 5.6 with 3.0.0-alpha5 (7.2 or 7.0 is still incompatible, see issue #580)


hairmare:
The moment when you realize that `composer/semver` either doesn't follow semver or doesn't consider itself an api 😕([this](https://github.com/composer/semver/commit/94b96b7b710a951b1d64e76a450928247c4c4ac8) should have been a major bump!).

This looks like it certainly will fix #487, I think there are more bits than just this needed for #580.

frecuencialibre:
@hairmare a few of us are seeing errors which i think have been introduced by this change https://github.com/LibreTime/libretime/issues/487#issuecomment-439767173 thanks all!

PR #595, opened 2018-11-19
UX: collapse advanced configuration sections in Schedule Dialog

frecuencialibre:
this commit collapses by default several sections of the add/edit show dialog:

screenshot from 2018-11-19 11-42-13

why?
- simplify interface. less distraction for users and less need to scroll way down to the "update show" submit button
- prerequisite to proposal i'll work on now to improve the autoplaylist UI

problems? 
- only that i would have like to have those little orange triangles rotate indicating open/closed state, but that'd require too much (crappy) code and time.


hairmare:
This makes it much easier for new users with a normal sized screen to navigate creating a show 👍

Issue #596, opened 2018-11-20
 Automated installation


Dr-Satan:
Hi, excuse me but the automated installation page is missing 
https://libretime.org/manual/automated-installation/

How to install Libretime easily on a ubuntu server ?

Robbt:
It is installable via the install script. It's best installed on a dedicated server and Ubuntu 16.04 is the best right now.

frecuencialibre:
instructions here: http://libretime.org/install/

the automated installation page was intentionally deleted because it contained out of date info.

todo: remove references to deleted pages in documentation

Dr-Satan:
Thank for your answer but how to download and extract the LibreTime repository ? From which source and in which server directory?

Robbt:
I just added the instructions to install.md here they are - 

You can download LibreTime from our [github repo](https://github.com/LibreTime/libretime) In order to run the latest codebase you can click the green Clone or download button. It is recommended to type ``` git clone https://github.com/LibreTime/libretime.git ``` (if you Debian based system lacks git then you can type ``` sudo apt-get install git ```) and this will download the source code to your repository. You can also click Releases and download a tarball of our latest releases. You will then need to extract this archive into a directory on your server. It is not recommended that you install LibreTime on the same computer you are using as a desktop. So you will probably need to use wget for instance ```wget https://github.com/LibreTime/libretime/releases/download/3.0.0-alpha.6/libretime-3.0.0-alpha.6.tar.gz``` and then extract it by typing ```tar -xvzf libretime-3.0.0-alpha.6.tar.gz``` in the directory you downloaded it into. 

Let me know if this makes sense, if not I can clarify and also update our docs to make it easy to understand.

Dr-Satan:
Thank you very much for your explanation, I will opt for a git clone. I want to install it on a remote server ubuntu 16.04. In which directory would you advise me to clone Libretime ?  (home ? root ? etc ? bin ?)

Robbt:
I'd put in a /home/user directory of whatever login you use to ssh into it but it doesn't really matter. Although you do need to run the install with sudo but I think that is in the docs.

frecuencialibre:
i'm closing this since it is more of a support request that appears to have been answered. @Dr-Satan note that for support requests you can also post in the [forum](https://discourse.libretime.org/). thanks all!

hairmare:
There are still some links that now point to 404 since our recent cleanup

[...]

PR #597, opened 2018-11-20
Update install.md to add instructions for how to download codebase

Robbt:
I'm adding directions for how to download LibreTime as per a request from #596


hairmare:
I personally wouldn't recommend installing this way on production for someone new to Linux... Does it need a warning or some kind of disclaimer?

Issue #78, opened 2017-03-13
Setup Weblate translations integration


hairmare:
I'm still torn about deciding to integrate https://github.com/LibreTime/libretime/commit/2e9583c687b6c4922d3160bc32be1642b92b1be1 from 2.5.x. There absolutely needs to be some automation to keep translation working and up and running, I'm pretty sure we have some russian customers that will greatly appreciate us not breaking this.

While transifex is free for opensource, it looks like http://zanata.org/ from redhat is actually the more libre way to go. I think our setup is pretty close to a podir config as per [the zanata docs](http://docs.zanata.org/en/release/user-guide/projects/podir-example/).

If no one has a better proposal (I would actually prefer a solution that allows github logins) I'll do some testing on my branch with real zanata and then open a pr when it looks ready.

* [x] configure zanata and push initial strings
* [ ] push new strings to zanata using travis
* [ ] pull translated strings into github repo and/or release tarballs

Fixes #107

hairmare:
I set up a zanata for LibreTime and tested it by round-tripping some de-DE strings. It may be found at https://translate.zanata.org/iteration/view/libretime/master.

I have the maintainer tooling somewhat prepared on a branch and am looking into how we can automate this through travis. I'll also have to test the install a couple of times so I don't break anything.

For now the zanata is useable, but you need to ping me if you want any changes to be pulled back into the LibreTime GitHub.

hairmare:
This is raising some questions over here 😕.

1. When should we push strings to zanata? After each merge to master or on each tag.
2. When do we want to pull strings from zanta? Should they get auto committed?

I think legacy upstream had a cronjob or similar for pulling strings. Zanata has [webhooks](http://docs.zanata.org/en/release/user-guide/projects/project-settings/#webhooks) but I only see one for 100% translated events.

I'm not sure I'm a fan of auto committing, IMO preferably zanta would create pull requests so we can hold off if we are sure some bad strings slipped in. It might also make sense to just bake a zanata pull into the release tarball and to update strings in the repo manually right before regenerating and pushing new ones.

Also, should regenerating strings be part of a PR that touches such strings or do we bundle multiple string changes into one run of the gettext tools?

jooola:
Zanata seems not maintained anymore, I propose to use Weblate https://weblate.org/en/.

jooola:
I created the project here https://hosted.weblate.org/projects/libretime/

The weblate admin will only have to approve our subscriptions as an open source project within a week.

[...]

PR #599, opened 2018-11-20
Auto playlist ui

frecuencialibre:
as of https://github.com/LibreTime/libretime/commit/b60d4581ee15da6f13a01ce981055f28cd4881e5 this is a bare-minimum solution to the problem described in #560. It does this:

screenshot from 2018-11-20 16-19-08

IMO this could be merged as-is, but would be much better if we removed the checkbox. Selecting "None" with the select field would be equivalent to an unchecked checkbox. I'd love to achieve this final result:

screenshot from 2018-11-20 16-25-37
 
I started attempting removing the checkbox, but ran up against an issue: selecting "None" seems to be "forgotten" by the UI. Said another way, if you 1. select a playlist, 2. save the show, 3. edit the show again and select "None", 4. save the show, 5. edit the show _again_, you'll see the show selected in step 1 still active. This behavior is present in the master branch prior to any changes in this PR. I stopped because I ran out of time and started touching so many files that I decided not to risk causing unintended side effects. All help or feedback appreciated!

### DIFF
--- a/airtime_mvc/application/forms/AddShowAutoPlaylist.php
+++ b/airtime_mvc/application/forms/AddShowAutoPlaylist.php
@@ -15,7 +15,7 @@ class Application_Form_AddShowAutoPlaylist extends Zend_Form_SubForm
 
         // Add autoplaylist checkbox element
         $this->addElement('checkbox', 'add_show_has_autoplaylist', array(
-            'label'      => _('Auto Schedule Playlist ?'),
+            'label'      => _('Add Autoloading Playlist ?'),
             'required'   => false,
             'class'      => 'input_text',
             'decorators'  => array('ViewHelper')
@@ -29,7 +29,7 @@ class Application_Form_AddShowAutoPlaylist extends Zend_Form_SubForm
         $this->addElement($autoPlaylistSelect);
         // Add autoplaylist checkbox element
         $this->addElement('checkbox', 'add_show_autoplaylist_repeat', array(
-            'label'      => _('Repeat AutoPlaylist Until Show is Full ?'),
+            'label'      => _('Repeat Playlist Until Show is Full ?'),
             'required'   => false,
             'class'      => 'input_text',
             'decorators'  => array('ViewHelper')
--- a/airtime_mvc/application/views/scripts/form/add-show-autoplaylist.phtml
+++ b/airtime_mvc/application/views/scripts/form/add-show-autoplaylist.phtml
@@ -7,6 +7,7 @@
         </dt>
         <dd>
             <?php echo $this->element->getElement('add_show_has_autoplaylist') ?>
+            <span class="show_autoplaylist_help_icon" aria-describedby="ui-tooltip-2"></span>
         </dd>
 
 	<div id="add_show_playlist_dropdown">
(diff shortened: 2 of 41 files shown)


Robbt:
I tried to test this and it ran into conflicts with master (I suspect this is because of where we merged in your previous change to make everything collapsed and this was based upon master before that was merged). I'm not sure but I'll let you take a first stab on resolving it but I can help if you are equally uncertain as to what happened.

As far as extending Zend_Form_Element_Select this seems like it could require a bit more digging into Zend1 that I think might be warranted to fix this I found this [discussion](https://stackoverflow.com/questions/45332405/extend-functionality-of-zend-form-element) on stackoverflow about extending it. 

I guess its just frustrating to still be stuck on Zend 1 and still have no clear path forward at this point aside from rewriting everything. I must say we are tremendously lucky that you @hairmare had this job experience working in Zend 1, it has been invaluable with just figuring out how to get LibreTime operational.

frecuencialibre:
thanks all for taking time to look into Zend etc. I think the merge conflict is now fixed.

Based on your feedback I doubt I'll have the bandwidth to dig into Zend, so if that's going to happen it would need to be a contribution from someone else. However, one quick change i could make would be to swap out the tooltip you can see in the first screenshot for the text you can see in the second. any preference either way?

> clear path forward

at this point, and especially after this first adventure into Zend 1 land, i'm thinking that my long-term (ie. post LT 3.0.0) plan to help design and create a mobile-first radio admin tool with great UX will probably be best built with [azuracast](https://github.com/AzuraCast/AzuraCast/). obviously there'd be lots of work to do there too. many features present in libretime are missing in azuracast (metadata management, smartblocks within playlists, graphic creation of playlists, user management integrated with the schedule, etc. etc.), but the project has much less technical debt, appears to be more stable, has mature docker support and upgrade paths, and even work started on an API to enable decoupling. who knows, i mention it here though in the spirit of joining forces and thinking flexibly about all options.

Robbt:
Well I was curious about this because I'm a sucker for pain and I did find people discussing [this exact problem](https://github.com/zendframework/zendframework/issues/2761) but upon further analysis it appears that they are talking about Zend 2.

My suggestion would be to combine the two, keep the hover smart tip but then also add the text underneath the checkmark and above the select option. That way it becomes very clear for them.
Also I noticed one small typo in the word information in the tool tip.

Also while it would probably make sense in an ideal UI world to have the autoplaylist turn on or off based upon whether they selected a playlist vs. requiring an additional checkbox, the additional checkbox ties directly into the database and the extra work to make the database flag set to 0 when none is selected and then 1 when a playlist is selected is probably not worth the effort. I have found when working with Zend to look for the easiest way of accomplishing the goal.

With regards to Azuracast, that might be a good path forward. My vision for LibreTime is to start building a well documented API and decoupling the Zend components and replacing the UI with a modern javascript framework. This is all easier said than done but until we get off of Zend we will continue to run into issues like this. But yeah many hands make light work and this project for better or worse has been a real community effort all along. And if it is going to work we need to make it easier for people to contribute and have it depend upon software that isn't unsupported.

frecuencialibre:
hopefully this is now ready for merge. and should close #560 

screenshot from 2018-11-25 14-46-15
screenshot from 2018-11-25 14-46-20

hairmare:
For some reason the diff looks like this will revert some recent changes.

hairmare:
To make stuff translatable it's usually enough to wrap the strings in `_()`. Rebuilding the translation files doesn't need to be in the same PR as any missing strings will default to what's passed to `_()`. I've been meaning to finally get back onto #78 to clean up things related to i18n.

frecuencialibre:
please disregard this PR. replaced by https://github.com/LibreTime/libretime/pull/611

PR #601, opened 2018-11-21
:heavy_minus_sign: drop wheezy support

hairmare:
Removes the Vagrant set up for Debian wheezy and makes the installer abort and complain if started
on wheezy.


Robbt:
Looks good I noticed you said 2.0.0.6 alpha vs 3.0.0.6.alpha - but that's just a slight typo, also it looked like some docs were changed that still said Airtime (but they didn't look actually modified in any way I could tell). Anyways I can merge this as is or we could fix that typo and merge it.

hairmare:
I fixed the version issue in my last force push... The change to the docs that still say "Airtime" are only there due to issues with my editor changing the trailing EOL char in the docs ad there are no real changes to the actual copy.

PR #603, opened 2018-11-21
docs: update upgrade instructions to remove references to deleted obselete instalation methods

frecuencialibre:
this one is in response to a task that surfaced in https://github.com/LibreTime/libretime/issues/596

i honestly have no idea what upgrade best practices are, so this just removes clearly erroneous text. a quick review would be great.


Robbt:
Looks good. I hope at some point we create packages and can have an automated installation that is more hands-off but this is definitely an improvement. Thanks for the work on improving the docs.

Issue #357, opened 2017-11-15
Smart Blocks overfill hour


jerry924:
I have a problem when using smart blocks in that they are not smart enough to find songs that fill hour exactly or close as possible.

The problem is when the next hour starts, whatever is oversubscribed just gets chopped and it is very annoying to listeners.

Here is how I manually work around this and I think we can improve smart blocks to do this automatically.

1) I have several tracks that are in an "Ambient Sound" genre. These are tracks that are just some ambient sounds that can easily be chopped without affecting the listeners.  Imagine noise of waves crashing on a beach, or city street noise, or even very light elevator-like/simple music which doesn't sound bad to get cut off randomly.

2) We have enough music that there is a variety in track lengths (over 1000 tracks).

3) If I create a smart block for 60 minutes, sometimes it is 56 minutes, sometimes it is 63 minutes, etc. The problem is the quiet time for 4 minutes or the chopped off song that had 3 minutes left to go in it.

4) The steps I follow manually to solve this:
      A) fill 60 minutes with a smart block
      B) if time remaining < 30 seconds, add ambient sound track - done
      C) if time remaining > 2:30, find a music track that is between (remainingtime-30s) and (remainingtime), but closest to (remainingtime) without going over and add it there and add ambient sound track to the end of that - done.
      D) if time remaining is > 30 seconds and < 2:30 then I delete last TWO tracks and repeat (B),(C)
      E) if time is oversubscribed, delete last track and check (B),(C),(D) above.

at step (C), 80% of the time I have a track that perfectly fits so all of my hours are +0.  The problem is when I use a smart block on a linked show and each link is derived with different content, it is very hard to go and manually fix them all. 

Given that I have a formula/process for this, seems like it could be coded into the smart block creation code just like this.

Smart block would need a new input which is a track or other smart block that spits out one ambient sound/choppable track. This setting would need to be stored in the smart block definition and used to fill in any < 30 second time at the end of a smart block.

hairmare:
I like the idea of choppable tracks. I can see an additional use case for them in combination with the podcast download thing. The thirty second window would need to configurable for that case and step A might produce an empty playlist.

If I recall correctly the actual scheduling code gets triggered some time before the smart block needs to be evaluated by a request to the mvc which is usually done by the playout daemon on a regular base. This makes hacking on the scheduling code a bit inconvenient. I usually shunt some code in my working copy to make hacking easier.

jerry924:
[...]

PR #605, opened 2018-11-24
Fill Remaining Time Smartblock Limit

Robbt:
So this is my work on building a smartblock that is aware of how much time is remaining in a show instance it is scheduled into and will grow or shrink based upon what is already scheduled. You can see #357 for more discussion on the rationale for the changes and the thought process behind the changes.

TODO still
- [x] -  figure out why it is overflowing the time w/ repeating tracks enabled - see #606 
- [x] - get rid of the sp_limit_value textbox via jquery when "fill remaining time" option is selected

### DIFF
--- a/airtime_mvc/application/forms/SmartBlockCriteria.php
+++ b/airtime_mvc/application/forms/SmartBlockCriteria.php
@@ -159,7 +159,8 @@ class Application_Form_SmartBlockCriteria extends Zend_Form_SubForm
             $this->limitOptions = array(
                 "hours"   => _("hours"),
                 "minutes" => _("minutes"),
-                "items"   => _("items")
+                "items"   => _("items"),
+                "remaining" => _("remaining time in show")
             );
         }
         return $this->limitOptions;
--- a/airtime_mvc/application/forms/SmartBlockCriteria.php
+++ b/airtime_mvc/application/forms/SmartBlockCriteria.php
@@ -160,7 +160,7 @@ class Application_Form_SmartBlockCriteria extends Zend_Form_SubForm
                 "hours"   => _("hours"),
                 "minutes" => _("minutes"),
                 "items"   => _("items"),
-                "remaining" => _("remaining time in show")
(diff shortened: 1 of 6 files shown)


Robbt:
Ok, so I'm thinking this is ready for test by other people and merging. It is working in all of my tests.

I didn't disable the "time remaining" for static smartblock because to go through and figure out how to make the jquery code work to remove a individual value from the drop down and then add it back with translated text when dynamic smartblock is chosen seems like too much work for little return.

Basically if you generate or preview a time remaining smartblock it will create a 24 hour playlist. This is because this is the maximum time that a show can ever be. If someone selects this for a static smartblock and generates they will just end up with a 24 hour long smartblock. This doesn't seem sufficiently broken to warrant the pain of dealing with jquery. I'm open to feedback from other people if they decide that this isn't the case and we should fix it before merging.

But otherwise I think this code will do a lot to make smartblocks and autoloading playlists more useful for stations wanting to avoid dead air.

JohnnyC1951:
This seems very good indeed. Is it dynamic? IE if an autoloading show arrives via Podcast import, will this will refactor the smartblock length?

Robbt:
Yes, the intention for this is to be used in conjunction with content where you don't necessarily know the length ahead of time. Since podcasts can vary in length if you have a fill remaining time smartblock after say a 56 minute length track in a 60 minute show it will treat it as a 4 minute long smartblock and schedule matches up until that limit. Combined with the fix to repeating tracks and adding an option for tracks you want to allow to overflow with #607 (such as ambient tracks) you can solve the issue that plagues syndicated content where either you have a bunch of content cut off halfway through the track or you have dead air because your filler wasn't long enough.

JohnnyC1951:
At the moment, I have a section of smartblock fillers of like 30 mins, 60 mins, 120 mins and I simply put one of show length after the show podcast smartblock. Nothing bad ever seems to happen, but I guess there is a lot of extra processing going on at show-load time.

PR #607, opened 2018-11-25
fixed allow repeat tracks smartblock so it doesn't overflow time limit

### LINKED ISSUE #606: Problem: Smart Block behavior is inconsistent in terms of limit overflow between blocks with repeating tracks and those that don't repeat tracks
In writing the code for #605 I determined that Smart Blocks that allow repeating tracks will keep scheduling until they overflow the time limit, whereas Smart Blocks that do not do this will never overflow the time limit.

Overflowing in this case means that it will schedule matching tracks until the length of the scheduled tracks is greater than the limit. This means that in almost all cases a track will be cut off in mid-play. This was the default behaviour before we launched LibreTime when we modified the tracks to schedule up until the limit but never over. See [this commit](https://github.com/LibreTime/libretime/commit/ba9f7879a47444751ca73edb4f259b9affbe1866)

In some special cases such as ambient filler it may be useful to have a smartblock that will overflow the time see #357 so it may make sense to add this as an additional option to the smartblock window.

In addition allowing tracks to repeat but not overflow could result in the same short track being selected multiple times in a row because it is the only one that is under a limit. For instance say there is 2 minutes remaining in a time limit but all songs are 3 minutes or longer and only a station ID is 10 seconds. The way this would work is that the station ID would be scheduled 12 times in a row at the end of the smartblock because it can be.

So it could also be the case that we want to keep this idiosyncrasy and document it as such to help people avoid nightmare scenarios with badly repeating tracks.

When investigating this it could be related to or contain the solution to #224.

Robbt:
This fixes #606 by employing the same test for each track to determine whether it will overflow the time limit of the smartblock before it schedules it.

### DIFF
--- a/airtime_mvc/application/models/Block.php
+++ b/airtime_mvc/application/models/Block.php
@@ -1355,14 +1355,22 @@ SQL;
         $sizeOfInsert = count($insertList);
         
         // if block is not full and repeat_track is check, fill up more
+        // additionally still don't overflow the limit
         while (!$isBlockFull && $repeat == 1 && $sizeOfInsert > 0) {
         	Logging::debug("adding repeated tracks.");
         	Logging::debug("total time = " . $totalTime);
         	
             $randomEleKey = array_rand(array_slice($insertList, 0, $sizeOfInsert));
-            $insertList[] = $insertList[$randomEleKey];
-            $totalTime += $insertList[$randomEleKey]['length'];
-            $totalItems++;
+
+            $projectedTime = $totalTime + $insertList[$randomEleKey]['length'];
+            if ($projectedTime > $limit['time']) {
+                $totalItems++;
+            }
+            else {
+                $insertList[] = $insertList[$randomEleKey];
+                $totalTime += $insertList[$randomEleKey]['length'];
+                $totalItems++;
+            }
             
             if ((!is_null($limit['items']) && $limit['items'] == count($insertList)) || $totalItems > 500 || $totalTime > $limit['time']) {
                 break;


airtime_mvc/application/forms/SmartBlockCriteria.php
```diff
@@ -446,6 +446,14 @@ public function startForm($p_blockId, $p_isValid = false)
         }
         $this->addElement($repeatTracks);
 
+        $overflowTracks = new Zend_Form_Element_Checkbox('sp_overflow_tracks');
+        $overflowTracks->setDecorators(array('viewHelper'))
+            ->setLabel(_('Overflow Time Limit:'));
+        if (isset($storedCrit["overflow_tracks"])) {
+            $overflowTracks->setChecked($storedCrit["overflow_tracks"]["value"] == 1?true:false);
```
hairmare:
```suggestion
            $overflowTracks->setChecked($storedCrit["overflow_tracks"]["value"] === 1);
```

The `$storedCrit["overflow_tracks"]["value"] == 1` comparison is already a boolean so the `?true:false` isn't needed. It's also likely that the comparison should be `===` rather than `==`.

docs/manual/library/index.md
```diff
@@ -102,7 +102,11 @@ To add an AND criteria, such as **Creator** containing *jimi* AND BPM in the ran
 
 By default, a smart block will not contain repeated items, which will limit the duration of the block if you do not have sufficient items meeting the specified criteria in your **Library**. To override the default behaviour, check the **Allow Repeat Tracks** box. The **Sort tracks by** menu offers the options of **random**, **newest** or **oldest** items first.
 
-<img src="static/Screenshot582-Smart_block_repeat_tracks.png" alt="Smart block repeat tracks" width="595" height="255" />
+![](static/Screenshot582-Smart_block_repeat_tracks.png)
```
hairmare: You can but the alt text between the brackets:
```suggestion
![Smart block repeat tracks](static/Screenshot582-Smart_block_repeat_tracks.png)
```
I think that we used an `` here is an artefact of the conversion from booktype.

airtime_mvc/application/models/Block.php
```diff
@@ -1355,14 +1375,30 @@ public function getListOfFilesUnderLimit()
...
+                $totalTime += $insertList[$randomEleKey]['length'];
+                $totalItems++;
+            }
+            else {
+                $projectedTime = $totalTime + $insertList[$randomEleKey]['length'];
+                if ($projectedTime > $limit['time']) {
+                    $totalItems++;
+                }
+                else {
+                    $insertList[] = $insertList[$randomEleKey];
```
hairmare: We might want to refactor the `getListOfFilesUnderLimit()` method at some point since it's becoming unwieldy (ie. the contents of this else are an exact copy of lines 1378-1389 above. I'm fine with not refactoring it in the PR though.

airtime_mvc/application/forms/SmartBlockCriteria.php
Robbt: I tried it with == and it worked, with === it was not working, so I assume that the type gets translated at some point by Zend or the interface but it did work without the extra ?true:false so I removed that.

hairmare: `$storedCrit["overflow_tracks"]["value"]` is probably a string that get cast to a boolean, so the `==` is fine in this case (i didn't check the actual type while reviewing). While it can be confusing this kind of auto-type-casting is something I consider as a neat part of PHP :)

docs/manual/library/index.md
hairmare: Yeah, this should have been converted as well during the great docs refactor. Re-adding the alt attribute should/could help with screen readers and i11y but isn't all that important.

PR #609, opened 2018-11-25
delete editor temp files

nikmartin:
There are two temporary files ending in ~ that don't belong - this PR deletes these files


hairmare:
Thanks!

PR #611, opened 2018-11-26
"Autoloading" playlists UI changes. closes #560

### LINKED ISSUE #560: Problem: How Automatic Playlists work is not intuitive with the current UI

frecuencialibre:
in the end i had to re-do this one. should be clean this time


Robbt:
Awesome everything looks good on here as far as I can tell. Thanks for doing this.

PR #612, opened 2018-11-26
make "Repeat Playlist Until Show is Full ?" string translatable

frecuencialibre:


frecuencialibre:
went ahead and merged this myself since i considered it sufficiently straightforward. hopefully no probz!

Robbt:
yeah i think that's fine.

PR #613, opened 2018-11-26
spanish translation for Autoloading Playlist section

frecuencialibre:
@xabispacebiker @matias-tecnosoul @Robbt how does "Lista Auto-Cargada" sound?


Robbt:
It seems fine to me based upon looking it up in a Spanish to English dictionary since my Spanish vocabulary is limited. I'm going to merge this because I think its correct and ideally we will have a separate translation interface when #78 is done.

PR #615, opened 2018-11-27
add clock icon for autoloading playlists in calendar, remove empty show alert

frecuencialibre:
unfortunately this also includes a bunch of formatting cleanup done automatically by my text editor, and now i don't have time to fix that.

### DIFF
--- a/airtime_mvc/public/css/fullcalendar.css
+++ b/airtime_mvc/public/css/fullcalendar.css
@@ -314,7 +314,8 @@ a.fc-event {
 .fc-event-time,
 .fc-event-title {
 	padding: 0 1px;
-	}
+	height: 12px;
+}
 	
 .fc .ui-resizable-handle { /*** TODO: don't use ui-resizable anymore, change class ***/
 	display: block;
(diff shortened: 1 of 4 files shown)


frecuencialibre:
hold up. not ready for merge. this would undo recent changes to airtime_mvc/public/css/styles.css

Robbt:
The formatting stuff shouldn't be an issue but I noticed it removed  .show_autoplaylist_help_icon { from line 207 of style.css - that doesn't seem intentional but a artifact of this being done before we did the other PR. So it probably should be changed.

Robbt:
Cool, those icons are so small but that's not really the problem of this. It seems to work but I'm still seeing the red ! for shows with autoloading playlists.

frecuencialibre:
could your browser be caching the js file that adds those? i typically refresh with the "disable cache" checkbox checked in the Network tab of firefox or chrome devtools.

Robbt:
yeah it was a caching issue. I just did a shift refresh and it worked. Going to merge this now as I think it is good.

PR #625, opened 2018-11-30
make add show dialogue scrollable. fixes #338

### LINKED ISSUE #338: Problem: There is no scrollbar for the new show dialog when it overflows the browser window.
This has been driving one of our volunteers crazy because the functionality with Airtime.pro is different than LibreTime. 

Basically the New Show left dialog overflows without any obvious means of scrolling. (I've found if you scroll on the far right of the browser window it works after you scroll past the Calendar)

The Calendar has its own scroll bar as well which further complicates it.

Here are some screenshots displaying the problem.

noscrollbarinnewshowwindow

frecuencialibre:
although the discussion in #338 gets into js frameworks, this was a simple css fix. tested in firefox and chrome. should be fine in ie.

### DIFF
--- a/airtime_mvc/public/css/add-show.css
+++ b/airtime_mvc/public/css/add-show.css
@@ -7,6 +7,11 @@
     font-size: 12px;
     /*width: 25%;*/
     width: 310px;
+    max-height: calc(100vh - 144px);
+    overflow-y: scroll;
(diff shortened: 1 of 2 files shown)


Robbt:
Your a CSS rock star.

PR #630, opened 2018-12-02
removed php escape of characters that broke smartblocks

### LINKED ISSUE #328: Problem with HTML encoding in smart block builder
**Problem**
There an issue with SmartBlock builder saving some characters as the HTML encoded value when the are entered in the 'sp_criteria_value' field. Can lead to SmartBlocks note returning the desired tracks.

**Steps to reproduce**
Set the 'genre' of a track or two to 'D&B'.
Create a new SmartBlock
Set the criteria to be 'Genre' 'is' 'D&B'
Click generate - (No results are returned - Unexpected)
Save the Smart Block
Close The Smartblock
Reopen the SmartBlock
Observe the criteria value is now 'D&amp;B'

I did try to find where this might be happening, but struggled. Maybe 'airtime_mvc/application/forms/SmartBlockCriteria.php' ?

Robbt:
Basically there was what I think is an unnecessary escaping of the values with the php function htmlspecialchars
I don't think this is necessary as it isn't done for any of the other strings that take characters and there is escaping done later on when the query is built.
This fixes #328 and so you can find your drum and bass with D&B in both the Genre & Smartblock

### DIFF
--- a/airtime_mvc/application/models/Block.php
+++ b/airtime_mvc/application/models/Block.php
@@ -1439,7 +1439,7 @@ SQL;
         foreach ($out as $crit) {
             $criteria = $crit->getDbCriteria();
             $modifier = $crit->getDbModifier();
-            $value = htmlspecialchars($crit->getDbValue());
+            $value = $crit->getDbValue();
             $extra = $crit->getDbExtra();
 
             if ($criteria == "limit") {


hairmare:
I think this was put in for XSS protection reasons. I haven't had the time to think about the ramifications of just removing this (or an alternative solution that makes sense and keeps the XSS protection here).

frecuencialibre:
would anonymous users be able submit this text? if not then preventing xss
is less of an issue.

Robbt:
Only people capable of setting up smart blocks could input text here. So maybe it was any of the otheradded to prevent people from doing something on airtime.pro ?

There is no similar XSS protection on say the smartblock title fields so I wonder why this was chosen in particular to need it. I also think that special characters are escaped later on before they are used as comparisons in the database - see https://github.com/LibreTime/libretime/issues/328#issuecomment-443511638

My guess is that this is unnecessary, and we know it's causing issues, so we should remove it unless we can figure out how it could even be exploited.

hairmare:
A combination of CSRF and XSS might feasible... I'm guessing this was put in due to some vulnerability scanner flagging it as an issue, most likely an old version of metasploit or some other toolkit. I believe XSS is also relevant if it needs a local user to exploit, one an evildoer owns one of your DJs personal computer this *could* be used as a vector for spreading to other users of the same LibreTime instance.  That said, I'm ok with merging this if it fixes an issue...

PR #636, opened 2018-12-12
Change dynamic smartblock to be default

Robbt:
This will make the dynamic smartblock the default choice vs. the static smartblock. Unfortunately this change requires updating the database and various changes to the default behavior. It needs testing and review to make sure the database changes take place on pre-existing instances.

### DIFF
--- a/airtime_mvc/application/controllers/downgrade_sql/airtime_3.0.0-alpha.7/downgrade.sql
+++ b/airtime_mvc/application/controllers/downgrade_sql/airtime_3.0.0-alpha.7/downgrade.sql
@@ -1,0 +1,1 @@
+ALTER TABLE cc_block ALTER COLUMN 'type' SET DEFAULT 'static';
--- a/airtime_mvc/application/controllers/downgrade_sql/airtime_3.0.0-alpha.7/downgrade.sql
+++ b/airtime_mvc/application/controllers/downgrade_sql/airtime_3.0.0-alpha.7/downgrade.sql
@@ -1,1 +1,1 @@
-ALTER TABLE cc_block ALTER COLUMN 'type' SET DEFAULT 'static';
+ALTER TABLE cc_block ALTER COLUMN type SET DEFAULT 'static';
--- a/airtime_mvc/public/js/airtime/library/library.js
+++ b/airtime_mvc/public/js/airtime/library/library.js
@@ -1277,7 +1277,7 @@ var AIRTIME = (function(AIRTIME) {
                     }
                     // remove 'Add to smart block' option if the current
                     // block is dynamic
-                    if ($('input:radio[name=sp_type]:checked').val() === "1") {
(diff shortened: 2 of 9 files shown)


frecuencialibre:
other than the sql tweak above this is working great for me. notes:

- i actually first tried simply making the schema change, _without_ all the 0 / 1 swapping in these commits, but then was seeing errors in zendphp.log regarding it's inability to calculate a length. i wonder if cherry picking `c7767c8` "avoid fail when blocks are created as dynamic by default" would have resolved this; didn't try it. after pulling this branch i'm no longer seeing any errors.
- both dynamic and static blocks behaved as i'd expect them to.
- [other stuff](https://github.com/LibreTime/libretime/pull/627#issuecomment-446538654) also surfaced  as i was clicking around, but i think unrelated to this change.

airtime_mvc/public/js/airtime/library/library.js
```diff
@@ -1277,7 +1277,7 @@ var AIRTIME = (function(AIRTIME) {
                     }
                     // remove 'Add to smart block' option if the current
                     // block is dynamic
-                    if ($('input:radio[name=sp_type]:checked').val() === "1") {
+                    if ($('input:radio[name=sp_type]:checked').val() === "0") {
```
frecuencialibre: i can't see what this actually does... even before this change i see no change in the enable/disabling of the + button above, nor to the contextual menus for tracks. i think this feature was broken, but honestly not a priority for me - people can just use conventional playlists.

screenshot from 2018-12-12 04-33-13

Robbt: If you are editing a smartblock and you navigate the left column to tracks a button at the top says "+ add to smartblock" this isn't grayed out when you have a dynamic smartblock but gives an error message "You cannot add tracks to dynamic blocks." it works for static smartblocks. So I don't think it's a big deal.

airtime_mvc/application/controllers/downgrade_sql/airtime_3.0.0-alpha.7/downgrade.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE cc_block ALTER COLUMN type SET DEFAULT 'static';
```
hairmare: Doesn't this also need an Upgrader class for schema version `3.0.0-alpha.7`? Like one of these:
https://github.com/LibreTime/libretime/blob/5b822afe14046157a8d9fbdff103eb383673e6be/airtime_mvc/application/upgrade/Upgrades.php#L523-L534
I think it won't automatically get applied without one but I'm not sure.

Robbt: I'm not sure. It probably does, I didn't test the upgrade directly. I just pushed the new Upgrader class as a PR.

Issue #633, opened 2018-12-10
Feature Request: Allow podcasts to aquire metadata from RSS feed when files are missing metadata


Robbt:
Basically a number of audio files we receive from a certain provider have no id3 metadata and really non-descriptive filenames. These same files have decent title data and descriptions in the RSS feed itself. 

This could probably be done in celery and would be an easy fix. It would go a long way to making the metadata played out over the feed make sense to listeners vs. inscrutable file names.

JohnnyC1951:
Could we modify the proposal to include the description?

Robbt:
I modified the description to make this more about metadata in general to represent the fact that it could include more than just track title but also description.

Robbt:
So this is more or less a duplicate of #431 - and so if we can address that issue we can solve this one as well. So aside from setting up a proxy solution the question is where the best place to insert this fix is. The way I enabled the addition of album name overriding was by adding additional parameters to the podcast-download celery task and passing a podcastname and override flag. This would get ridiculous if we started doing this for every field we wanted to change AND the podcast_name is from the system vs. the RSS feed so it is readily available. 

I don't think that the actual episodes RSS metadata ie title & description etc aren't stored in an easily accessible spot in the database, so if I took a similar approach and passed them as a parameter it would require modifying a bit of code and adding the title and description to the podcast_episode database table. 

The right approach might be to add this data to podcast_episodes and then intercept the callback where podcast_episodes is updated and then modify the metadata from inside of PHP vs. passing stuff to python and celery. That's what I'll investigate next.

Robbt:
So this is somewhat working, it might need refinement and could use some testing. But basically the newest PR pulls the title from the feed and if album override is enabled it will override the title for the track with the one from the RSS feed. We should probably update the UI to show album & title override or decide if we want to have a separate flag for title override in the options etc. 

I personally am of the opinion that overriding it all at once as one option is probably a good idea and we just change it to metadata override. But I'd like to hear other peoples thoughts.

JohnnyC1951:
See my recent comments in other threads. There is overlap.

PR #637, opened 2018-12-12
give user ability to edit podcast name

### LINKED ISSUE #631: Problem: multiple podcasts from same domain cannot be differentiated
**Problem:** 
the current [album override method](https://github.com/LibreTime/libretime/issues/89) for querying downloaded podcasts with a smartblock uses the domain from which they were downloaded to set the album. 
at our station we need several podcasts uploaded to archive.org. eg: 
https://archive.org/services/collection-rss.php?query=uploader:noticiasdeabajoml@gmail.com
https://archive.org/services/collection-rss.php?query=uploader:latinoamericarompeelcerco@gmail.com
both are giving the same album metadata, and so are impossible to program separately. 

**Proposed solution:** 
from a user's perspective, it would be great to be able to manually specify a text string that would be set in the downloaded podcasts' album metadata. currently a modal dialogue asks you for a url, which then cannot be edited. I'd of course defer to you @Robbt regarding whether any of this is feasible, but think a minimal solution would be to add a "Podcast name" field to the modal. Better yet would be to make the interface more closely match that of playlists and smartblocks: that instead of a modal, it look something like the screenshot below, and enable editing after initial creation. This however is all above and beyond the minimum we'd need as a station to get past this blocker.  thoughts? thanks!

screenshot from 2018-12-03 17-04-53

Robbt:
This fixes #631 

It makes the podcast name an editable field, which will allow users to differentiate if different podcasts have the same name for some reason.

The only question remains how to handle the auto-generated smartblocks. Some options:
1. add a button that says generate automatic smartblocks and playlists for this podcast. 
2. check the options and do this automatically but that could result in a lot of different playlists if people changed the name multiple times and I don't think we want to automatically delete old ones.
3. flash a message on the screen letting people know that they will need to manually update smartblocks that matched the old podcast name.

### DIFF
--- a/airtime_mvc/application/services/PodcastService.php
+++ b/airtime_mvc/application/services/PodcastService.php
@@ -166,15 +166,15 @@ class Application_Service_PodcastService
      * This will automatically create a smartblock and playlist for this podcast.
      */
 
-    public static function createPodcastSmartblockAndPlaylist($podcast, $title = "Podcast")
+    public static function createPodcastSmartblockAndPlaylist($podcast, $title = null)
     {
         if (is_array($podcast)) {
             $newpodcast = new Podcast();
             $newpodcast->fromArray($podcast, BasePeer::TYPE_FIELDNAME);
             $podcast = $newpodcast;
-            if ($title == "Podcast") {
-                $title = $podcast->getDbTitle();
-            }
+        }
+        if ($title == null) {
+            $title = $podcast->getDbTitle();
         }
         // Base class
         $newBl = new Application_Model_Block();


JohnnyC1951:
Having done it, and doing it 3 times an hour, Iterating the smartblocks is pretty easy. Couldn’t it just be changed to match?

frecuencialibre:
awesome!!! 

i hadn't thought of some of these options, but good thinking. personally, i like the button to generate them. it's explicit, prominent. not lost in settings you never knew existed because you never read the documentation

i also agree it's best to be cautious about making changes to exiting ones or worse deleting them. i know we for example will probably be modifying automatically generated smartblocks, to add additional content via additional search criteria.

> Iterating the smartblocks is pretty easy. Couldn’t it just be changed to match?

what do you mean by "iterating"? is this related to option 2? (i just edited the comment to enumerate 'em)

JohnnyC1951:
I will butt out, I think.  I handle Podcasts quite differently to the plan vanilla LT version. I bring the podcast description (and title), via cc_files, right into the shows, using witchcraft and a big pinch of non-celery salt. I have been doing this since AT 2.3. I will keep out and work around whatever you guys end up with.

Robbt:
So I pushed a button to generate a smartblock - it required adding a whole new endpoint to the rest module and modifying the existing smartblock code. 

Issues:  if you haven't saved the new podcast name it will generate the smartblock for the previous name of the podcast ie the one that is saved in the database. So it is likely that people will hit this before saving and wonder why the podcast has the old name.

Also it gives no feedback so they could hit it a bunch of times and go - it's not working - and then find like 30 smartblocks

So solutions.
- Provide some kind of message that a smartblock has been created and display the podcast name.
- Show code to detect if the smartblock name is different than what was in the database and warn the user that they need to save first. (not sure how easy it would be to do this)
- Prevent the automatic smartblock code from creating duplicate smartblocks (would need to do even more work to query the database for smartblock with the name of podcast already)
- Save the podcast name change when they click smartblock generate (not sure if this can be done in one action)

frecuencialibre:
i needed to add a closing bracket } on line 59 of podcast.js

frecuencialibre:
ok, first up a personal thank you again for jumping on this one since it's a blocker for my station. :)

this is already working great for me, so IMO we already have a minimum merge-able product: tracks download and have their album set to the name i specify; the button also successfully creates a new smartblock that queries episodes using with the saved name. 

...and as usual i have a laundry list of UI improvements that _could_ be made should you evaluate that they'd be worth the effort. 

but first regarding your bullet point solutions above:
- i see that we could easily pass `$scope.podcast.title` from the `createSmartblock` js function to the API endpoint. how complicated would it be for that to be the value used to create the smartblock/playlist?
- i agree that we need some visual feedback. the established visual language is something like [this](https://github.com/LibreTime/libretime/blob/5b822afe14046157a8d9fbdff103eb383673e6be/airtime_mvc/public/js/airtime/library/spl.js#L901) i think.
- if we do these two things then i don't think it'd be necessary to get into checking whether the podcast title is identical.

and now my podcast UI laundry list... in general my goal is to make the podcast creation/editing workflow match the established visual language. as i stated above, please take these with a grain of salt - all are optional IMO, nothing critical that would warrant sacrificing our health on the altar of the almighty zend 1. in no particular order:
1. i was going to submit a PR from my fork to yours, but it was going to include a million unrelated commits. please just see [this commit](https://github.com/frecuencialibre/libretime/commit/15b73580c11f242dfe0d56ccf1bdc3aea9640518).
2. how complicated would it be to abandon the create podcast popup (which doesn't exist anywhere else in libretime) for the standard edit pane? currently when you submit the popup you see the list of episodes, but it isn't then clear that if you want to edit your podcast name and settings you have to click the "back" chevron button and then "edit".
3. how complicated would it be to put the podcast episodes below the podcast edit fields? this would match the playlist/smartblock content creation experience by putting content on the right. we could then remove the chevron buttons on the left which i've also seen cause confusion. i'm thinking something like this:
screenshot from 2018-12-14 06-21-05
Save etc. buttons are below the content "preview" table in playlists and smartblocks, so we should probably respect that now that i'm thinking of it. putting these buttons _both_ above and below the preview table is actually yet another UI improvement i'd like to do for the 3 relevant content types (sb/pl/pc).
4. when you save a podcast, rather than getting a little green success message by the button, the pane closes.
5. i've had it happen several times that something goes wrong with the importation of a podcast episode and it then spins for all eternity. any way to be able to cancel import?
6. how complicated would it be to make the url editable on the edit pane? (not really a common use case. this is just to match standard UI. low priority) depending on your answer to this one we should also fix css so as to not cut off the "Override album name with podcast name..." option text.
7. "Override album name with podcast name during ingest." could be either just Override album name with podcast name" or to stick with the verb import already used on the blue button.
8. if i'm understanding correctly, the smartblock and playlist will only really work if it can query the automatically set album name. hence the generation of these is dependent on that setting. perhaps we should consider conditionally hiding/disabling that button, and somehow visually connecting it to the setting checkbox.
9. even though my global "Podcast Album Override" setting remains checked by default, the "Override album name with podcast name" individual podcast setting is unchecked when i create a new podcast. the behavior i expected is that the global setting determine the default upon new podcast creation, but that the individual podcast setting is the only one that actually has any effect. also language in these two should be uniform.

whew! sorry that got long, beyond the scope of this pr, and probably related to other issues in the queue. looking forward to hearing your triage thoughts.

Robbt:
Those are all good ideas and insightful critiques of how things are setup. I am not really qualified to pick this apart and implement them though because I don't have a lot of experience with Angular 1 which the SourceFabric developer decided to use when implementing podcasts.

So for 1 - I think you can push commits onto my PR directly as you are a maintainer so you don't need to open up a separate PR just check out my PR and then make that change and push it to my remote repo and it'll be reflected here.

2. I could figure this out but I'm not sure its worth the effort it would take.
3. I like this idea but it would involve deciphering and reimplementing the podcast/angular stuff
4. Changing this probably just involves changing the callback in podcast.js for save (I did see this)
5. This might be worth investigating when I'm figuring out how to pull in the title etc. in #633 
6. Editing the URL is probably a bad idea unless we incorporate some of the other changes and do some kind of advanced checking to ensure they don't accidentally mess it up.
7. Yeah this is confusing because it is only used if the global is set off, so perhaps the best thing to do would be to have this shown as checked if the global is in place, but not sure how much more work that would generate
8. Yeah this should probably require the override to be in place and check for it
9. I have never checked this and the global one always works for me, this was added as a way for people to just choose individual podcasts to override but it isn't clear so we can clarify and test it further. I'd say if the global setting is checked this should be shown as checked and grayed out with maybe a tooltip that explains it. But lets prioritize making it workl.

frecuencialibre:
Ok so the commits I've pushed should resolve the concerns @Robbt [expressed above](https://github.com/LibreTime/libretime/pull/637#issuecomment-446837134) regarding the user hitting the button a bunch of times, etc. I also resolved a few quickies. I'm now signing off that this is ready to merge.

updated laundry list stati:
1. done
2. popup - beyond this PR's scope
3. podcast episodes below edit fields - beyond this PR's scope
4. done
5. cancel podcast import process - beyond this PR's scope
6. change url - bad idea, won't do
7. wording - please review. i added a tooltip as well, but haven't touched anything related to the global settings.
8. generate button dependency on album overwrite setting - done
9. global v. individual podcast setting coherency - beyond this PR's scope

Robbt:
This looks good. I just noticed in testing this that it is breaking the autogeneration of smartblocks on a global level and is creating playlists and smartblocks based upon the title of Podcast vs. the actual podcast name. So that seems like an unintentional consequence of the way the refactoring of the function createPodcastSmartblockAndPlaylist($podcast, $title = "Podcast").

I'll take a stab at fixing this so it can work either way. 

I also think we might want to add a check to make the checkbox hardcoded if the global setting is turned on and possibly revisit whether we want to have the automatic creation of smartblocks for podcasts turned on by default. In some ways this seems like a cleaner approach but it would require users to know they can go into settings to do this, vs. it just happening. More to do, I'll see if I can fix the global setting w/o mucking up the improvements made.

Robbt:
Ok, I fixed this, I realized the issue was simply that you had the check for $title="Podcast" inside of the code block that was designed to hydrate the Podcast object when the user was pushing the button, as the podcast object is already instantiated when the global function is called that was never being called. I took it out of the block and also changed it to null vs. a string because I think that is more specific and will prevent issues if there is a podcast with the name of "Podcast". But otherwise I think things are working now and this can be merged. I'm not sure the intended protocol of the C4 when there are multiple contributors to a PR and whether we can or should merge it or expect someone else who hasn't contributed to merge it. But I don't think it matters all that much at this point. I'll hold of on merging for now to give someone else a chance to review it.

frecuencialibre:
thanks for the cleanup. yeah that way is better. also i don't know if you saw that the clicking the "generate..." button actually triggers two backend actions: sb/pl generation, and also saving the podcast. hopefully this way hopefully we avoid any discrepancies between the saved podcast and the generated smartblock.

PR #619, opened 2018-11-28
Remove the version restriction on mutagen

Robbt:
This fixes #610 and possibly other issues regarding mutagen. In theory it could introduce new bugs but I suspect the version restriction wasn't necessary and was geared towards simply avoiding the need to rewrite tests.  We can respond to new issues if we are able to find them but this should fix a number of potential issues based upon the number of [bug fixes](https://mutagen.readthedocs.io/en/latest/changelog.html) from 1.31 to 1.41.

PR #639, opened 2018-12-13
Fix nonmp3 podcast download

### LINKED ISSUE #271: Problem: certain podcasts fail to download files with "can't sync to an MPEG frame" error

### LINKED ISSUE #519: Downloaded m4a podcast's failing to import with "can't sync to an MPEG frame" error
As the subject suggests, am having issues with podcasts that are in M4A format importing to Libretime.

Summary:

- Files are being created in /tmp when adding podcast (that's m4a format) and being downloaded
- File appears to be fully downloaded ok, but then failing to import to library.
- Files in /tmp/ are owned by "celery:celery"
- Example Podcast: http://podcast.djhardwell.com/podcast.xml
- I'm seeing the following in one of the celery worker logs:

```
[2018-09-18 11:41:03,657: INFO/Worker-1] podcast-download[php_5ba0e4352a6c74.76761676]: Error during file download: can't sync to an MPEG frame
[2018-09-18 11:41:03,718: INFO/MainProcess] Task podcast-download[php_5ba0e4352a6c74.76761676] succeeded in 26.512139958s: '{"status": 0, "episodeid": 223, "error": "can\'t sync to an MPEG frame"}'
```

I've read that the python `mutagen` package required by `airtime_analyzer-0.1-py2.7.egg` could be a cause (it's currently mutagen==1.31) - I've tried upgrading it to the latest (mutagen 1.41.1) and it's no help unfortunately.

System details:
- Running Ubuntu Xenial
- silan 0.3.2-2build3
- Python 2.7.12
- Celery args are: `--time-limit=1200 --concurrency=4 --config=celeryconfig -l INFO`

Would welcome suggestions on where to look / potential fixes - Anyone know if merging in https://github.com/LibreTime/libretime/pull/464 will help here?

Regular MP3's are fine - Could we perhaps just find a way to pipe any other formats through something like `ffmpeg` before having the files imported for now which should fix this (not sure where to look, or I'd already be trying this now)

I believe what may be cause of the problem is that the podcast's "album art" actually changes multiple times throughout the mix (as the track names change, the album art updates to reflect the track names) - This is pretty common with EDM podcasts from big name artists/shows (ASOT, Tiesto etc) when you look at the podcasts they have on iTunes.

Robbt:
This fixes #519 by taking away the mp3 specific code and replacing it with mutagen library as a whole which supports most filetypes except for WAV, and I don't think there are very many WAV based podcasts.
I tested it with the m4a based podcast referenced in #519 and was able to download without an issue whereas before it was failing.

Also fixes #271

### DIFF
--- a/python_apps/airtime-celery/airtime-celery/tasks.py
+++ b/python_apps/airtime-celery/airtime-celery/tasks.py
@@ -9,9 +9,7 @@ import posixpath
 import shutil
 import tempfile
 import traceback
-from mutagen.mp3 import MP3
-from mutagen.easyid3 import EasyID3
-import mutagen.id3
+import mutagen
 from StringIO import StringIO
 from celery import Celery
 from celery.utils.log import get_task_logger
@@ -155,15 +153,12 @@ def podcast_download(id, url, callback_url, api_key, podcast_name, album_overrid
             with tempfile.NamedTemporaryFile(mode ='wb+', delete=False) as audiofile:
                 r.raw.decode_content = True
                 shutil.copyfileobj(r.raw, audiofile)
-                # currently hardcoded for mp3s may want to add support for oggs etc
-                m = MP3(audiofile.name, ID3=EasyID3)
-                logger.debug('podcast_download loaded mp3 {0}'.format(audiofile.name))
-
+                metadata_audiofile = mutagen.File(audiofile.name, easy=True)
+                logger.debug('podcast_download loaded {0}'.format(audiofile.name))
                 # replace album title as needed
-                m = podcast_override_album(m, podcast_name, album_override)
-
-                m.save()
-                filetypeinfo = m.pprint()
+                metadata_audiofile = podcast_override_album(metadata_audiofile, podcast_name, album_override)
+                metadata_audiofile.save()
+                filetypeinfo = metadata_audiofile.pprint()
                 logger.info('filetypeinfo is {0}'.format(filetypeinfo.encode('ascii', 'ignore')))
                 re = requests.post(callback_url, files={'file': (filename, open(audiofile.name, 'rb'))}, auth=requests.auth.HTTPBasicAuth(api_key, ''))
         re.raise_for_status()
--- a/python_apps/airtime-celery/airtime-celery/tasks.py
+++ b/python_apps/airtime-celery/airtime-celery/tasks.py
@@ -154,7 +154,6 @@ def podcast_download(id, url, callback_url, api_key, podcast_name, album_overrid
                 r.raw.decode_content = True
                 shutil.copyfileobj(r.raw, audiofile)
                 metadata_audiofile = mutagen.File(audiofile.name, easy=True)
-                logger.debug('podcast_download loaded {0}'.format(audiofile.name))
                 # replace album title as needed
                 metadata_audiofile = podcast_override_album(metadata_audiofile, podcast_name, album_override)
                 metadata_audiofile.save()
--- a/python_apps/airtime_analyzer/setup.py
+++ b/python_apps/airtime_analyzer/setup.py
@@ -28,7 +28,7 @@ setup(name='airtime_analyzer',
       packages=['airtime_analyzer'],
       scripts=['bin/airtime_analyzer'],
       install_requires=[
-          'mutagen', # got rid of specific version requirement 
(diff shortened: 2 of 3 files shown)


Robbt:
ugh it looks like I forgot to checkout of the remove mutagen 1.31 version PR before I did this code change. So that is getting pulled in. It shouldn't be an issue if we merge #619 before this one.

hairmare:
I've been holding off on merging this since I'd like to do some testing. At a quick glance and also considering the mutagen docs it looks more than fine.

frecuencialibre:
did some testing on this one, and unfortunately the jury is still out. 

first, without a re-install, i simply checked out this PR branch, and then tested two podcasts, the hardwell one above, and an [mp3 one](https://archive.org/services/collection-rss.php?query=uploader:noticiasdeabajoml@gmail.com) we use. ... and manual import of the most recent episodes (following the steps in the wiki) of both podcasts worked!

great, but then i tried a fresh install. the mp3 podcast still worked, but the m4a one failed with the same `can't sync to an MPEG frame` error in zendphp.log. more info:
- i noticed that i [apparently was still running mutagen 1.31](https://github.com/LibreTime/libretime/pull/619#issuecomment-448720417).
- python_apps/airtime_analyzer/setup.py does in my tested codebase does show the removal of the mutagen version restriction
- manual import as an individual track through the web UI of the [first episode of the m4a podcast](http://test.podcast.djhardwell.com/HOA250.m4a) worked. 

any suggestions on how to further troubleshoot? manually update mutagen?

Robbt:
I noticed it indeed was at 1.31 on my ubuntu-xenial, perhaps this is because this version of mutagen is provided by the distro by default and so it uses that vs. downloading the newest version from pip.

I decided to test pushing the version manually up to 1.41.1 or above in the requirements and it is still failing now for me as well, which is odd considering I had it working previously. 

I'm very perplexed and I'm going to look at the code to see what could have happened.

Robbt:
Ok, I think I have figured this out, for some reason the line `logger.debug('podcast_download loaded {0}'.format(audiofile.name))` was causing the error for me but when I removed that everything worked even with mutagen 1.31 - I am suggesting we keep the mutagen 1.41.1 though because 1.31.0 had some other errors and is still coming by default. I'm not sure why as I can't find any places referencing it. It is kind of a messy commit to modify the airtime_analyzer code in fixing this but since I already had that commit log bundled I don't think it will cause too much confusion to anyone browsing it in the future.

@frecuencialibre can you see if this new version works for you with the m4a podcast ?

hairmare:
On a fresh debian-stretch vagrant machine the pypo install seems to install mutagen==1.36 which then gets updated to mutagen==1.41.1 as soon as the installer reaches the airtime_analyzer dir. I'm not sure why it's doing this since pypo doesn't actually specify a mutagen version (I would expect mutagen to install in a latest version without any requirements but I don't seem to understand pip reqs properly 😛).

I just imported a head of tracks and some podcasts to my test instance and will do some listening to be sure everything still looks sane before I merge this. I think we can address what's happening with mutagen installed by pypo vs analyzer in a later PR if needed.

BTW the only distro that already has  mutagen 1.41.x is Fedora 29 which is at 1.41.0. Debian and Ubuntu both have 1.40.x in their latest testing repos ([source](https://pkgs.org/download/mutagen)).

hairmare:
This also fixes non mp3 formats for non podcast cases... ie. if you can upload it and it has metadata that are reachable by mutagen then those will get considered.

PR #640, opened 2018-12-13
Smartblock ui: add collapsed "advanced options" section

frecuencialibre:
this PR hides a bunch of settings in the Smartblock edit pane that I've seen confuse users within a collapsible "Advanced options" section:

screenshot from 2018-12-13 09-29-04

screenshot from 2018-12-13 15-07-57

### DIFF
--- a/airtime_mvc/application/views/scripts/form/smart-block-criteria.phtml
+++ b/airtime_mvc/application/views/scripts/form/smart-block-criteria.phtml
@@ -1,87 +1,8 @@
 <form class="smart-block-form" method="post" action="">
     <dl class='zend_form search-criteria'>
         <div id='sp-success' class='success' style='display:none'></div>
-        <dt>
-            <label class='sp-label'>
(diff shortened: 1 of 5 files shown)


Robbt:
This looks good but I noticed there was a lot of spacing between the preview and the advanced tracks dropdown on firefox.
If you remove this line it seems to fix it. I didn't test what impact this had on chrome.
https://github.com/LibreTime/libretime/blob/4752c728b9411b6aa5007b34884cc68e8f9c0c7e/airtime_mvc/public/css/dashboard.css#L593

Here is how it shows up without removing it the min-height of 6em.
withoutfix
here it is with removing that line
withfixscreenshot

It pushes it down properly when the advanced options is expanded as well. So I'd recommend that change but otherwise I think it is a great improvement in terms of making the interface simpler for people to learn.

frecuencialibre:
hopefully the commit i just fixed pushed fixes that spacing issue. that min-height is actually important for smaller screens, so i fixed this in another way.

PR #645, opened 2018-12-17
adding phing 2.12.0 to composer.json

### LINKED ISSUE #642: Problem: Include phing in vendor when building for release
When the release tarball is built, it includes all php dependencies from composer under `vendor/`. The `composer.lock` references the `phing` dependency, which is missing from `composer.json`. Because of this it is downloaded when running composer installer. Bundling it in with all the other dependencies would allow me to build Debian/Ubuntu packages without requiring the build container having access to the Internet.

Robbt:
This fixes #642 

Not sure if we want to pin phing to this specific version but I know 3.x is Php 7.1 and above and 2.12.0 was installed by default by my ubuntu-xenial install. There are also 2.x versions up to  2.16.1. I also didn't test this but just wanted to toss in the code up there.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,6 +15,7 @@
         "zendframework/zendframework1": "^1.12",
         "composer/semver": "^1.4",
         "php-amqplib/php-amqplib": "^2.6"
+        "phing/phing": "2.12.0"
     },
     "require-dev": {
         "phpunit/phpunit": "^4.3",


paddatrapper:
I think 2.12.0 was the version that ws installed while I was building packages too

hairmare:
We don't use phing anywhere, it's a dependency of zf1, but our case doesn't use it. See https://github.com/LibreTime/libretime/issues/642#issuecomment-447989534 for details.

Robbt:
Ok, I'm going to defer to lucas on this and close it. @paddatrapper let me know if we need to try to address this an alternative way.

PR #649, opened 2018-12-18
removed the recaptcha

### LINKED ISSUE #506: Problem: LibreTime uses a ReCaptha  Version 1 that has been discontinued
If a user enters the wrong login, LibreTime will ask for a recaptcha but the recaptcha it is using is no longer valid and so it fails.

Robbt:
This fixes #506 - it just removes all of the recaptcha code because it depended upon a defunct service and was ineffective.


hairmare:
Awesome this also gets rid of quite some $_SERVER usage.

PR #658, opened 2018-12-23
Modify javascript to bring focus to search window by default

### LINKED ISSUE #339: Problem: The cursor doesn't default to the track search bar upon page load
This is another persistent request from one of our users that should improve usability. 

Basically the idea is that upon page load the cursor should default to the search bar.
Thus if someone is looking at the tracks page and they start typing then what they are typing will show up in the search box.

This seems like a logical default position for the cursor and would avoid an extra click for most use cases.

Robbt:
Fixes #339 
Thanks to @jerry924 for the initial suggestions for how to fix this.

### DIFF
--- a/airtime_mvc/public/js/airtime/library/library.js
+++ b/airtime_mvc/public/js/airtime/library/library.js
@@ -942,6 +942,25 @@ var AIRTIME = (function(AIRTIME) {
 
         var simpleSearchText;
 
+        /* set focus into the input box when page loads */
+        $(window).ready(function(){
+            $libContent.find("#library_display_filter label").find('input').focus();
+        } );
+        /* if the user clicks on an a anchor link ie most navigation buttons or there is no focus
+         * then bring focus back to the search window
+         * this should handle most navigation cases for the site currently without taking focus away from
+         * other input fields or otherwise interfering with the UI interaction */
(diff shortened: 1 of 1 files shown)


airtime_mvc/public/js/airtime/library/library.js
```diff
@@ -942,6 +942,25 @@ var AIRTIME = (function(AIRTIME) {
...
 
+        /* set focus into the input box when page loads */
+        $(window).ready(function(){
+            $libContent.find("#library_display_filter label").find('input').focus();
+        } );
+        /* if the user clicks on an a anchor link ie most navigation buttons or there is no focus
+         * then bring focus back to the search window
+         * this should handle most navigation cases for the site currently without taking focus away from
+         * other input fields or otherwise interfering with the UI interaction */
+        $(window).click(function(){
```
frecuencialibre: i'm wondering if we could simplify this to just use click events on the "showbuilder" main menu links. if we add the event to a container div rather than the link then pretty sure we won't interrupt anything, just take advantage of the event "bubbling" up:
```
        $(".media_type_selector").click(function() {
            $libContent.find("#library_display_filter label").find('input').focus();
        });
```

Robbt:
This was fixed in #664

Issue #526, opened 2018-10-01
Use standard ID3 "comment" metatag field instead of "description"


frecuencialibre:
Hi all, 

As we've been cleaning up audio file metadata and importing our radio's library into libretime, we've come up against an issue that is complicating our efforts to import music and then quickly add it to playlists/smart blocks. 

It seems he standard ID3 "comment" field is not searchable by Libretime. Can anyone else confirm this? 

Instead, I see a "description" field, which is searchable, but which doesn't exist within [standard ID3 tag fields](https://en.wikipedia.org/wiki/ID3), and so cannot be easily added by third party metatag editors.

Our use case, which IMHO could be streamlined greatly in the UI at some point, is that radio collective members are trying to import music _directly_ into a scheduled show. Currently the best way we've found is to add custom metatags to files prior to import, and then use either a playlist or a smart block to query our custom tags. It would be great to be able to use the comment field rather than mangling the genre, for example. 

thanks!
-ryan

hairmare:
The `comment` field should get mapped somewhere. I did some digging in the code but wasn't able to find where. I always thought it would end up in the description field on the ui but never checked.

It looks like analyzer isn't doing anything to the tag:
https://github.com/LibreTime/libretime/blob/1b273062c47498fbed8e5fa0490e39a5dbeb37a9/python_apps/airtime_analyzer/airtime_analyzer/metadata_analyzer.py#L130

Analyzer then goes on to send it's data to the `CcFiles::updateFromArray()` method by way of the "modern" RESTful API where it gets processed by this:
https://github.com/LibreTime/libretime/blob/5b822afe14046157a8d9fbdff103eb383673e6be/airtime_mvc/application/models/airtime/CcFiles.php#L189-L239

I haven't figured out where/why it isn't storing the comment in the description or what is actually up here.

Robbt:
So LibreTime also apparently doesn't save any modifications to metadata made from inside of the program back to the file's id3 tag. Once it is imported any changes made will happen only inside of the database. Would it make sense to try to write back changes in the UI to the files ? I can see a few issues with this but thought I'd bring it up for discussion as it is somewhat relevant to this.

JohnnyC1951:
I use the 'Composer' and 'Encoder' fields in my RSS importer scripts. Not pretty, but it works.

hairmare:
> So LibreTime also apparently doesn't save any modifications to metadata made from inside of the program back to the file's id3 tag.

[...]

PR #659, opened 2018-12-23
added track title override for podcasts

### LINKED ISSUE #633: Feature Request: Allow podcasts to aquire metadata from RSS feed when files are missing metadata

Robbt:
This fixes #633 
It could be merged as is but we might want to refine it further.
Basically it jumps on the album_override and also overrides the track title with the title from the RSS feed.
The easiest way to do this was to pull the info from the episodes array where it resides.

I need to test it also to see if it works for podcasts that are imported by the scheduler when new tracks are found vs. done manually via the UI (this is working).

So we should decide if we want to make title overriding a separate option or just change the description of album override to include track title override and maybe make it metadata override. In almost all cases the RSS feed will have better or at least as good of metadata as the title mp3 and in many cases it will be superior.  

I also didn't import the description because it isn't tied to any metadata and comments aren't used currently but this would make sense.


JohnnyC1951:
I make my titles '$ALBUM - $TRACKTITLE'. It works well for me.
It looks like 'Science Weekly - Ice crater found on Mars'

frecuencialibre:
@JohnnyC1951 is `Robbt:podcast-metadata-override` working for you? specifically i see we need to test "podcasts that are imported by the scheduler when new tracks are found vs. done manually via the UI". i'm neck deep in podcast js right now...

> So we should decide if we want to make title overriding a separate option or just change the description of album override to include track title override and maybe make it metadata override. In almost all cases the RSS feed will have better or at least as good of metadata as the title mp3 and in many cases it will be superior.

i vote for a simplified single setting that enable both the album override as well as this metadata for the stream. regarding Johnny's "$ALBUM - $TRACKTITLE" pattern, could it be possible that the metadata that the smartblock searches for, and which is now configurable by the user, could be the same that is also shown in stream metadata (eg. contains title? artist?). just brainstorming, not sure if that's a good idea, and i've never looked at the stream metadata. does it include the album?

side tangent: i believe we still need to enable this by default, correct? do we need a separate issue for that one?

> I also didn't import the description because it isn't tied to any metadata and comments aren't used currently but this would make sense.

IMO if comments aren't shown in the stream metadata then that's beyond scope of this PR.

Robbt:
It's beyond the scope of this particular PR because we need to fix the comments/description issue highlighted in #526 for this to work. The issue there is that description exists in the database whereas this works by modifying the ID3 tag and there is no ID3 tag for description only comments. So we can revisit this or add a fix to import the description when we get around to resolving the whole comments being invisible and description not being tied to metadata. 

Todos before we merge this
- [ ] Test that it imports podcasts that are done by celery when a RSS feed is updated vs manual import
- [ ] Update the UI interface to reflect that podcasts override updates album & track title
- [ ] Update the docs to reflect this change (or write new docs for this)

JohnnyC1951:
I worked around that years ago. Glad folks are catching up.

Robbt:
I can confirm that indeed this doesn't work for podcasts that automatically import because addPlaceholder creates a PodcastObject and the podcastObject doesn't have the track title stored. So for this to work we need to modify the database to store the podcast title as I originally thought.

frecuencialibre:
Updating here based on conversations with @Robbt, the metadata fields shown by LibreTime in the stream are the tracks Artist and Title. Other info, such as the Station name and Show name can also be shown, but that isn't relevant here.

Hence a solution here would be to override the Artist (aka creator) field, and use it in the smartblock query. We also need to account for backwards compatibility with existing smartblocks that query the album field, possibly querying artist `OR` album. Additionally, we agreed that overriding track title with the feed item title would be a good idea since the feed title is almost always better. 

@Robbt is this correct?

Robbt:
Ok, so ignoring the various PRs and commits I pulled along. I modified the database schema and regenerated it to include episode titles and descriptions.
I also had the podcast_album_override write both the author & album to = podcast_title and track_title = episode title. description is pending the change over from comments mentioned in a different PR.

Robbt:
In addition I'm going changing the preference settings from podcast album override to "podcast metadata override" and including this short description "Enabling this means that podcast tracks will get their metadata set from the podcast feed values"

frecuencialibre:
awesome! so you mean the label and tooltip beside the "Overwrite..." checkbox?

frecuencialibre:
ah no you mean the site wide global settings, not the podcast specific one. 

which reminds me - how will those two settings work together? i had assumed that the global is the default to which all new podcasts are set, but then the only setting that is actually used to determine whether or not to overwrite is the one on the individual podcast. or some other better way?

Robbt:
The individual override is only used if the global setting is turned off. I tested it and it appears to work. I have tested this and it was working but I noticed that if you create and delete a podcast track and then try to delete it from inside of the Podcast view it will produce an error.

The other thing to note is that if a track lacks Album metadata it will get the podcast name regardless of the setting. I don't know if we want to make this the case for the track and artist/creator as well. Ie check and see if it is missing and if so then implement the metadata from the feed.

`2019-01-13T16:54:05+00:00 ERR (3): localhost [ErrorController.php:54 - errorAction()] - An internal application error has occurred.: Exception: Could not recall file with id: 13 in /vagrant/airtime_mvc/application/models/StoredFile.php:601
Stack trace:
#0 /vagrant/airtime_mvc/application/controllers/LibraryController.php(260): Application_Model_StoredFile::RecallById(13)
#1 /vagrant/vendor/zendframework/zendframework1/library/Zend/Controller/Action.php(516): LibraryController->deleteAction()`
Basically it can't delete a file that no longer exists. I don't think this error was introduced by this PR though.

One oddity that I think is possibly related to this above issue is the continually spinning download button when a track is added and deleted repeatedly even if it is actually downloaded it still has an issue where it thinks that the track wasn't downloaded possibly because a new guid is created locally but the podcast episodes refers to a different one in the library.

I changed the language for the individual podcast episode override as well.

Robbt:
I created a bug for the issue I found with deleting podcast tracks at #689.
This is ready for testing and merge and any final suggestions in terms of UI language etc.

frecuencialibre:
> This is ready for testing and merge

so in the [todo list above](https://github.com/LibreTime/libretime/pull/659#issuecomment-450185347) it looks like the schema update is now done, and so now we need to update UI, and do docs. in order to finalize UI, would you say we should include #690 within the scope of this PR?

Robbt:
So yeah the documentation hasn't been changed but I also don't know that we even have documentation for Podcasts at this point. Let's do the documentation once we resolve #690 since that will be more clear and this includes the changes in the UI for what it does.

frecuencialibre:
i'm seeing the podcast import celery task eventually time out:

```
2019-01-19T21:33:56+00:00 WARN (4): fl.local Celery task podcast-download with ID php_5c43943ecf2335.95191751 timed out
```

following this throwing this error in zendphp.log:
```
2019-01-19T21:18:55+00:00 INFO (6): fl.local json_decode() expects parameter 1 to be string, object given
```

also, the upgrade sql change is to simply remove `NOT NULL` from both column additions?

frecuencialibre:
ok i finally got this up and running on a fresh vagrant ubuntu 16 instance from the PR branch. works great!: 

> Basically it jumps on the album_override and also overrides the track title with the title from the RSS feed.

yup. specifically:

1. the track title is set to the podcast episode title. (<item><title>this here</title></item> in the xml)
2. the track artist is set to the podcast name set in the libretime UI
3. whether or not this happens is dependent on the global setting. the individual podcast checkbox setting changes nothing.

i think this fits the description of what this PR sets out to do. # 3 above is a problem that's already present and which should be resolved in a separate PR as per [our discussion](https://github.com/LibreTime/libretime/issues/690#issuecomment-454115994).

PR #669, opened 2018-12-27
[wip] modified the code to treat a file mutagen fails to load as a mp3

### LINKED ISSUE #668: Mp3 files not detected as mp3 by mutagen fail to import
**Describe the bug**
Basically mutagen fails to detect certain mp3 files using Mutagen.File(audiofile.name) because something is up with their mp3 data, but the files import fine via the web interface this just fails because of the podcast override rename function.

**To Reproduce**
Add this RSS feed as a podcast - https://archive.org/services/collection-rss.php?query=uploader:latinoamericarompeelcerco@gmail.com
Download no 6 (another issue with URL encoding of # signs on archive.org is preventing some redirects from working.
It fails to import with 'NoneType' object does not support item assignment in the airtime-celery message

**Expected behavior**
It should download the file and apply the podcast rename
**LibreTime version:**
latest git aka alpha.6

**Installation method:**
 - OS: Ubuntu
 - OS Version= 16.04.5 LTS (Xenial Xerus)] 
 - Method: vagrant 
 - Details: latest git
**Additional context**
This RSS feed worked before the refactoring of mutagen code to not be MP3 specific.

Robbt:
This fixes #668

It basically attempts to treat any file that fails to be detected by Mutagen as a MP3 as the files were treated previously to work around files that might not be detected correctly for some reason.
It might make sense to add additional logic checks to determine that a file at least has mp3 filename extension etc, so I want to do that before we merge this.

### DIFF
--- a/python_apps/airtime-celery/airtime-celery/tasks.py
+++ b/python_apps/airtime-celery/airtime-celery/tasks.py
@@ -153,8 +153,14 @@ def podcast_download(id, url, callback_url, api_key, podcast_name, album_overrid
             with tempfile.NamedTemporaryFile(mode ='wb+', delete=False) as audiofile:
                 r.raw.decode_content = True
                 shutil.copyfileobj(r.raw, audiofile)
+                # mutagen should be able to guess the write file type
                 metadata_audiofile = mutagen.File(audiofile.name, easy=True)
-                # replace album title as needed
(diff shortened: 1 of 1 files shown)


xabispacebiker:
This saved my day ;)

PR #670, opened 2018-12-29
Use zf1s packages to gain php 7.2 compatibility

### LINKED ISSUE #580: Problem: ZendFramework 1 is incompatible with Php 7.2 and thus Ubuntu 18.04

hairmare:
This PR replaces the zend/zendframework1 dependency with composer packages from the [zf1s orga](https://github.com/zf1s/). The zf1s packages already contain loads of fixes to allow zend framework 1 to run on php 7.2 and the maintainer is willing to merge more PHP 7.2 compat changes as needed. The switch to modularized zf1 packages will also make replacing zf1 piece by piece easier is we decide to go down that route.

This will allow us to run LibreTime on Ubuntu Bionic as well as the upcoming CentOS 8 which both bump PHP to 7.2. 

As mentioned in https://github.com/LibreTime/libretime/issues/580#issuecomment-450513733 I've tested this on Ubuntu 18.04 (Bionic Beaver) as well as Ubuntu 14.04 (Trusty Thar). I checked that the program grid, uploading tracks, basic playout and, podcast downloading works.

This is a rather large change and as such I'd like to ask everyone to do some serious testing with this change to ensure that all or your use cases still work. If you find any issues, please post the output of `ph p--version` as well as the full error message and stacktrace from either Apaches error.log or the zendphp.log and I'll fix them in this PR. To be sure we don't break any existing use cases there also needs to be some testing on the old still supported php 5.x versions. 

Fixes #580


hairmare:
I had to WIP this since it seems that I missed some of my local changes when upstreaming this (as evidenced by the failing tests). I'll de-WIP it once I ensure that I don't have any local changes left that need upstreaming.

hairmare:
Depends on https://github.com/zf1s/zend-loader/pull/1

frecuencialibre:
modern distro support is going to be huge. :+1: :+1: :+1: 

> This is a rather large change and as such I'd like to ask everyone to do some serious testing with this change to ensure that all or your use cases still work.

does a list of use cases to test exist anywhere?

paddatrapper:
Do we have a dev setup guide lying around somewhere? Preferably for someone who has never used Vagrant before...

Robbt:
No guide in particular but `sudo apt install vagrant` and then pull down this PR and `vagrant up ubuntu-bionic` and I think it'll work. I've ran into times when I needed to install a newer version of vagrant than my distro supported out of the box but it's pretty straight forward. Writing some docs would be a good idea as well.

paddatrapper:
Thanks. I'll try that and add some docs to the wiki

Robbt:
So I just tried this via vagrant up ubuntu-bionic and I got the Rest error with z1fs
`2019-01-03T20:50:47+00:00 ERR (3): localhost [Logging.php:166 - loggingShutdownCallback()] - Declaration of Zend_Rest_Route::assemble($data = Array, $reset = false, $encode = true) must be compatible with Zend_Controller_Router_Route_Module::assemble($data = Array, $reset = false, $encode = true, $partial = false) at /vagrant/vendor/zf1s/zend-rest/library/Zend/Rest/Route.php:52
array (
  'type' => 64,
  'message' => 'Declaration of Zend_Rest_Route::assemble($data = Array, $reset = false, $encode = true) must be compatible with Zend_Controller_Router_Route_Module::assemble($data = Array, $reset = false, $encode = true, $partial = false)',
  'file' => '/vagrant/vendor/zf1s/zend-rest/library/Zend/Rest/Route.php',
`

I think that is related to the change you merged with zf1s but for some reason it isn't applied upon installation. I'll look in further. Using PHP 7.2.10-0ubuntu0.18.04.1 (cli) (built: Sep 13 2018 13:45:02) ( NTS )

Robbt:
Yeah so far the UI seems pretty functional after the update. I've been able to add podcasts and create smartblocks.

hairmare:
@paddatrapper There are some docs on the vagrant setup at http://libretime.org/vagrant/ maybe we should merge those with your efforts.

paddatrapper:
@hairmare I will create a MR with any changes I have to that instruction list then

hairmare:
Looking at your wiki page adding something about the `--provider` arg might make sense. I've been wanting to start switching to base boxes that also support `--provider=libvirt` for a while now but never got around to pushing the change.

paddatrapper:
@hairmare - tracking it in #674 so we don't pollute this PR.

On that note - I haven't found any issues with PHP 7.2 testing this PR after applying the same fix @Robbt applied.

hairmare:
I just pushed the `composer update` fix. I had to run `composer update 'zf1s/*'` to ensure that phpunit wasn't bumped to a version that breaks travis.

frecuencialibre:
i just checked out `radiorabe:dev/zf1s` and ran `vagrant up` with no args from the working directory. setup failed with this error: 

```
    ubuntu-bionic: Do not run Composer as root/super user! See https://getcomposer.org/root fordetails
    ubuntu-bionic: Loading composer repositories with package information
    ubuntu-bionic: Installing dependencies from lock file
    ubuntu-bionic: Package operations: 31 installs, 0 updates, 1 removal
    ubuntu-bionic:   - Removing zendframework/zendframework1 (1.12.20)
    ubuntu-bionic:
    ubuntu-bionic:   [RuntimeException]
    ubuntu-bionic:   Could not delete /vagrant/vendor/zendframework/zendframework1/library/Zend/OpenId/Consumer/Storage/File.php:
    ubuntu-bionic:
    ubuntu-bionic:
    ubuntu-bionic: install [--prefer-source] [--prefer-dist] [--dry-run] [--dev] [--no-dev] [--no-custom-installers] [--no-autoloader] [--no-scripts] [--no-progress] [--no-suggest] [-v|vv|vvv|--verbose] [-o|--optimize-autoloader] [-a|--classmap-authoritative] [--apcu-autoloader] [--ignore-platform-reqs] [--] [<packages>]...
The SSH command responded with a non-zero exit status. Vagrant
assumes that this means the command failed. The output for this command
should be in the log above. Please read the output to determine what
went wrong.
```

paddatrapper:
I get that error on other branches too... Seems to be a vagrant thing more than related to this PR

paddatrapper:
@frecuencialibre if you `rm -r vendor` in the git repo this PR should work for you because it then won't pull `zendframework1` down

frecuencialibre:
thanks @paddatrapper your fix made composer complete successfully.

unfortunately i've now hit another error: 

```
    ubuntu-bionic:  * Installing pypo and liquidsoap...
    ubuntu-bionic: python /vagrant/python_apps/pypo/setup.py install --install-scripts=/usr/bin --no-init-script
    ubuntu-bionic: /vagrant/python_apps/pypo
    ubuntu-bionic: running install
    ubuntu-bionic: running bdist_egg
    ubuntu-bionic: running egg_info
    ubuntu-bionic: error: [Errno 13] Permission denied: 'airtime_playout.egg-info'
```

[python permissions error](https://github.com/googlesamples/assistant-sdk-python/issues/236)? sorry to be blindly posting errors, very new to python, vagrant etc.

frecuencialibre:
update: i got past this one with 
```
sudo chown -R myuser:mygroup python_apps
sudo vagrant provision
```
had some files in there that were owned by root. not sure why. now continuing troubleshoot:

```
    ubuntu-bionic: -----------------------------------------------------
    ubuntu-bionic:                 * Basic Setup DONE! *                
    ubuntu-bionic:                                                      
    ubuntu-bionic:  To get started with Airtime, visit             
    ubuntu-bionic:  or, if you've set up your own web configuration,    
    ubuntu-bionic:  the Airtime webroot on your webserver               
    ubuntu-bionic: -----------------------------------------------------
==> ubuntu-bionic: Running provisioner: install-mkdocs (shell)...
    ubuntu-bionic: Running: /tmp/vagrant-shell20190110-14886-1o0md0b.sh
    ubuntu-bionic: Updating Apt.
    ubuntu-bionic: Ensuring Pip is installed.
    ubuntu-bionic: Updating Pip.
    ubuntu-bionic: Ensuring Mkdocs is installed.
    ubuntu-bionic: Traceback (most recent call last):
    ubuntu-bionic:   File "/usr/bin/pip", line 9, in <module>
    ubuntu-bionic:     
    ubuntu-bionic: from pip import main
    ubuntu-bionic: ImportError
    ubuntu-bionic: : 
    ubuntu-bionic: cannot import name main
```

paddatrapper:
LibreTime should be up and running now. I also can't get the docs build to work on anything, but it isn't required to test this

frecuencialibre:
oh, sure enough!! 

http://localhost:8080


```
$ vagrant ssh ubuntu-bionic
vagrant@vagrant:/var/log$ php --version
PHP 7.2.10-0ubuntu0.18.04.1 (cli) (built: Sep 13 2018 13:45:02) ( NTS )
Copyright (c) 1997-2018 The PHP Group
Zend Engine v3.2.0, Copyright (c) 1998-2018 Zend Technologies
    with Zend OPcache v7.2.10-0ubuntu0.18.04.1, Copyright (c) 1999-2018, by Zend Technologies
```
thanks! checking use cases now

frecuencialibre:
works great!!!

i checked: track import, playlist creation/editing, webstream creation, podcast creation/edit, calendar manipulation, master source connection, stream settigns changes, user creation/edit, user assign to show, login/logout, user settings edit, set autoloading playlist.

i obviously haven't clicked every single thing, but the only functionality areas i haven't tested are those that are scheduled, like podcast download, and autoloading playlists.

frecuencialibre:
update: an autoloading playlist i scheduled worked, as did automatic download of a podcast. i think we're ready to go!

install
```diff
@@ -602,6 +602,8 @@ case "${dist}-${code}" in
         code="bionic"
         is_ubuntu_dist=true
         is_ubuntu_bionic=true
+        # dirty hack for bionic since postgresql has the wrong locale otherwise
+        localectl set-locale LANG=en_US.utf8
```
frecuencialibre: unfortunately this self described dirty hack is blocking my attempts to build a single container docker image. initially the error message is that `localectl` can't be found, but even if i install systemd onto the minimal docker bionic image, i still get a `Failed to create bus connection: No such file or directory` error that i think gets deep into the nature of the docker beast. what does postgresql do without this fix? how can i check whether alternative dirty hacks solve the problem?

i did notice these lines in the libretime-core dockerfile within David's multicontainer build:

        locale-gen "en_US.UTF-8"
        echo -e "LC_ALL=en_US.UTF-8\nLANG=en_US.UTF-8" >> /etc/default/locale
        [...]
        ENV LC_ALL=en_US.UTF-8
        ENV LANG=en_US.UTF-8
        ENV LANGUAGE=en_US.UTF-8

hairmare: Without the hack PostgreSQL doesn't get configured with the proper locale and the setup routine later fails when it tries to create the database. IMO this is part of how the Debian PostgreSQL package is trying to do an opinionated
config of PostgreSQL.

Personally I don't think the the monolithic docker image approach is how docker is intended to be used. Possibly alternative containerizations technologies might fit the bill much nicer if you really need to run systemd and all it's bits and pieces inside a container.

Could this be solved by using packages inside of the image? IMO the whole install script has always been a dirty hack written by legacy upstream to help some users test the latest code. As far as I can tell it has done about the same amount of harm as it has done good (ie. @radiorabe it killed one of our developers Linux install because he didn't run it in a VM when we just started out with legacy upstream).

paddatrapper: Can `locale-gen` not be used instead and avoid the dependency on systemd (and the message bus)? That way container solutions that don't provide a message bus, such as Docker (I think that is the issue), can still be support

hairmare: I'll give it a try... Would `locale-gen` alone work or does it also need to touch `/etc/default/locale`?

hairmare: I'm also not sure how the locale hack would behave in a non Vagrant context given that we have [this](https://github.com/LibreTime/libretime/releases#issues-no-locale-3.0.0-alpha.6) in the release notes so any suggestions to fix it are very welcome.

I isolated it in the bionic switch to make sure that old support stuff doesn't break for this reason.

Could it make sense to address the locale stuff in another PR so we can get PHP 7.2 out of the door? I'd probably also like to give `locale-gen "C"` a try for that matter.

frecuencialibre: reporting back here until another locale fix PR is created. simply replacing 

         localectl set-locale LANG=en_US.utf8

with 

        locale-gen "en_US.UTF-8"
        echo -e "LC_ALL=en_US.UTF-8\nLANG=en_US.UTF-8" >> /etc/default/locale

allows the install process to continue, but does not solve the postgresql issue. errors in the `install` report log:

```
Setting up postgresql-10 (10.6-0ubuntu0.18.04.1) ...
/var/lib/dpkg/info/postgresql-10.postinst: 2: /etc/default/locale: -e: not found
dpkg: error processing package postgresql-10 (--configure):
 installed postgresql-10 package post-installation script subprocess returned error exit status 127

[...]

dpkg: dependency problems prevent configuration of postgresql:
 postgresql depends on postgresql-10; however:
  Package postgresql-10 is not configured yet.

dpkg: error processing package postgresql (--configure):
 dependency problems - leaving unconfigured

[...]

Errors were encountered while processing:
 postgresql-10
 postgresql
E: Sub-process /usr/bin/dpkg returned an error code (1)

[...]

-----------------------------------------------------
              * Configuring PostgreSQL *              
-----------------------------------------------------
System has not been booted with systemd as init system (PID 1). Can't operate.
```

hairmare: > on the contrary, my understanding is that it's best to try to keep layers small.

See [here](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#minimize-the-number-of-layers) and [here](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#run) for best practices pertaining to layer sizing in docker. There ist also some info on [decoupling applications](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#decouple-applications). I understand this last link as a recommendation to not use an init system in a container unless you can't find a way around it:

> Limiting each container to one process is a good rule of thumb, but it is not a hard and fast rule.

In LibreTime's case there is nothing that really depends on something taking over PID 1 duties except for the installer. That is unless you want to do a bloated "full" LibreTime container which I personally wouldn't recommend and thus can't help you out with.

```
-----------------------------------------------------
              * Configuring PostgreSQL *              
-----------------------------------------------------
System has not been booted with systemd as init system (PID 1). Can't operate.
```

You seem to run into this during `docker build`. Since docker governs the build process it doesn't run it with systemd as PID 1 and thus the automated postgresql install can't work. You can get around this by calling the installer with `--no-postgres` so it skips trying to install and configure PostgreSQL but then you would need to take care of PostgreSQL beforehand. 

There are some examples on how to install PostgreSQL in the official library/postgres images at https://github.com/docker-library/postgres but they all assume that the PostgreSQL processes get to run in their own containers.

frecuencialibre: thanks for all your help. i definitely have no objection to a multi-container setup. in fact that's what we're using at my station. my goals with the docker build are simplicity/best practices, and also staying as close to upstream as possible, ie. avoiding fixes like https://github.com/ned-kelly/docker-multicontainer-libretime/issues/13

i'm going to continue this conversation in other threads.

Robbt:
What do we want to do before we push this ?

paddatrapper:
I don't think there is anything else required to merge this. Are there any doc related things that need updating to reflect new php support? The support matrix in the wiki at least will need changes to Debian Buster and Ubuntu Bionic

hairmare:
I'd like to do the 3.0.0-alpha.7 release pretty soon after we merge this as it will expose it to a larger audience. I think the [milestone](https://github.com/LibreTime/libretime/milestone/8) covers what's still open for that to happen.

When updating in-situ with a vagrant install from master you might need to remove the vendor for everything to work out. Releases shouldn't be affected by this issue.

If we want to I can update this to use
```
    locale-gen "en_US.UTF-8"
    echo -e "LC_ALL=en_US.UTF-8\nLANG=en_US.UTF-8" >> /etc/default/locale
```
as the dirty hack for ensuring a proper locale on Ubuntu Bionic. I'm assuming we would want to push some fixes for the locale issue before we announce official Bionic and Buster support. Maybe we can figure out how to get rid of the hack completely. Checking Buster so see if it's also affected by the PostgreSQL issue/hack could also be of interest.

paddatrapper:
I'm happy to test the Buster postgreSQL locale hack on Monday.

The rest of the milestone is pretty much dependant on this PR though. I can't really test the install without the new zf1s dependencies fixing the Vagrant build

paddatrapper:
It seems that Buster install works fine without the hack, but locales are generated in the installer later on [here](https://github.com/LibreTime/libretime/blob/master/install#L1110)

Running this on Buster using PHP 7.3, I get the following error:

```
compact(): Undefined variable: extras#0 [internal function]: exception_error_handler(8, 'compact(): Unde...', '/vagrant/vendor...', 404, Array) #1 /vagrant/vendor/zf1s/zend-view/library/Zend/View/Helper/HeadLink.php(404): compact('rel', 'type', 'href', 'media', 'conditionalStyl...', 'extras') #2 /vagrant/vendor/zf1s/zend-view/library/Zend/View/Helper/HeadLink.php(172): Zend_View_Helper_HeadLink->createDataStylesheet(Array) #3 /vagrant/airtime_mvc/application/controllers/ErrorController.php(68): Zend_View_Helper_HeadLink->__call('appendStyleshee...', Array) #4 /vagrant/airtime_mvc/application/controllers/ErrorController.php(9): ErrorController->setupCSS() #5 /vagrant/vendor/zf1s/zend-controller/library/Zend/Controller/Action.php(133): ErrorController->init() #6 /vagrant/vendor/zf1s/zend-controller/library/Zend/Controller/Dispatcher/Standard.php(281): Zend_Controller_Action->__construct(Object(Zend_Controller_Request_Http), Object(Zend_Controller_Response_Http), Array) #7 /vagrant/vendor/zf1s/zend-controller/library/Zend/Controller/Front.php(954): Zend_Controller_Dispatcher_Standard->dispatch(Object(Zend_Controller_Request_Http), Object(Zend_Controller_Response_Http)) #8 /vagrant/vendor/zf1s/zend-application/library/Zend/Application/Bootstrap/Bootstrap.php(105): Zend_Controller_Front->dispatch() #9 /vagrant/vendor/zf1s/zend-application/library/Zend/Application.php(384): Zend_Application_Bootstrap_Bootstrap->run() #10 /vagrant/airtime_mvc/application/airtime-boot.php(84): Zend_Application->run() #11 /vagrant/airtime_mvc/public/index.php(68): require_once('/vagrant/airtim...') #12 {main}

#0 /vagrant/vendor/zf1s/zend-controller/library/Zend/Controller/Front.php(965): Zend_Controller_Plugin_Broker->postDispatch(Object(Zend_Controller_Request_Http))
#1 /vagrant/vendor/zf1s/zend-application/library/Zend/Application/Bootstrap/Bootstrap.php(105): Zend_Controller_Front->dispatch()
#2 /vagrant/vendor/zf1s/zend-application/library/Zend/Application.php(384): Zend_Application_Bootstrap_Bootstrap->run()
#3 /vagrant/airtime_mvc/application/airtime-boot.php(84): Zend_Application->run()
#4 /vagrant/airtime_mvc/public/index.php(68): require_once('/vagrant/airtim...')
#5 {main}
```

hairmare:
PHP has this in their [docs](http://php.net/manual/en/function.compact.php)

> Before PHP 7.3, any strings that are not set will silently be skipped.

So it looks like we'll need to fix some more stuff before we're ready for 7.3...

Most likely initializing `$extras` to `array()` before the following if should be enough...

https://github.com/zf1s/zend-view/blob/61363f26f90ac3b92a11b40d5085130b878a3000/library/Zend/View/Helper/HeadLink.php#L399-L404

I'll look into taking care of this later this week.

hairmare:
I rebased this onto master, removed the locale hack and, force pushed it.

My rationale for removing the locale hack is that we need to look into it and it shouldn't be introduced as part of this pr.

This only supports up to php 7.2 for now but I think we should merge it soonish rather than wait for php 7.3 support. 7.3 is in the pipeline and I'll make more update PRs when the needed fixes are here.

Let me know if I should update anything else so we can get this merged.

Robbt:
Sounds good to me. I'm willing to go ahead with the merge.

install
hairmare: Resolving this since I removed the locale hack.

frecuencialibre:
i already tested functionality following this change as systematically as i reasonably could and it worked great. i won't be able to test the most recent vagrant changes until later this week, and know Robbt is deep in another problem today. @paddatrapper to the rescue? happy rainy sunday folks, take care

PR #683, opened 2019-01-10
Add support for Buster in the installer

paddatrapper:
Fixes #682

PR #685, opened 2019-01-11
Installer - Add warning for the deprecation of Debian Jessie

### LINKED ISSUE #684: Debian Jessie is EOL
Debian Jessie ended security updates on 17/05/2018, but a limited number of packages and architectures are supported until around 2020 by the LTS project.

We should check if the packages required by LT are supported by the LTS project. If they aren't, we should drop support for Jessie.

paddatrapper:
Jessie is out of security support and will be dropped inLibreTime 3.0.0-alpha.8

Fixes #684

### DIFF
--- a/install
+++ b/install
@@ -633,7 +633,7 @@ case "${dist}-${code}" in
         code="jessie"
         is_debian_dist=true
         echo -e "ERROR: Debian Jessie is archived and does not receive any security or other updates since 2018-05-17." >&2
-        echo -e "The LibreTime installer dropped support for installing LibreTime on Jessie in 3.0.0-alpha.8." >&2
+        echo -e "The LibreTime installer will drop support for installing LibreTime on Jessie in 3.0.0-alpha.8." >&2
         sleep 6
         is_debian_jessie=true
         ;;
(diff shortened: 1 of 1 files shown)


install
```diff
@@ -632,6 +632,9 @@ case "${dist}-${code}" in
     debian-8|debian-jessie)
         code="jessie"
         is_debian_dist=true
+        echo -e "ERROR: Debian Jessie is archived and does not receive any security or other updates since 2018-05-17." >&2
+        echo -e "The LibreTime installer dropped support for installing LibreTime on Jessie in 3.0.0-alpha.8." >&2
```
hairmare:
```suggestion
        echo -e "The LibreTime installer dropped support for installing LibreTime on Jessie in 3.0.0-alpha.7." >&2
```
I'd assume that this would make it into alpha.7... Not that it's that important.

LGTM

paddatrapper: My thought was warn for one release and then deprecate. So release alpha.7 simply warning about the change and then drop full support in alpha.8. I do need to fix the grammar though

hairmare:
I'll try to remember to update `/docs/vagrant.md` once #683 is merged. I think it's would be ok to already remove Jessie from the docs even though the installer only warns for now.

hairmare:
Allright, then we'll update the Jessie docs one we really drop support.

PR #678, opened 2019-01-07
Vagrant - make libvirt provider available

paddatrapper:
Use generic roboxes for Vagrant bases. This fixes #674. It also updates the [docs](http://libretime.org/vagrant/) to reflect the options available.

~Currently rabbitMQ fails to start in Ubuntu 16.04, but I am still testing if this is as a result of these changes or not.~ (It fails to start on master too)

PR #700, opened 2019-01-18
Vagrant - make libvirt provider available

### LINKED ISSUE #674: Problem: Vagrant base boxes lacks libvirt support for ubuntu xenial
The Bento base Ubuntu Xenial box does not support libvirt. This causes issues with Debian vagrant boxes, as this is the default provider. This could also affect the other base boxes, but I still need to check...

Documentation of how to use Vagrant for LT also needs updating:

http://libretime.org/vagrant/

https://github.com/LibreTime/libretime/wiki/Vagrant-Developer-Setup (to be merged into the website docs)

paddatrapper:
Replaces #678 

Fixes #674

Need to test the following hosts (using any/all vagrant boxes):

- [ ] Debian Stretch
- [x] Debian Buster/Sid
- [x] Ubuntu 16.04
- [ ] Ubuntu 18.04
- [ ] Fedora

Edit: wow! 700 Issues and PRs so far :tada:


paddatrapper:
@Robbt please check if the instructions in the docs reflect what you experienced on 16.04

Robbt:
Additional things I needed to do to get it working under Ubuntu 18.10 and I'm assuming will apply to other Ubuntu hosts is - also install the `sudo apt-get -y install libvirt-dev` 
I also needed to do `sudo usermod -a -G libvirt USERNAME` and then reboot my system because it uses the sock authentication by default based upon membership in libvirt.

paddatrapper:
logging out and back in *should* work because groups are applied on login. I'll add a note that it may require a reboot

Robbt:
I'd be happy to merge this now because it is working for our scenarios (latest Debian & Ubuntu) and if it doesn't work on other distributions the pre-existing virtualbox will continue to work for people since we got it working without changing the images.

If people run into issues with getting libvirt working then we can update the docs accordingly with the steps required to make it work on their particular setup, but since this is primarily going to be used by developers then I don't think it's critical that we do testing on all of the potential setups.

paddatrapper:
Sounds good :+1:

frecuencialibre:
i'm unfortunately still having problems getting my debian 9 laptop to recognize libvirt.

from the docs here http://libretime.org/vagrant/ i've followed almost all the instructions:

```
sudo apt install vagrant vagrant-libvirt libvirt-daemon-system vagrant-mutate libvirt-dev
sudo usermod -a -G libvirt $USER

# Reboot
```

but am still getting:

```
r@rdebian /var/www/html/libretime (buster-install *=) $ vagrant up debian-buster --provider=libvirt
The provider 'libvirt' could not be found, but was requested to
back the machine 'debian-buster'. Please use a provider that exists.
```

my user _is_ in the libvirt group: 

```
r@rdebian /var/www/html/libretime (buster-install *=) $ id
uid=1000(r) gid=1000(r) groups=1000(r),24(cdrom),25(floppy),27(sudo),29(audio),30(dip),44(video),46(plugdev),108(netdev),114(bluetooth),118(scanner),128(libvirt),999(docker)
```

so i'm guessing this is still happening since i haven't run the buster equivalents of these commands: 

```
vagrant box fetch bento/ubuntu-16.04
vagrant mutate bento/ubuntu-16.04 libvirt
```

would those be?

paddatrapper:
Those just convert the box to support libvirt. Your issue seems to be before that. I think it is an issue with the version of vagrant et. al. provided in stretch. I would suggest using virtualbox unfortunately

frecuencialibre:
:cry: should we add this to the docs? or is my machine just weird?

paddatrapper:
I would like to play around on a stretch system and see before ruling it out. The issue is I need to format a machine to do so because testing virtualisation under a virtual machine doesn't work nicely...

frecuencialibre:
seems like other stuff might be higher priority as we're getting ready for
buster...

Robbt:
yeah i've had some issues with libvirt although I did get it working for ubuntu-xenial, the other distros not so much, and mixing virtualbox and libvirt can result in some conflicts around networking that I'm not sure how to resolve without rebooting.

paddatrapper:
Network conflicts can usually be solved by stopping VirtualBox daemon, destroying the box and recreating it under libvirt. But the two providers definitely don't play well together

paddatrapper:
@frecuencialibre I did some testing on a stretch install and realised some of the instructions were wrong. See the changes made in #712

PR #704, opened 2019-01-20
Fix smartblock multi criteria

### LINKED ISSUE #458: bug when using mutiple criteria in smart blocks
When I define multiple criteria in a smart block the operand 'and' automatically appears.
That's fine, but when I edit the smart blocks, the operand 'and' changes into 'or'.
I tested the outcome with the preview-option and the subset of records is really changed.
So, it's not only a cosmetic issue but the sql-command is really different from the one I created.
This results in many more hits due to the 'or'-operand instead of the more restrictive 'and'-operand.

Robbt:
This fixes #458 
At this point I have it mostly working upon creation but need to modify the parsing of the database when smartBlockCriteria reloads the form with $startForm.

It may not be perfect either and could use testing, currently only testable when a smartblock is first created with the preview until I finish up the refactor of the form mentioned above.


Robbt:
startForm is turning out to be annoying to refactor because of the hack to add criteriaGroup and the fact that the original designers only lumped everything by criteria - so now I have to figure out how to add another loop inside of 2 other loops without breaking functionality for normal criteria. 
Part of me thinks it might make more sense to work on dumping all of the hack jquery all together and rewrite it with something more modern like vue.js because the way it is currently setup is a nightmare of disabled elements with jquery and I keep finding new jquery bugs. But I'm not there yet because I'm still confident that I can perform some clever hack to keep things working without rewriting and I'm not sure how long it will take to completely refactor the form with vue.js as that is a whole new process.

Robbt:
So after wrestling with this for quite a few hours attempting to refactor things I realized it could be done pretty easily via creating separate arrays with slightly different names by appending the criteriagroup to the criteriakey thus separating them while the actual values of everything are still the same. I did this after I wrote the other function that matches criteria so it might benefit from some refactoring. 

But as best I can tell this is actually fully working and ready for testing. Before we merge it I think I might want to clean up the jquery code in smartblock.js because it has some bugs related to multiple criteria of the same genre and modifiers. It makes assumptions that if something has the same criteria  it should be grouped together underneath. 

So it could use some cleanup but its kind of a nightmare and I am pretty sure no new bugs were introduced by the work I did, I was just able to reveal some new bugs while playing with the UI so intensely.

frecuencialibre:
thanks @Robbt for all this work!

1. this now works:
screenshot from 2019-01-23 07-52-01

2. unfortunately this does not:
screenshot from 2019-01-23 08-04-13

3. and then as you noted, clicking the red x to delete "does not contain hip" causes the UI to mistakenly do this:
screenshot from 2019-01-23 08-04-29

4. however, this works and accomplishes the same as # 2 above:
screenshot from 2019-01-23 08-06-47

5. i'm seeing a "Smart block generated and criteria saved" green success message upon generating a preview above the search criteria. this is new right? i don't think this fits with "previewing" your WIP smartblock.

6. unrelated? tracks with a metatag empty (eg. a completely blank Genre) appear not to be returned by the "does not contain" modifier.

i can have a look at number 3 above if you'd like. tho i hate to separate you from your beloved "jquery spaghetti code" ;P

and then a question, not sure if i'm understanding correctly, but i think you said you ended up abandoning an initial approach to resolving this? are the code changes here all needed for the final solution you chose? (i'm noticing 14 files changed, and a few logging lines sprinkled in i think as well.)

de nuevo muchas gracias desde nuestra pequeña estación!

Robbt:
Ok thanks for testing.
The issue with number 2 above is that you are basically saying this track does not contain this or does not contain that. So due to the logic involved with the way OR works in SQL this will run into the same problem we originally experienced with it combining them all of the ands into ORs and returning the results of all tracks. I could break out a logic venn diagram but basically if something is not a turtle OR something is not a zebra will return everything because a turtle is not a zebra and a zebra is not turtle. Does that make sense ?

Also I think the logic still matches with does not contain with null entries if this is the only the modifier otherwise it removes that line when combining them because I couldn't figure out the best way of handling this and I figured that if it was matching something in the same criteria you wouldn't want to match on blank. For instance if something does not contain "purple" and does contain "green" you wouldn't want to match blanks. So I figured the only time the null would be relevant is if it was only "does not contain" x. Now I can see that if you have multiple "does not contain"  X and "does not contain" Y then you would also want the null results but otherwise I think my assumption was right.

Preview has always saved the smart blocks because it is just a label change from "Generate" similar to how generating the smartblock criteria under the podcasts will automatically save the form. We could change the language if this is confusing to people but I think it'd be a fair amount of work to create a temporary smartblock that is tested and then discarded.

How do we want to deal with the challenge of "does not contain" OR modifiers and the confusion this could cause. We could flash a warning and disallow it or we could try to reinterpret it as we imagine it they wanted it.

As far as the jquery stuff if you want to look into that it'd be welcome. We also need to figure out why after adding and removing modifiers the date time criteria don't work and this is a jquery issue.

So I think it might make a lot of sense for us to build some tests so that we can be assured this is working with whatever changes so we make see #269 for my initial attempt at this.

frecuencialibre:
Regarding number 2, thanks for the explanation. i now think that it's ok to expect people to do it as separate criteria like in 4.

regarding null values, understood. i now agree. this is a non-issue for our station anyway, as this functionality will be used to filter strings of multiple words. ie. contains "turtle" but not "zebra" or "horse". null values won't contain turtle.

> Preview has always saved the smart blocks

ah, ok. should we maybe remove that green success message? or is it not sufficiently confusing?

> How do we want to deal with the challenge of "does not contain" OR modifiers and the confusion this could cause. 

i'd say we shouldn't worry. your explanation is logical, i was mistaken. the code is already doing what the UI says it should.

i have a little time today to look at this js, (and maybe some css so the rows don't break), and so will push to this branch.

is this PR dependent upon #269? unfortunately that one's out of my league...

Robbt:
It's not dependent upon #269 but I think that adding tests is a good idea. It's actually fairly straight forward to write the tests. I already wrote a number of them I just didn't feel like it was sufficient to be considered worth pushing. It's mainly labor intensive at this point because you have to define each of the tracks in JSON and the smartblock form in JSON. I could probably work some kind of extract from postgres to json to speed up the process but I didn't do it. I might do it just to make sure this is working as described.

Robbt:
Ok, another thing I just realized is that there is no way of setting up an OR statement linking options of different criteria. 

For instance if I wanted to have the newest track downloaded by a specific podcast OR uploaded by a specific user there is no way of currently doing that.

Maybe we should just get rid of the confusing "New Modifier" and "New Criteria" and replace them with "Add Or Critera" and "Add And Criteria" and make it so that by default the "Modifer" under "Add Or Criteria" is the same but it can be changed via the drop down.

frecuencialibre:
> Maybe we should just get rid of the confusing "New Modifier" and "New Criteria" and replace them with "Add Or Critera" and "Add And Criteria" and make it so that by default the "Modifer" under "Add Or Criteria" is the same but it can be changed via the drop down.

ok i think it makes sense to do this since fixing the jquery bug mentioned above means also touching this code. 

i'm thinking maybe we should abandon ability to add multiple OR criteria "within" a criteria (as seen in screenshots 2 and 3 above). you'd just have to select which metadata field you want to target every time you add a new criteria, as seen in screenshot 4. do you agree with this approach?

frecuencialibre:
so nothing fancy here, still thinking about how to accurately visually communicate what a search query that mixes OR and AND

screenshot from 2019-02-02 09-44-03

frecuencialibre:
please disregard my last comment. changing ideas...

i think i'm seeing an issue when adding an OR modifier on a single metatag, and then hitting preview.

in the preview i'm seeing that a correct number of "tracks match your search criteria. ", but then only tracks matched by the first criteria are shown in the preview.

more soon...

Robbt:
Interesting, yeah there is a possibility that there are some bugs in my code and that I need to revisit it. I think that we should work out the UI changes, and then I plan on closing my test PR #269 and moving those tests into here so that we can confirm everything is working as it should.

If you find bugs with the preview it is possible that I need to diagnose the code further and implement the same code. I have had less LibreTime recently so I'm not sure when I'll be able to get to it. At least I got my stations ancient drupal site back up and running so I can focus my energy elsewhere.

Robbt:
Hmm, that looks good. It looks like you decided against making the criteria visible for OR fields. So that should probably be a separate PR anyways. I'll do some testing of the code change you made. Like I said this is some scary jquery nightmare code to mess with, I already dealt with it once to get the datetime stuff working and I'd prefer to never have to deal with it again but the only alternative I see is a rewrite/replacement and as my knowledge of the code is incomplete this will take an unknown amount of time to do.

I'll try to confirm the preview bug and move my smartblock testing PR over here.

Robbt:
Unfortunately the changes you made broke the brittle mods I made previously to enable the extra datetime stuff.
See this screenshot - I don't think we can do static positioning without refactoring the way we handle datetime and extra datetime stuff.
extradatetime-smartblock-ui

Maybe we should just try to avoid refactoring the UI in this PR and see if we can get it working with the clunky previous UI. Especially if the bugs in the UI aren't easily fixable without significant refactor. It's probably best to avoid introducing new ones if we want to merge this PR.

Robbt:
I just tested the preview bug you mentioned and adding a modifier and then hitting preview showed the correct track listing and added the additional tracks to the list as well so I couldn't reproduce it. If you want to revert the UI modification PR that is the simple way of resolving these issues otherwise we might just be digging ourselves into a deeper hole.

airtime_mvc/public/css/styles.css
```diff
@@ -703,15 +703,29 @@ table.library-get-file-md.table-small{
...
+.db-logic-label-and {
+    top: 42px;
 }
 .sp-invisible{
     visibility: hidden;
 }
 
 .sp_input_select{
-    width: 140px;
+    flex: 0 0 33%;
```
Robbt: If we set this to flex: 0 0 10%; it works for the datetime extra selects but everything is super cramped.

frecuencialibre:
screenshot from 2019-02-04 09-58-34

here's the preview bug

airtime_mvc/public/css/styles.css
Robbt: I tried changing this to .sp_input_select{
    flex-shrink: 3;
}
and it seems to fix the issue but does create some possible issues with the 'or' and 'and' overlapping. If you want to try this it is a possible approach to fixing the problem.

frecuencialibre:
the preview bug was a problem on my machine where the audio files didn't actually exist. 

i now think this PR is ready for testing.

frecuencialibre:
ok so i'm seeing a problem with error checking that needs fixing before this is merged. try trying to save a smartblock with required fields unset. i'm getting a 500 error in the response from the POST to `playlist/save`

Robbt:
Ok, the previous bugs in jquery code not withstanding this does appear to work nicely and I like the final touches with the jquery UI so I'm going to merge this now.

PR #711, opened 2019-01-21
passed $show variable when smartblock added via playlist

### LINKED ISSUE #708: Time Remaining Smartblock Doesn't Work When Embedded In a Playlist
**Describe the bug**
Basically the Time Remaining Playlist doesn't pick up the $show and thus defaults to 60 minutes when added via a playlist. This interferes with the usage of Time Remaining Smartblocks with AutoLoading Playlists

**To Reproduce**
Add a Time Remaining Smartblock to a Playlist
Add that Playlist to a Show
It won't get the remaining time

To see it work directly add a time remaining smartblock to a show directly and it will fill the time 

**Expected behavior**
It should see calculate its length based upon the time remaining in the show the playlist is added to.

**LibreTime version:**
master as of right now

**Installation method:**
 - OS:Ubuntu
 - OS Version 16.04.5
  Method:  vagrant

**Additional context**
This has to do with the $show parameter not being created or passed via whatever mechanism adds playlists. So its probably going to need some more refactoring or analysis to get this working. I knew it was too good to be true when I implemented it in under 3 hours.

Robbt:
I'm pretty sure this fixes #708

### DIFF
--- a/airtime_mvc/application/models/Scheduler.php
+++ b/airtime_mvc/application/models/Scheduler.php
@@ -279,7 +279,7 @@ final class Application_Model_Scheduler
                     } else {
                         $defaultFadeIn = Application_Model_Preference::GetDefaultFadeIn();
                         $defaultFadeOut = Application_Model_Preference::GetDefaultFadeOut();
-                        $dynamicFiles = $bl->getListOfFilesUnderLimit();
+                        $dynamicFiles = $bl->getListOfFilesUnderLimit($show);
                         foreach ($dynamicFiles as $f) {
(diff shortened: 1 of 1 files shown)


hairmare:
Won't this break if you try to add a playlist containing a time Remaining smartblock to more than one show?

Robbt:
I don't think so. It's only passing the variable of the show instance that a playlist is being added to. I don't know how you could add a playlist to multiple shows in the same action.

hairmare:
Sounds about right 👍

PR #713, opened 2019-01-22
Move sql from #659 to alpha.7

Robbt:
Basically I jumped the gun on using alpha.8 for the SQL before alpha.7 was out. So it would probably make sense to move the SQL from #659 back to alpha.7 before release.

### DIFF
--- a/airtime_mvc/application/controllers/downgrade_sql/airtime_3.0.0-alpha.7/downgrade.sql
+++ b/airtime_mvc/application/controllers/downgrade_sql/airtime_3.0.0-alpha.7/downgrade.sql
@@ -1,1 +1,3 @@
 ALTER TABLE cc_block ALTER COLUMN type SET DEFAULT 'static';
+ALTER TABLE podcast_episodes DROP COLUMN IF EXISTS episode_title;
+ALTER TABLE podcast_episodes DROP COLUMN IF EXISTS episode_description;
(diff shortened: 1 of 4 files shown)


Robbt:
Ok, yeah I'm not sure how the versioning works in terms of these updates and when or how they are applied. Do we need to set a "version" somewhere to tell the system to apply the updates for that version ? Because if so we should do that so the .8 code is applied. There is another PR in progress that is using .8 as well and I'm not sure if it'll be merged before we release .7. 

Either way it'd be good to figure out the best practices for this so that we can avoid confusion in the future. I've tried to avoid database updates in the past but we are getting to the point where they are inevitable and as I do more of them they aren't as annoying as they were at first. I just want to make sure that people don't have to manually apply the changes etc. Feel free to close this,

hairmare:
The next PR that needs to do additional migrations will need to use alpha.9. The subdirs in the upgrade_sql dir get applied one after another ordered by version number.

Robbt:
So wouldn't it make sense to just move this back to .7 so we can reserve .8 for the next branch since the PR I mentioned just merged  ? Or are you suggesting that we just do a separate database increment for each PR independent of the specific alpha version ? I am down with that but just wanted to make sure we are on the same page as it could end up with the database versions being far ahead of the actual release branch.

Robbt:
Going to close this with the idea that we will just increment database changes on a per PR issue disconnected from the alpha version itself.

Issue #698, opened 2019-01-17
Settings menu item not visible for Program Manager users


delrizzo:
Hi Gang,
Really appreciate the work you're all doing here. We are in the midst of setting up a broadcast station, and have switched from 2.5.1 to Libretime. Noticing some things and I hope to get more involved as time goes on and our station becomes more self-reliant.

Anyhow, In creating our user database yesterday, I noticed that users designated at "Program Managers" do not have the "Settings" tab in the left side menu. This doesn't seem to be the case for "Admin" or even (at least) "DJ" level users. Because there is no username link at the top right as there was in 2.5.1, users need to access that Settings tab in order to be able to change their default assigned password. I think this is possibly a bug, so reporting here. I did a check of previous issues and found nothing that referred to this, so if it exists, my apologies.

**Describe the bug**
"Settings" menu tab not visible for "Program Manager" users.

**To Reproduce**
1. Create a new user
2. Give the user "Program Manager" access
3. Logout, and login again as that user
4. Notice that there is no "Settings" tab in the menu

**Expected behavior**
I expected there to be a Settings tab, which would give the user access to be able to modify their individual settings, or at the very least change their password.

**LibreTime version:**
0370dfb7c

**Installation method:**
 - OS: Ubuntu
 - OS Version 18.04.1 LTS
 - Method: ./install
 - Details: Installed with PHP 7.1
![screen shot 2019-01-17 at 1 00 58 am](https://user-images.githubusercontent.com/30699584/51306999-74de8c00-19f3-11e9-9383-301623997d96.png)

**Screenshots**

**Client (please complete the following information if applicable):**
 - OS: Mac 10.12.6
 - Browser Chrome
 - Version 71.0.3578.98

paddatrapper:
Hi @delrizzo. This is definitely a bug. Thanks for reporting it!

hairmare:
Does clicking their username work? That should navigate to `/user/edit-user` as well.

In the ACL bit the roles are a hierarchy so a program manager should always see more than a DJ.

That the Settings menu can only be seen by admins was introduced in 3aee9a49bc524450011310419dee1adb23b9a702 on the saas-dev branch.

delrizzo:
I have not been able to find the username to click on... maybe I'm missing something? I was wondering why that isn't there (vs being there in Libretime).
Further to that, a DJ can see the Settings tab, but a guest cannot. And there is no text link for user where it used to be in airtime.

Robbt:
[...]

PR #715, opened 2019-01-23
masterpanel ui tweaks

frecuencialibre:
this is an opinionated second half of a solution to https://github.com/LibreTime/libretime/issues/698

it does this:

screenshot from 2019-01-22 18-15-08

i also removed the timezone since IMO it isn't something that changes needs prominence.

i'd also like to actually hide the update indicator for non-admin users, but haven't since i wanted to hear whether folks think that's a good idea. i've had to explain to radio collective members about updates etc., and it's been more of a distraction than anything else. if this does sound like a good idea i could use some guidance regarding possibly modifying `VersionNotify.php`

### DIFF
--- a/airtime_mvc/application/layouts/scripts/layout.phtml
+++ b/airtime_mvc/application/layouts/scripts/layout.phtml
@@ -92,6 +92,14 @@ j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=
     <div id="nav">
         <?php echo $this->navigation()->menu(); ?>
     </div>
+        
+    <div class="btn-group">
+        <a href="<?php echo $this->baseUrl . '/login/logout'; ?>">
+            <button id="add_media_btn" class="btn btn-small dashboard-btn btn-danger">
+                <span><?php echo _("Logout") ?></span>
+            </button>
+        </a>
+    </div>
 
     <script>
     <?php
--- a/airtime_mvc/application/layouts/scripts/layout.phtml
+++ b/airtime_mvc/application/layouts/scripts/layout.phtml
@@ -92,14 +92,6 @@ j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=
     <div id="nav">
         <?php echo $this->navigation()->menu(); ?>
     </div>
-        
-    <div class="btn-group">
-        <a href="<?php echo $this->baseUrl . '/login/logout'; ?>">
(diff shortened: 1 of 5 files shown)


frecuencialibre:
i don't have a strong opinion regarding the red button. whatever works. i
removed it because 1. it's redundant if we have the login link in the user
block in the top right as per common website practice, and 2. the button's
bootstrap color and look didn't match the airtime black and orange theme

Robbt:
Yeah I'd prefer that we keep the red logout button at the bottom left and also add it to the top right. I think that removing the top right is a little confusing as I find myself trying to click there to logout before remembering but for people who are very new to the UI they could find the big red button easy to use and we do want to encourage people to logout (although I don't know what our cookie timeout session is set to).

frecuencialibre:
ok i re-instated the red button. no probz it was just an aesthetic thing. anything else?

Robbt:
Honestly from a UI perspective my only other concern/question is whether we should make the settings default to general vs. the profile for admin users and revert or change some of the work done in #705 but this is just a UI flow concern that I should of brought up before we merged the changes there.

frecuencialibre:
yeah i think this current PR only touches the header .phtml and some css. your comment above sounds like a new, separate issue.

Robbt:
Sounds good. Let's just merge this then and we can revisit that in the future if and when we update the Settings workflow.

xabispacebiker:
Hi mates,

I think we are taking very important steps too quickly without a great consensus. We are applying here custom interface layout preferences without considering other opinions or opening a debate.

I personally don't like the buttons on the right because it causes a lot of confusion to the users. After different installations in real cases, users complained that they could not find the way to change their name, password or simply log out from their account. After installing the button on the left side and following a logic (navigation on the left, settings under a tab), new users never complained again. 

I have a lot of experience and I have invested a lot on web design and user experience and from my point of view is much more logical the logout button on the bottom left and all related to the settings under a tab on that same column.

In addition, this gives us the option of freeing space in the header,  even if it is simply to have a clean clock, very useful when making live shows.

Regarding the logout button color, that red was taken from airtime.css. It is used in other buttons and parts of the template.

Robbt:
Well this step was taken about a year ago, but I agree that we should come up with a consistent UI. Another thing that would be great would be for a responsive UI that enables basic tasks to be performed from a small screen such as a smart phone because there have been times when I've had to pull my car over to resolve an issue via my phone and it is annoyingly difficult to say the least.

paddatrapper:
I think re-designing the UI is a good thing to do, but it will probably be best to do it with the new API - make it responsive and API-driven in one

PR #716, opened 2019-01-23
Clarify podcast settings and workflow

### LINKED ISSUE #690: Clarify podcast settings and workflow
Podcast track metadata override settings currently exist in two places: station-wide general settings, and on the edit screen of individual podcasts. 

Currently "The individual override is only used if the global setting is turned off." - https://github.com/LibreTime/libretime/pull/659#issuecomment-453847011 

I personally find this confusing since i think a user would have no way of expecting that behavior, but wanted to hear other opinions.

I'd like to propose that 
1. we actually remove the "Podcast Album Override" setting from station-wide general settings. 
2. the individual podcast metadata overwrite checkbox be enabled by default.
3. the individual podcast metadata overwrite checkbox do what it says on the tin. ;) 
possible label: "Overwrite episode metadata tags", and tooltip: "Overwrite downloaded podcast episodes' "Artist" metadata tag with the Podcast Name specified above. This tag can then be used as a search criteria by a smartblock. [More information](http://libretime.org/???)"  The docs could then cover that the album is also overwritten for backwards compatibility, and that the metadata will be overwritten if empty even if this checkbox is unchecked.

frecuencialibre:
- [x] remove the "Overwrite episode album names" checkbox from individual podcast settings
- [x] disable by default the "Podcast Automatic Smartblock and Playlist" setting in general settings.
- [x] fix changed strings in translation files

@Robbt regarding the links to the smartblock and playlist, opening separate panes via js isn't too complicated, but i need their numeric ids. any pointers on how to get the API to return those when `/rest/podcast/smartblock` is hit?

edit: no longer going to include links to generated smarblock / playlist

Closes #690


frecuencialibre:
i had included adding links to the newly generated smartblock and playlist in the todo list for this PR since that would tidily close #690, but actually what i need is an evaluation of whether adding those is worth some of your time @Robbt helping me to get the api response to include their ids. if we're not going to add those links then i'll go ahead and finish this PR without that feature since the UI is currently broken.

frecuencialibre:
Hi all, this is now finally ready for testing and review. Thanks!

Robbt:
Looks good. Going to merge.

Issue #166, opened 2017-04-16
Problem: Python magic does not match what major distros package


hairmare:
## Issue

When installing with some pre-release packages I chose to install python-magic as a package. Turns out there are multiple libraries called python-magic and the one from the [official file command](http://www.darwinsys.com/file/) is not the one that [analyzer is using](https://github.com/LibreTime/libretime/blob/54ded86f6de3233c53e1dcbcb4f81902e508bed1/python_apps/airtime_analyzer/setup.py#L34). As detailed on https://github.com/ahupp/python-magic/issues/33 they share the same namespace in python and no one seems willing to do anything about it.

Since this is rather confusing I made a table to illustrate it:

| pypi package | python namespace | distro package | distro version | pypi version |
| ---- | ---- | ---- | ---- | ---- |
| `python-magic` | `magic` | not available | not available | 0.4.13 |
| `file-magic` | `magic` | `python-magic` | >= 5.11 | 0.3.0 |

This becomes a problem when we start packaging LibreTime in distro native formats because we need to package an additional dependency. I think the pip installed magic version takes precedence over what a distro installs but the packages both having the same namespace doesn't help at all.

## Proposed Solution
Given that the python-magic package installs the *official* binding on both [Debian](https://packages.debian.org/unstable/python-magic) and [CentOS](https://centos.pkgs.org/7/centos-x86_64/python-magic-5.11-33.el7.noarch.rpm.html) it think it makes most sense to rewrite our code to use that.

The needed package is also on [pypi](https://pypi.python.org/pypi/file-magic/0.3.0) <strike>but I can't seem to work out why it's version is too low</strike> and it's versioning doesn't match the distro package version because it's a separate python api version instead of being in sync with the main package.

See #169 for an implementation of this solution

paddatrapper:
So this has become an issue again... Debian has moved to [python-magic](https://github.com/ahupp/python-magic) from Buster onwards, which means that Ubuntu 18.04 and 18.10 also [use it](https://launchpad.net/ubuntu/+source/python-magic)

PR #719, opened 2019-01-23
Use python-magic for mime type detection

### LINKED ISSUE #717: `file-magic` newest release
**Describe the bug**
With `file-magic` 0.4.0 which went out live on 1st of Jan 2019
https://pypi.org/project/file-magic/#history
there is an issue that the `airtime_analyzer` no longer works properly:
```
2019-01-23 05:28:42,251 [analyzer_pipeline] [ERROR]  'module' object has no attribute 'open'
  File "/usr/local/lib/python2.7/dist-packages/airtime_analyzer-0.1-py2.7.egg/airtime_analyzer/analyzer_pipeline.py", line 66, in run_analysis
    metadata = MetadataAnalyzer.analyze(audio_file_path, metadata)
  File "/usr/local/lib/python2.7/dist-packages/airtime_analyzer-0.1-py2.7.egg/airtime_analyzer/metadata_analyzer.py", line 44, in analyze
    ms = magic.open(magic.MIME_TYPE
```

**To Reproduce**
Install `file-magic == 0.4.0`.

**Expected behavior**
No error should happen or fix `file-magic == 0.3.0` in the `requirements.txt` file.

**LibreTime version:**
`8dd55e2`

paddatrapper:
Fixes #717

### DIFF
--- a/python_apps/airtime_analyzer/airtime_analyzer/metadata_analyzer.py
+++ b/python_apps/airtime_analyzer/airtime_analyzer/metadata_analyzer.py
@@ -40,12 +40,9 @@ class MetadataAnalyzer(Analyzer):
                 m.update(data)
             metadata["md5"] = m.hexdigest()
 
-        # Mutagen doesn't handle WAVE files so we use a different package 
-        ms = magic.open(magic.MIME_TYPE)
-        ms.load()
-        with open(filename, 'rb') as fh:
-            mime_check = ms.buffer(fh.read(2014))
+        mime_check = magic.from_file(filename, mime=True)
         metadata["mime"] = mime_check
+        # Mutagen calculates durations incorrectly for wave files
         if mime_check == 'audio/x-wav':
             return MetadataAnalyzer._analyze_wave(filename, metadata)
 
--- a/python_apps/airtime_analyzer/setup.py
+++ b/python_apps/airtime_analyzer/setup.py
@@ -31,7 +31,7 @@ setup(name='airtime_analyzer',
           'mutagen>=1.41.1', # got rid of specific version requirement 
           'pika',
           'daemon',
-          'file-magic',
+          'python-magic',
           'nose',
           'coverage',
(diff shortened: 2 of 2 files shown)


paddatrapper:
Ok it is still needed, but the separate handling of wave files can be dropped.

paddatrapper:
The python-magic package in Debian from Buster on has been moved to use [python-magic](https://github.com/ahupp/python-magic) and not file-magic. We're going to run into #166 again soon

paddatrapper:
File-magic was provided by the package python-magic from the source file until stretch. Since Feb 2018, the python-magic package is provided by the same upstream as pypi python-magic and src:file no longer provides a python package

paddatrapper:
I agree this should probably be held back and in the mean time, just require `file-magic<0.4.0` in the `setup.py` script.

The monkey patch would work, though the exact implementation would need to change due to the changes in file-magic 0.4.0 (#717).

paddatrapper:
I have managed to find the [ITP bug](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=877849) which describes some of the actions taken, but not why file-magic dropped the python package... By the looks of things file and python-magic have the same maintainer, who made the decision to move. The compat layer seems to be by design upstream at least.

hairmare:
Thanks for the link, I found this https://github.com/ahupp/python-magic/issues/180 which explains how the compat layer is a branch that integrates the libmagic api into the new python-magic.

This is somewhat dependent on https://github.com/ahupp/python-magic/issues/21 as well since that is where the original [libmagic-compat](https://github.com/ahupp/python-magic/compare/libmagic-compat) branch is from. We should probably wait with changing setup.py until the branch is merged.

I guess our best bet for now is to reply on the compat layer in packages and use the original api in our code. That way it will work both with the new and old packages. WDYT?

paddatrapper:
I think that is a good solution. I'll close this then

PR #728, opened 2019-02-05
fix podcast non audio enclosure problem

### LINKED ISSUE #675: Podcast feeds that contain non-audio enclosures fail to download
**Describe the bug**
If you have an RSS feed that includes multiple enclosures and one of them is not a audio file it will fail to download the podcast.

In some ways this isn't a bug because the RSS spec only allows 1 enclosure per item but mediaRSS and many CMS such as Wordpress and Drupal can be configured in such a way that the RSS contains other files in addition to audio as the enclosures. Also this bug only seems to happen if the audio file isn't the first enclosure.

**To Reproduce**
1. Add wcrsfm.org/audio/user/7/feed as a podcast
2. Try to import the latest show
3. It will fail
4. Look at the airtime-celery logs and see type None error

**Expected behavior**
It would download the audio file and ignore any images or other file types.

**LibreTime version:**
Github master circa alpha.6

**Installation method:**
 - OS: Ubuntu
 - OS Version [18.04] 
 - Method: vagrant
- Running the PR to test PHP 7.2 but confirmed on a separate system where the bug was discovered

**Additional context**
The error.
`
Jan 04 06:08:40 vagrant celery[3109]: [2019-01-04 06:08:40,927: INFO/MainProcess] Task podcast-download[php_5c2ef8681d60f1.35101349] succeeded in 0.800080617002s: '{"status": 1, "episodeid":
Jan 04 14:19:39 vagrant celery[3109]: [2019-01-04 14:19:39,954: INFO/MainProcess] Received task: podcast-download[php_5c2f6b7be86f80.50602071]
Jan 04 14:19:40 vagrant celery[3109]: [2019-01-04 14:19:40,029: INFO/Worker-1] podcast-download[php_5c2f6b7be86f80.50602071]: Error during file download: 'NoneType' object does not support i
Jan 04 14:19:40 vagrant celery[3109]: [2019-01-04 14:19:40,041: INFO/MainProcess] Task podcast-download[php_5c2f6b7be86f80.50602071] succeeded in 0.0868143140033s: '{"status": 0, "episodeid"
Jan 04 14:19:47 vagrant celery[3109]: [2019-01-04 14:19:47,687: INFO/MainProcess] Received task: podcast-download[php_5c2f6b83a6d953.08791423]
`

#464 is an attempt to fix this problem but in testing it I found it just hid any episodes containing a non-audio file vs. allowing them to be downloaded from the UI.

Robbt:
So this was inspired by #464 
 and fixes #675

The problem is that SimplePie appears to not support multiple enclosures of different types. At least with my test URL. From the [SimplePie docs](http://simplepie.org/api/class-SimplePie_Item.html#_get_enclosures)
```get_enclosures( )

Get all available enclosures (podcasts, etc.)

Supports the <enclosure> RSS tag, as well as Media RSS and iTunes RSS.

At this point, we're pretty much assuming that all enclosures for an item are the same content. Anything else is too complicated to properly support.```

So we had to manually search through the tags find enclosures and then look for audio ones and create a new enclosure object based upon these attributes.

This worked for my test URL but it would be handy to test it on others as well.
It only does the search if the first enclosure isn't an audio type and then it stops once it finds an audio enclosure and assumes that is the right one.

This should be a good workaround for poorly formatted CMS that add images as enclosures in RSS feeds.

### DIFF
--- a/airtime_mvc/application/services/PodcastEpisodeService.php
+++ b/airtime_mvc/application/services/PodcastEpisodeService.php
@@ -416,12 +416,18 @@ class Application_Service_PodcastEpisodeService extends Application_Service_Thir
                 // we need to check if this is an array otherwise sizeof will fail and stop this whole script
                 if (is_array($testenclosures)) {
                     $numenclosures = sizeof($testenclosures);
-                    for ($i = 0; $i < $numenclosures; $i++) {
+                    // now we loop through and look for a audio file and then stop the loop at the first one we find
+                    for ($i = 0; $i < $numenclosures + 1; $i++) {
                         $enclosure_attribs = array_values($testenclosures[$i]['attribs'])[0];
                         if (stripos($enclosure_attribs['type'], 'audio') !== false) {
                             $url = $enclosure_attribs['url'];
                             $enclosure = new SimplePie_Enclosure($enclosure_attribs['url'], $enclosure_attribs['type'], $length = $enclosure_attribs['length']);
                             break;
+                        }
+                        // if we didn't find an audio file we need to continue because there were no audio item enclosures
+                        // so this should keep it from showing items without audio items on the episodes list
+                        if ($i = $numenclosures) {
+                            continue;
                         }
                     }
                 }
--- a/airtime_mvc/application/services/PodcastEpisodeService.php
+++ b/airtime_mvc/application/services/PodcastEpisodeService.php
@@ -410,15 +410,23 @@ class Application_Service_PodcastEpisodeService extends Application_Service_Thir
                 // this is a rather hackish way of accessing the enclosures but get_enclosures() didnt detect multiple
                 // enclosures at certain points so we search through them and then manually create an enclosure object
                 // if we find an audio file in an enclosure and send it off
+                Logging::info('found a non audio');
                 $testenclosures = $enclosures = $item->get_item_tags(SIMPLEPIE_NAMESPACE_RSS_20, 'enclosure');
-                $numenclosures = sizeof($testenclosures);
-                for ($i = 0; $i < $numenclosures; $i++) {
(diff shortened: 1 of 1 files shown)


frecuencialibre:
this seems to be working great! i'm testing using this feed: https://noticiasdeabajoml.wordpress.com/feed/

the only UI improvement that i'll mention here, just in case it wouldn't be too difficult to include, is that should a feed <item> either 1. not include any enclosures at all, or 2. only include enclosures that are not compatible audio files, then that <item> (aka. episode) should not be included in LibreTime's listing of episodes that can supposedly be downloaded. 

awesome!

Robbt:
Yeah if a feed item doesn't have URLs in its enclosures then I don't think it will make the episode. We'd need to test it to see what happens with only non-audio items, have you done this yet ? It might work, I'm not looking at the code right now but we'd just need to create or find some example feeds. It might actually make sense to do some test-driven development here as well but alas I have not gotten into the swing/routine of that yet.

frecuencialibre:
awesome, so scenario 1 is already covered. 

regarding non-audio, i think that yes, they are being included. see the third “Las Insurrectas...” item here: 

screenshot from 2019-02-07 11-05-20

re: big daddy TDD haha yeah i too have gone around and around on that one.

Robbt:
Ok, I actually found a bug in my code where it had non-audio items or items without enclosures at all and I was able to fix it. I added another test that I'm not sure is necessary to try to ignore items with non audio enclosures. Please test and see if you are still seeing any issues. Yeah going all TDD on this would take some more effort in the front end but might help development speed up down the line.

frecuencialibre:
This works perfectly on my dev instance. re: TDD i think it makes most sense if you foresee continuing to actively develop a codebase for a good amount of time. Thanks @Robbt !

PR #731, opened 2019-02-08
added uploaded and genre column, removed album and sorted by uploaded to default library view

### LINKED ISSUE #579: Feature Request: Make default library view sorted by Column: Uploaded On and showing newest tracks
The Track Library view by default includes a column for Title, Create, Album, Length and is usually sorted by Title.

I want to suggest that we change this to include the Column Uploaded, and make the track sort so that newest tracks are at the top of the list.

I suggest this because most of the tech support I've done for LibreTime has been explaining to people who just uploaded a track how to find it and this involves clicking on columns, finding uploaded, checking it and then clicking on uploaded. 

So this is the most useful view for most people. Once you start searching you can sort by title, or length etc because you have a specific goal for the library. If you are just opening the library without a idea what to look for it is probably because you are looking for a track you just added and requiring search for that is not easy.

Robbt:
This fixes #579 
Everytime I show someone how to use LibreTime this is the first thing I do, (click columns and uploaded and then sort by newest) so I think that making it the default view makes sense as most of the time people will be most interested in scheduling the tracks they just uploaded.


Robbt:
I also added Genre and removed Album. This is more of a subjective decision but one that could encourage people to use Genre as a place to group "types" of tracks such as station IDs etc and Album is less likely to be something that people need to see by default when ordering tracks. This was also in response to a suggestion from a user on the [discourse forum](https://discourse.libretime.org/t/about-library-layout/221)

frecuencialibre:
I tested this and it worked as expected with a fresh vagrant install and it looks good to me. This change will have no effect if you've ever modified the columns in your instance, as that config will then be saved in the database and also your browser `localStorage`. Also it appears that these settings persist across multiple users.

PR #735, opened 2019-02-12
Fix podcast missing extension

### LINKED ISSUE #721: "Bad file extension" error upon m4a podcast import
**Describe the bug**
Podcasts fail to import. I'm seeing this error in zendphp.log:

```
2019-01-24T18:39:16+00:00 ERR (3): fl.local [MediaController.php:161 - postAction()] - Bad file extension.
#0 /usr/share/airtime/php/airtime_mvc/application/models/airtime/CcFiles.php(99): CcFiles::createAndImport(Array, '/tmp/phpP6Lmrw', 'videoplayback')
#1 /usr/share/airtime/php/airtime_mvc/application/modules/rest/controllers/MediaController.php(145): CcFiles::createFromUpload(Array)
#2 /usr/share/airtime/php/vendor/zendframework/zendframework1/library/Zend/Controller/Action.php(516): Rest_MediaController->postAction()
#3 /usr/share/airtime/php/vendor/zendframework/zendframework1/library/Zend/Controller/Dispatcher/Standard.php(308): Zend_Controller_Action->dispatch('postAction')
#4 /usr/share/airtime/php/vendor/zendframework/zendframework1/library/Zend/Controller/Front.php(954): Zend_Controller_Dispatcher_Standard->dispatch(Object(Zend_Controller_Request_Http), Object(Zend_Controller_Response_Http))
#5 /usr/share/airtime/php/vendor/zendframework/zendframework1/library/Zend/Application/Bootstrap/Bootstrap.php(105): Zend_Controller_Front->dispatch()
#6 /usr/share/airtime/php/vendor/zendframework/zendframework1/library/Zend/Application.php(384): Zend_Application_Bootstrap_Bootstrap->run()
#7 /usr/share/airtime/php/airtime_mvc/application/airtime-boot.php(84): Zend_Application->run()
#8 /usr/share/airtime/php/airtime_mvc/public/index.php(68): require_once('/usr/share/airt...')
#9 {main}
2019-01-24T18:39:32+00:00 WARN (4): fl.local Celery task DbId: 165
DbTaskId: php_5c4a058a527e36.67507114
DbTrackReference: 165
DbName: podcast-download
DbDispatchTime: '2019-01-24 18:35:54'
DbStatus: SUCCESS
ThirdPartyTrackReferences:
  DbId: 165
  DbService: podcast
  DbForeignId: '166'
  DbFileId: null
  DbUploadTime: '2019-01-24 18:39:32'
  DbStatus: null
  CeleryTaskss:
    CeleryTasks_0: '*RECURSION*'
 episode 166 unsuccessful with message 400 Client Error: Bad Request for url: http://fl.local/rest/media
```

**To Reproduce**
1. Add one of these feeds with m4a enclosures: https://podsync.net/Z4JjXtcV3 or https://podsync.net/UVYAstc9m
2. The xml should parse correctly and show up, but then try to manually import an episode and you should get the error above.

**Expected behavior**
podcast episode imports successfully

**LibreTime version:**
current HEAD

**Installation method:**
 - ubuntu 16.04.5 LTS via docker multicontainer

Robbt:
This fixes #721 
This is a rather hackish way of dealing with the files that lose their original name and extension and it simply adds .mp3 to any files that get through the podcast download without a file extension. Non audio files will most likely fail during the mutagen side of things and this is just a way of bypassing the error that files without extensions fail the rest/media request.
I've tested it and it appears to work for the podcasts referenced in #721 and when you download the files from the interface the correct extension is appended. So I think this is a go.


frecuencialibre:
I was finally able to test this and import now works great! Unfortunately now I'm having an issue with scheduled playout of imported tracks. Interestingly previews of the tracks _do_ work, (ie. the popup window that loads http://yourLTinstance.com/audiopreview/audio-preview/audioFileID/4/type/audioclip), but then when i put them in shows i get silence for their entire track length, and the ON AIR red goes away. can you reproduce this? could this have to do with the fact that files from the [test feed](https://podsync.net/Z4JjXtcV3), such as http://podsync.net/download/UVYAstc9m/Apm-YyccCsQ.m4a, with its supposed m4a extension pre-redirect, are now being treated as mp3.

more notes for testers:
- a fresh install or re-running `python  /usr/share/airtime/php/python_apps/airtime-celery/setup.py install` is necessary in order to see this change in effect.

Robbt:
I'll test this but you should look at your logs in pypo-liquidsoap

frecuencialibre:
doh! already killed that instance, but will check now. 

would using python-magic be overkill? it seems from the discussion here https://github.com/LibreTime/libretime/pull/719 "our best bet for now is to reply on the compat layer in packages and use the original api in our code. "

paddatrapper:
What is still needed here? Is it ready for review and merge?

Robbt:
I think it needs testing but otherwise it should be ready to merge. Its kind of a hack to assume files without a suffix are mp3s so we should try to ensure it doesn't have side effects.

paddatrapper:
How do you suggest testing this? I haven't used the podcast stuff at all, so I am not very familiar with it

Robbt:
Well basically the way to test it would be to have a Podcast feed where the files were mp3s but they lacked the mp3 extension. I think that is the test case this was being developed for. 

It was kind of a hack approach to fix a problem that my station wasn't directly experiencing but @frecuencialibre identified with some audio they were trying to ingest.

paddatrapper:
@Robbt can you fix the merge conflict here?

Robbt:
So the only real difference was the formatting in python and the use of double quotes instead of single quotes so I kept the pretty formatting. Just need an RSS feed where the mp3 files omitted the extension for some reason. It might be a good idea to test it with files that have no extension but aren't mp3 to make sure that it doesn't create a new issue.

paddatrapper:
I don't know of any RSS feeds where this is the case and don't have the knowledge to create one

Robbt:
Ok, I'll try to craft one just to test it, not even 100% sure that this PR is a great idea but it was intended to help @frecuencialibre when they were running into weird malformed podcasts and if it doesn't break when the files aren't mp3s it is probably worth assuming they are mp3s just to try and import them.

PR #742, opened 2019-02-18
Apistatus

mikeopensauce:
Added some new data to the JSON output on the API call "onAirLightAction". This shows the status of the two main streams and also whether they are connected to the on air feed.
I use this to show on my wall clock whether we are actually streaming data and whether our stream is online as opposed to just trusting that my connection is OK at my end because BUTT says I am connected.
I have also fixed a couple of missing quotes on a couple of pages which were showing up in red in my code but I'm just being picky!

### DIFF
--- a/airtime_mvc/application/controllers/ApiController.php
+++ b/airtime_mvc/application/controllers/ApiController.php
@@ -185,6 +185,10 @@ class ApiController extends Zend_Controller_Action
         $result["on_air_light"] = false;
         $result["on_air_light_expected_status"] = false;
         $result["station_down"] = false;
+        $result["master_stream"] = false;
+        $result["live_stream"] = false;
+        $result["master_stream_on_air"] = false;
+        $result["live_stream_on_air"] = false;
 
         $range = Application_Model_Schedule::GetPlayOrderRange();
 
@@ -208,6 +212,25 @@ class ApiController extends Zend_Controller_Action
 
         if ($result["on_air_light_expected_status"] != $result["on_air_light"]) {
             $result["station_down"] = true;
+        }
+
+        $live_dj_stream = Application_Model_Preference::GetSourceStatus("live_dj");
+        $master_dj_stream = Application_Model_Preference::GetSourceStatus("master_dj");
+        $live_dj_on_air = Application_Model_Preference::GetSourceSwitchStatus("live_dj");
+        $master_dj_on_air = Application_Model_Preference::GetSourceSwitchStatus("master_dj");
+
+        if($live_dj_stream == true){
+            $result["live_stream"] = true;
+            if ($live_dj_on_air == "on") {
+                $result["live_stream_on_air"] = true;
+            }
+        }
+
+        if($master_dj_stream == true){
+            $result["master_stream"] = true;
+            if ($master_dj_on_air == "on") {
+                $result["master_stream_on_air"] = true;
+            }
         }
 
         $this->returnJsonOrJsonp($request, $result);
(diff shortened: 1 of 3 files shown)


rallep71:
How do I proceed to insert the patch? I can use it just fine to process it in node red :D

mikeopensauce:
> How do I proceed to insert the patch? I can use it just fine to process it in node red :D

Once this code has been pulled into Master you should be able to pull it down but I'm not yet sure of the next stage in the procedure. I've just joined this group!
Mike

rallep71:
can't I patch this manually now?
how does manual patching work?

ryanrain:
github patches of PRs can be found by appending `.patch` at the end of the PR url, eg: https://github.com/LibreTime/libretime/pull/742.patch

rallep71:
yes, i know. wget https://patch-diff.githubusercontent.com/raw/LibreTime/libretime/pull/742.patch and then? i don't know the commands to patch it, can anyone briefly explain that to me? thank you

ryanrain:
https://duckduckgo.com/?q=git+apply+patch&t=ffsb&ia=web
https://stackoverflow.com/questions/2249852/how-to-apply-a-patch-generated-with-git-format-patch
`git apply 742.patch`

Robbt:
So @mikeopensauce what problem does this fix. The code seems fine but I never had any issues with the master or show sources being displayed as connected. 
It's currently being setup through getCurrentPlaylistAction in the ScheduleController and I'm assuming this is passed to the dashboard.js and accessed.

Are you developing a stand-alone widget that accesses the API ? I don't see any problem adding this information to this API call but wanted to be clear as to whether you are attempting to fix this for a separate view than the built-in dashboard (which has always worked for me).

mikeopensauce:
Hi Robbt
This isn't a fix. This is a feature. We use a studio clock which can display things like On air status and stream status. Currently we use some local code to check whether our stream is connected to our LibreTime server.
These additions allow us instead to display whether a) the stream is connected at the Libretime end and b) if that stream is actually live.
Currently I have to believe that my streamer is telling the truth. Libre time may be on air but my stream may have dropped and it could be playing from a scheduled playlist instead of the stream.
I thought it might be useful to others to provide those statuses (stati?) when they access the on air API call.

Regards
Mike

Robbt:
Ok this code looks good. Yeah, I just hadn't applied the previous fix that you pushed through on my local master branch and thus was running into the 500 errors you fixed. I really appreciate the work on improving the API. I think the next best thing to do would be to document our API properly and determine what functionality we could add to it so that we can prepare for a next version of the project that is less dependent upon zend1 MVC.

I'm going to merge this now.

rallep71:
https://libretime.org/manual/airtime-api-authentication/

http://libretime.server.box/api/api-action/format/json/api_key/xxxxxxxxxxxxxxxx

Bad Request!

The requested action is not supported!

hm, how can i connect to this api?

mikeopensauce:
Rallep71 you need to change api-action for the action you are calling as per the instructions [here](https://libretime.org/manual/airtime-api-authentication/) e.g. http://libretime.server.box/api/on-air-light/format/json/api_key/xxxxxxxxxxxxxxxx

rallep71:
http://libretime.server.box/api/on-air-light/format/json/api_key
key not need it?

output {"on_air_light":true,"on_air_light_expected_status":true,"station_down":false}

with my key the same output
http://libretime.server.box/api/on-air-light/format/json/api_key/xxxxxxxxxx

output {"on_air_light":true,"on_air_light_expected_status":true,"station_down":false}

but who can i see whats playing, who is on air....etc....

rallep71:
Does the api work at all or is it turned off?
http://libretime.server.box/api/life-info/format/json/api_key/

life-info does not exist

the api documentation doesn't really help either

Robbt:
http://libretime.server.box/api/live-info/api_key/#### should work - it is live vs. life

PR #745, opened 2019-02-18
fixed autoloading playlist to avoid errors when trying to schedule empty playlist

### LINKED ISSUE #411: Over 300,000 empty playlists created by LibreTime
This has happened twice now on our installation. Log in to LT to make a change on the calendar. The calendar takes several minutes to load. Look in Playlists and there are over 350,000 empty playlists that have been created by the system. The solution is to manually go into the database and use an SQL command (`DELETE FROM cc_playlist WHERE creator_id IS NULL;`) to clear out the playlists.

Does anyone know why this happens?

screen shot 2018-01-17 at 11 18 17 pm

Robbt:
This fixes #411

### DIFF
--- a/airtime_mvc/application/common/AutoPlaylistManager.php
+++ b/airtime_mvc/application/common/AutoPlaylistManager.php
@@ -13,7 +13,6 @@ class AutoPlaylistManager {
      * @return bool true if $_AUTOPLAYLIST_POLL_INTERVAL_SECONDS has passed since the last check
      */
     public static function hasAutoPlaylistPollIntervalPassed() {
-        Logging::info("Checking autoplaylist poll");
         $lastPolled = Application_Model_Preference::getAutoPlaylistPollLock();
         return empty($lastPolled) || (microtime(true) > $lastPolled + self::$_AUTOPLAYLIST_POLL_INTERVAL_SECONDS);
     }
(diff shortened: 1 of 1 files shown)


airtime_mvc/application/common/AutoPlaylistManager.php
```diff
@@ -13,7 +13,7 @@ class AutoPlaylistManager {
      * @return bool true if $_AUTOPLAYLIST_POLL_INTERVAL_SECONDS has passed since the last check
      */
     public static function hasAutoPlaylistPollIntervalPassed() {
-        Logging::info("Checking autoplaylist poll");
+        //Logging::info("Checking autoplaylist poll");
```
paddatrapper: Is there any point in keeping this as a comment? Why not just delete it?

airtime_mvc/application/common/AutoPlaylistManager.php
```diff
@@ -60,7 +64,17 @@ public static function buildAutoPlaylist() {
                 elseif (!$repeatuntilfull) {
                     break;
                 }
-
+                // we want to avoid an infinite loop if all of the playlists are null
+		if ($playlistid == null && $introplaylistid == null && $outroplaylistid == null) {
```
paddatrapper: Inconsistent indenting (tab vs spaces)

Robbt:
Ok, I pushed a commit to fix the formatting issue and remove the commented out logging info and then rebased it and forced it so that it is still a single commit. (trying to be cleaner about my PRs).

PR #748, opened 2019-02-22
Podcast interface fix

xabispacebiker:
The podcast section is not displayed properly as attached image:

captura de pantalla 2019-02-22 a las 12 34 30

Apparently the relevant css was removed from styles.css, so instead of adding it again I just added additional "podcast-metadata-row" divs to the layout.  

captura de pantalla 2019-02-22 a las 13 08 15

Also removes the save button at the bottom, as it has no sense. (There is already one at the top)


frecuencialibre:
awesome!!!

i think the save button at the bottom was not deleted, but i actually prefer to keep it, consistent with the show modals beside the calendar.

i also see that the "Unpublish" button is also broken and that that was caused by my changes in 8f8e9ef3f317b6c3ec5b2e7c6b5505a8bcd9bfaa . sorry about that, i'll try to track it down.

PR #768, opened 2019-03-06
[wip] added expires metadata to tracks

### LINKED ISSUE #622: Feature Request: Add an Expires On to Track Metadata for Smartblock Filtering
One problem with automatic scheduling of tracks is that certain date sensitive tracks such as promotions for upcoming events or seasonal notices can become stale and sound out of place if aired. Human schedulers can make this determination based upon contextual awareness and/or the addition of dates into the track title but the smartblock filters have no way of incorporating this without it being explicitly set in the code. 

The solution would be to add an additional column to the database that is called 'expires on' and this would allow someone to edit a track and select a date that the track shouldn't be aired after. In addition the smartblock filter would have this field added as a date field so that people could filter their smartblocks to not air if the 'expires on' is after now.

**Describe alternatives you've considered**
The challenge here is that it will require additional work be done editing tracks after uploading and more complex smartblock setups that won't be intuitive if people don't understand how to set them up. Another option could be to have a script hide or delete a track after it has expired so that it doesn't show up on the tracks database. Other alternatives could be to have a user entered parseable date in the title with a key word such as EXP: 12/31/18 but this would be too easy to mess up the syntax although it would be easier for a human to read. 

Another option would be to allow the ability to set air between date on a track. And give it a date range ie a starting date and an ending date. This would make it easier to setup future promos that will air for a set period of time and potentially simplify the smartblock creation to have it filter tracks where now is between their airing date.

**Additional context**
The most explicit way of doing this is probably to simply add another field to the database. As we currently don't write metadata back to tracks we don't need to consider how this data will be saved in the track metadata but if we decided to do something like this in the future it would require a field setup to store this particular information.

This could probably use some mockups and/or workflow planning before we start actually coding anything to determine what makes the most sense from a UI/UX perspective.

Old description:
This will benefit the ability for us to automate time sensitive PSA and announcements. Basically as a radio station you will want to run certain announcements only up until a certain date. For instance a promo for an event, there currently isn't any way to automatically but if we added a new date column that was user selectable in the Edit Tracks section this could make it possible to have smartblocks automatically filter out tracks that are expired.

Robbt:
This fixes #622
I need to fix UTC and allow null values (currently saves current time for expires upon track edit save)
also need to load values from database when tracks are edited and integrate it with smartblock


airtime_mvc/application/controllers/upgrade_sql/airtime_3.0.0-alpha.10/upgrade.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE cc_files ADD COLUMN expirestime timestamp(6);
```
hairmare: This is missing an alter on the default value of the type field.

Robbt:
So I have decided to refactor this because I think the inclusion of a expires date w/o a start date is incomplete.

Issue #604, opened 2018-11-22
Application_Model_Preference::setValue in 2.x -> 3.x upgrade


hairmare:
See https://github.com/LibreTime/libretime/issues/487#issuecomment-440754658 for where this was somehow triggered during a Airtime to LibreTime sidegrade... The "Invalid number of results" thing smells like a  weird edge case in the 2.x vs 3.x upgrade path.

https://github.com/LibreTime/libretime/blob/643117494366bb2f1cc15d10da1701a080134b8e/airtime_mvc/application/models/Preference.php#L62-L65

`//this case should not happen` is which I'm opening a separate issue.

@itcrowdsource what exact Airtime version did you upgrade from? Where you using watched folders previously?

It could be an easy fix if anyone finds the time to untangle the `Application_Model_Preference::setValue` method (which is quite a quagmire). Most likely we would need you to to some digging int SQL on the state of the `cc_prefs` table.

Whe should also improve the error message in the snippet above to tell us what value `setValue` was called on.

Also #71 ;)

PR #776, opened 2019-03-27
merged database schema changes to alpha.7 release

### LINKED ISSUE #771: Failed to import database from Airtime 2.5.1 to Libretime 3.0.0-alpha.6
**Describe the bug**
On a fresh install of libretime from git on a unbuntu 18.04 which works, the database import from an airtime 2.5.3 instance (following http://libretime.org/manual/backing-up-the-server/) brings errors and gives "Oops!
something went wrong!" : 

```
sudo tail -f /var/log/postgresql/postgresql-10-main.log
2019-03-10 16:36:54.323 CET [23737] airtime@airtime ERROR:  relation "celery_tasks" does not exist at character 378
2019-03-10 16:36:54.323 CET [23737] airtime@airtime STATEMENT:  SELECT celery_tasks.id, celery_tasks.task_id, celery_tasks.track_reference, celery_tasks.name, celery_tasks.dispatch_time, celery_tasks.status, third_party_track_references.id, third_party_track_references.service, third_party_track_references.foreign_id, third_party_track_references.file_id, third_party_track_references.upload_time, third_party_track_references.status FROM celery_tasks INNER JOIN third_party_track_references ON (celery_tasks.track_reference=third_party_track_references.id) WHERE celery_tasks.status=$1 AND celery_tasks.task_id<>$2
2019-03-10 16:36:58.029 CET [23754] airtime@airtime ERROR:  column "import_status" does not exist at character 64
2019-03-10 16:36:58.029 CET [23754] airtime@airtime STATEMENT:  SELECT "id", "name", "mime", "ftype", "directory", "filepath", "import_status", "currentlyaccessing", "editedby", "mtime", "utime", "lptime", "md5", "track_title", "artist_name", "bit_rate", "sample_rate", "format", "length", "album_title", "genre", "comments", "year", "track_number", "channels", "url", "bpm", "rating", "encoded_by", "disc_number", "mood", "label", "composer", "encoder", "checksum", "lyrics", "orchestra", "conductor", "lyricist", "original_lyricist", "radio_station_name", "info_url", "artist_url", "audio_source_url", "radio_station_url", "buy_this_url", "isrc_number", "catalog_number", "original_artist", "copyright", "report_datetime", "report_location", "report_organization", "subject", "contributor", "language", "file_exists", "soundcloud_id", "soundcloud_error_code", "soundcloud_error_msg", "soundcloud_link_to_file", "soundcloud_upload_time", "replay_gain", "owner_id", "cuein", "cueout", "silan_check", "hidden", "is_scheduled", "is_playlist", "filesize", "description" FROM "cc_files" WHERE "id" = $1
```
Is it possible to migrate the database from Airtime 2.5 to Libretime 3 ? Import only the datas as the schema.xml seems to have changed ?

thks

Robbt:
This fixes #771 in theory. Basically I added new database schema changes for new versions without adding the appropriate release info to Upgrades.php in the controller and so they were never automatically applied. If we are going to continue using alpha.7 for our next release and not jump versions then it probably makes sense to merge these back into a singular release SQL change.

This shouldn't break anything but it might require people to manually apply the database changes. If they had already upgraded their git to alpha.7. On the other hand the way it was setup currently they would already have a broken install if they hadn't applied the database changes.

So merging this will clean up the code for a new alpha.7 release, the alternative is to start the process of hopping releases and add incremental changes and new versions every schema change. So we would skip alpha.7 and jump to alpha.9 and I could add the incremental database changes to Upgrades.php. Either of these will work. The benefit to staging each schema change as a release # even without an official release is that anyone upgrading to the latest git will in theory always have the latest schema and shouldn't have to manually apply schema changes.

The drawback to this approach is that our versioning won't be as incremental and we will end up skipping versions numbers between actual releases. We just need to decide as a project which route we want to take. If people think we should aim to minimize disruption to people pulling directly from git master or whether we want to try and keep our release versions sequential.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ dist: trusty
 language: php
 php:
 # the latest and greatest, has some issues that are excluded below in matrix.allow_failures
+- 7.3
 - 7.2
 - 7.1
 # the 7.0 build demonstrates that everything is basically ok for 7.0, users might want to wait for 7.1 to run it
@@ -28,12 +29,15 @@ matrix:
   # there are some issues with phpunit as well as some deep in zf1 on 7.2
   - env: PYTHON=false
     php: 7.2
+  # there are some issues with phpunit as well as some deep in zf1 on 7.3
+  - env: PYTHON=false
+    php: 7.3
   exclude:
-  # by excluding all of python we make way to just runu python tests in one seperate instance
+  # by excluding all of python we make way to just run python tests in one seperate instance
   - env: PYTHON=true
   include:
   # using latest to run python on since it will last the longest
-  - php: 7.1 
+  - php: 7.3
     env: PYTHON=true
 addons:
   apt:
--- a/airtime_mvc/application/controllers/LibraryController.php
+++ b/airtime_mvc/application/controllers/LibraryController.php
@@ -364,6 +364,7 @@ class LibraryController extends Zend_Controller_Action
     {
         $user = Application_Model_User::getCurrentUser();
         $isAdminOrPM = $user->isUserType(array(UTYPE_SUPERADMIN, UTYPE_ADMIN, UTYPE_PROGRAM_MANAGER));
+        $isAdmin = $user->isUserType(array(UTYPE_SUPERADMIN, UTYPE_ADMIN));
 
         $request = $this->getRequest();
 
@@ -379,6 +380,10 @@ class LibraryController extends Zend_Controller_Action
             $form->makeReadOnly();
             $form->removeActionButtons();
             $this->view->permissionDenied = true;
+        }
+        // only admins should be able to edit the owner of a file
+        if (!$isAdmin) {
+            $form->removeOwnerEdit();
         }
 
         if ($request->isPost()) {
--- a/airtime_mvc/application/controllers/downgrade_sql/airtime_3.0.0-alpha.7.1/downgrade.sql
+++ b/airtime_mvc/application/controllers/downgrade_sql/airtime_3.0.0-alpha.7.1/downgrade.sql
@@ -1,0 +1,3 @@
+ALTER TABLE podcast_episodes DROP COLUMN IF EXISTS episode_title;
+ALTER TABLE podcast_episodes DROP COLUMN IF EXISTS episode_description;
+
--- a/airtime_mvc/application/controllers/downgrade_sql/airtime_3.0.0-alpha.7.2/downgrade.sql
+++ b/airtime_mvc/application/controllers/downgrade_sql/airtime_3.0.0-alpha.7.2/downgrade.sql
@@ -1,0 +1,1 @@
+ALTER TABLE cc_blockcriteria DROP COLUMN IF EXISTS criteriagroup;
--- a/airtime_mvc/application/controllers/downgrade_sql/airtime_3.0.0-alpha.7/downgrade.sql
+++ b/airtime_mvc/application/controllers/downgrade_sql/airtime_3.0.0-alpha.7/downgrade.sql
@@ -1,5 +1,1 @@
 ALTER TABLE cc_block ALTER COLUMN type SET DEFAULT 'static';
-ALTER TABLE podcast_episodes DROP COLUMN IF EXISTS episode_title;
-ALTER TABLE podcast_episodes DROP COLUMN IF EXISTS episode_description;
-ALTER TABLE cc_blockcriteria DROP COLUMN IF EXISTS criteriagroup;
-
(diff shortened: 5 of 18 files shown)


paddatrapper:
I think our database version and release number should be decoupled - each change to the model should have a DB version bump and each release is compatible with a specific DB version. That way we do not need to squash DB changes and it makes keeping up to date with the tip of git easier.

My suggestion would be to merge this, and then move to numbering database versions at 10 and each PR that modifies the DB incrementing that number

paddatrapper:
If there are no objections, I'll merge this on Wednesday and document the database schema pattern in the wiki

Robbt:
I'm fine with this.

Robbt:
So I read the wiki and I'm not sure exactly how the new versioning for the database is envisioned to work.

Here is the type of code that needs to be added everytime we do a database upgrade.

https://github.com/LibreTime/libretime/blob/4af83dfa6a9abe83427c03fc9ebf86875f2344b9/airtime_mvc/application/upgrade/Upgrades.php#L536-L548

It may be helpful to do some kind of versioning that simply adds another digit. For instance we could try
3.0.0-alpha.8.1 and 3.0.0-alpha.8.2 and increase that each release.

There was at least one issue related to the version documented in #604 that we should be aware of that someone ran into. It wasn't fully determined whether the versioning caused the error and I have tested an upgrade from 2.5.x installed via deb and it worked other than the missing installs.

Doing granular decimal upgrades would allow us to keep the database upgrades connected to a certain release but not force us to try to combine database changes from multiple PRs. The basic protocol would be to increment the version by .1 every time a database change is added.

To test this I could unroll these changes and make them incremental for updates to the alpha.7 release.

Either way the important thing is that we are consistent. And in that mind I would like us to start adding comments above the database changes pointing to the specific issues like was done in this upgrade.
https://github.com/LibreTime/libretime/blob/4af83dfa6a9abe83427c03fc9ebf86875f2344b9/airtime_mvc/application/upgrade/Upgrades.php#L464-L477

I'm ok with us using the Wiki for now to maintain the developer documentation as it is quick but it might be something that we want to build a more structured and robust section into our mkdocs site in the future as version controlling it in git will allow us to work together better and keep track of changes easier than the wiki itself. But yeah in general I like the changes that you did. I will try to add additional information about all of the steps required to do a database schema change and subsequent upgrade.

Robbt:
I added additional details for the other steps that are needed involving Propel to enact a database schema change. So feel free to respond here and we can discuss whether to do granular decimal changes or you can provide a counter example of what you mean by incrementing by 10. I just wasn't sure if you meant we would call the next database change .80 and then .90 ? 

In my mind the decimal change gets the best of both worlds. By doing these we will need to make sure that every schema change updates the Upgrades.php which I forgot to do when I submitting my last two PRs.

Also let me know if you think adding incremental alpha.7.1 and alpha.7.2 changes would be a better idea than just merging them together for this PR. I'm happy to do that as I think it makes more sense. I also have an alpha.7.3 change to submit to fix an issue I noticed with the podcast_descriptions but I have held off until we decide how we want to do versioning.

paddatrapper:
Actually your suggestion is better than what I was going for. My initial idea was to start DB versions at 10 (just to keep it distinct from the alpha.X that is currently there), then the next update to the schema would be 11, the following would be 12. But your alpha.X.Y decimal pattern works better for the reasons you mention.

Robbt:
Great I just pushed a  change to this PR to make it incremental and add the appropriate upgrade classes to Upgrades.php vs. merging everything underneath one update. Once this is tested and merged I'll push another fix for the podcast_description column I ran into.

Robbt:
Ok, I just added a super verbose with examples edit to the database update procedures. Hopefully people read it before they do database changes. It will be helpful for me to have a reference as well and will make database changes less tricky in the future. I'm going to boot up an Airtime 2.5.x instance and see if this change fixes the issue I ran into earlier.

paddatrapper:
Testing applying this on top of an existing git master instance seemed to work without issue

Robbt:
We need to test it when upgrading from a previous version of LibreTime and see if the database changes are applied. I had a bunch of syntax errors in the first commit I pushed. Somehow my virtual box ubuntu 14.04 install got hosed so I need to start over with a sample airtime install.

Robbt:
Ok after the numerous commits to fix mistakes. This looks ready to merge. I was able to test it from a airtime 2.5.1 deb installed on Ubuntu 14.04 and it appears that everything is working with LibreTime installing directly from the git. 

The next step will be testing the upgrade via a deb once we release alpha.7 and see if people can actually just dpkg -i libretime.deb on their old airtime systems (if there any still in operation).

PR #781, opened 2019-04-01
use local filesize check for pypo to avoid mismatch in file size comparison

Robbt:
This in theory fixes #756 because it will use the actual path size of a file vs. what is stored in the database (which could be slightly different and evidently is in some cases). This can cause a file that is currently being played to be overwritten by pypo thus causing dead air and possibly other issues down the line. 

This PR needs to be tested but based upon the bug report just removing the file copy all together solved all of the problems they were experiencing. I have had similar problems with pypo and I suspect that they are less frequent because we typically play longer tracks. 

I removed the reference to src_size being derived from the media field because I'm pretty sure it was different but someone could debug it further by printing the two sizes to the error log and running tests to confirm that this indeed is the case.

Either way I wanted to get started with a PR for this as this could help us solve a long standing problem and also maybe fixes #232 and might be related to #307.
It also hopefully fixes #699 
Since many of these bugs have been difficult to recreate I think it would be worthwhile to test and see whether with this patch applied we start to see a number of logs stating that a track already exists and preventing a copy.

This was what we should have seen but I suspect the file size comparison always failed and thus it just copied the file anyways potentially disrupting the track that was actually playing.

This could be something we need to readdress should we implement a cloud based S3 file system in the future for Libretime but for most install cases that are all depending upon a local file system hopefully this will work.

Please test and let me know. I'll share notes from my testing as soon as I get a chance.

PR #784, opened 2019-04-03
modified episodes_description to be longvarchar aka text

### LINKED ISSUE #780: Podcast episodes with long descriptions fail to import
**Describe the bug**
Basically we use a varchar of length 4096 for the episode_description column in the podcast_episodes table.

If a podcast contains a description longer than this it fails to import with this error.
```
Next PropelException: Unable to execute INSERT statement [INSERT INTO "podcast_episodes" ("id", "podcast_id", "publication_date", "download_url", "episode_guid", "episode_title", "episode_description") VALUES (:p0, :p1, :p2, :p3, :p4, :p5, :p6)] [wrapped: SQLSTATE[22001]: String data, right truncated: 7 ERROR:  value too long for type character varying(4096)] in /vagrant/airtime_mvc/application/models/airtime/om/BasePodcastEpisodes.php:762
Stack trace:
```

**To Reproduce**
1. Add a podcast with a episode with a long description
2. Click on the episode to download it
3. Watch it fail and look at the log

**Expected behavior**
It should import tracks regardless of the length of the description. I'll push a simple PR to change the database column to a TEXT vs. VARCHAR and it should resolve the issue.

**LibreTime version:**
Current master

**Installation method:**
Vagrant Ubuntu 18.04

Robbt:
This fixes #780 - it is a simple database change and shouldn't impact things other than resolving this issue.

### DIFF
--- a/airtime_mvc/application/controllers/downgrade_sql/airtime_3.0.0-alpha.7.3/downgrade.sql
+++ b/airtime_mvc/application/controllers/downgrade_sql/airtime_3.0.0-alpha.7.3/downgrade.sql
@@ -1,0 +1,1 @@
+ALTER TABLE podcast_episodes ALTER COLUMN episode_description TYPE VARCHAR(4096);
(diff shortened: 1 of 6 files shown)


paddatrapper:
Can this be merged before #781? I assume there is dependency based on the DB version numbers, but I guess that can be changed?

Robbt:
#781 doesn't touch the database so this can be merged without messing with it, as far as I know there are no other open PRs that touch the database other than this one.

PR #788, opened 2019-04-07
Update Stream Settings section of docs w/ new screen shots and link to icecast2 tutorial

Robbt:
This includes a notice to users that they should change their default password and fixes various incorrect statements.


paddatrapper:
Why did decide to use the wiki for the Icecast tutorial?

Robbt:
I just put the tutorials in the Wiki as a placeholder so that I could get feedback. My intention is to integrate them into the Docs but I wanted to figure out a best practice here and also refine the videos and determine where we want to host them. I'm not sure that we should rely upon YouTube as the primary host and probably not my personal account. 

So short answer, putting the video tutorials in the wiki intended as a placeholder.

paddatrapper:
Ah I see. Makes sense. I'm happy to merge this then. Maybe create an issue for the video hosting question?

I'm not sure the best way to provide feedback on the tutorials because the wiki has no comment/discuss functionality... Any ideas?

Robbt:
As far as commenting on the videos you could either comment on Youtube under the video itself or on the [discourse forum post](https://discourse.libretime.org/t/new-tutorials-available-on-libretime-wiki/270/2) with any recommendations or changes you suggest.

Thanks.

PR #802, opened 2019-04-22
Fix playlist smartblock remaining

Robbt:
Refactored the time remaining smartblocks to track the time limit when multiple time remaining smartblocks are added under a single playlist.
Previously it would just pass the same time remaining to both smartblocks which would result in overscheduling.
Now it factors in the time added for each smartblock in a playlist and passes this onto the block so that it doesn't over schedule..


Robbt:
This fixes #708

Issue #376, opened 2017-12-01
Problem: The documentation is out of date


Robbt:
This is an attempt to create a list of all assets and info on the LibreTime manual at libretime.org that need to be updated as they refer to an old version of 2.5.1 or we have changed the features etc.

## Pages in need of update

- [x] [Features](http://libretime.org/features/)  - is all TBD
- [x] [On The Air in 60 seconds](http://libretime.org/manual/on-air-in-60-seconds/) - needs instructions refined for LibreTime
- [x] [GettingStarted](http://libretime.org/manual/getting-started/) - remove info about Registering since we don't do that - and need to update Image assets w/ new screenshots thanks @nikmartin 
- [x] [Systems-General](http://libretime.org/manual/preferences/) - refers to settings that don't exist like Soundcloud and e-mail settings - how do we handle e-mail w/ LibreTime ?
- [x] [Users](http://libretime.org/manual/users/) - needs updated screenshots
- [x] [Media Folders](http://libretime.org/manual/media-folders/) - this functionality doesn't exist in LibreTime should be at least updated w/ a notice and link to PR
- [x] [Stream Settings](http://libretime.org/manual/stream-settings/) - at least need to update and remove Scheduled Play as we don't have that feature
- [x] [Support/Feedback](http://libretime.org/manual/support-feedback/) - put info about Fork and links to Github and discourse forum vs. referring to SourceFabric
- [ ] [Listen](http://libretime.org/manual/listen/) - needs new screenshots
- [x] [Upload](http://libretime.org/manual/upload/) - needs new screenshots for new UI
- [x] [Tracks](http://libretime.org/manual/tracks/) - needs updated screenshots
- [x] [Playlists](http://libretime.org/manual/playlists/) - needs updated screenshots and maybe updating how to create and manage playlists. Needs to talk about Autoloading playlsts
- [x] [SmartBlocks](http://libretime.org/manual/smartblocks/) - updated screenshots
- [x] [WebStreams](http://libretime.org/manual/webstreams/) - updated screenshots
- [x] [Podcasts](http://libretime.org/manual/podcasts/) - Replace the TBD stuff with content
- [x] [Calendar](http://libretime.org/manual/calendar/) - UPDATED - could use a little bit more info on automatic playlists
- [x] [Help](http://libretime.org/manual/help/) - needs links to LibreTime 
- [x] [Recording](http://libretime.org/manual/recording-shows/) - currently not supported by LibreTime - needs info regarding this
[...]

PR #807, opened 2019-04-24
Updated the upload page to match the current UI.

Robbt:
This basically gets rid of the obsolete add-media portion of the page and updates the docs to match the current UI.


Robbt:
This fixes part of #376

PR #809, opened 2019-04-24
Move library docs into new new menus

Robbt:
This follows up on #793 and represents an attempt to move the existing documentation into the newly created menu items that were introduced.

As of this first commit it does not update much of the text and adds a very limited podcasts page. It does get rid of the TBD. I will either make additional commits to this if I get the chance before someone reviews it or this could be merged as is because it improves the existing manual.


Robbt:
This fixes part of #376

Issue #485, opened 2018-04-28
No podcast PodcastService.php error


momai:
Hello! I installed LibreTime alpha-3 on debian 9 and I have a problem with going to feeds/station-rss, I get an error in the browser. Failed to load resource: the server responded with a status of 500 (Internal Server Error).

logfile zendphp.log, I see an error:

 `[ErrorController.php:54 - errorAction()] - An internal application error has occurred.: ErrorException: DOMDocument::loadXML(): StartTag: invalid element name in Entity,
 line: 2 in /usr/share/airtime/php/airtime_mvc/application/services/PodcastService.php:517
Stack trace:
#0 [internal function]: exception_error_handler(2, 'DOMDocument::lo...', '/usr/share/airt...', 517, Array)
#1 /usr/share/airtime/php/airtime_mvc/application/services/PodcastService.php(517): DOMDocument->loadXML('<?xml version="...')
#2 /usr/share/airtime/php/airtime_mvc/application/controllers/FeedsController.php(19): Application_Service_PodcastService::createStationRssFeed()
#3 /usr/share/airtime/php/vendor/zendframework/zendframework1/library/Zend/Controller/Action.php(516): FeedsController->stationRssAction()
#4 /usr/share/airtime/php/vendor/zendframework/zendframework1/library/Zend/Controller/Dispatcher/Standard.php(308): Zend_Controller_Action->dispatch('stationRssActio...')
#5 /usr/share/airtime/php/vendor/zendframework/zendframework1/library/Zend/Controller/Front.php(954): Zend_Controller_Dispatcher_Standard->dispatch(Object(Zend_Controller_Request_Http), Object(Zend_Controller_Response_Http))
#6 /usr/share/airtime/php/vendor/zendframework/zendframework1/library/Zend/Application/Bootstrap/Bootstrap.php(105): Zend_Controller_Front->dispatch()
#7 /usr/share/airtime/php/vendor/zendframework/zendframework1/library/Zend/Application.php(384): Zend_Application_Bootstrap_Bootstrap->run()
#8 /usr/share/airtime/php/airtime_mvc/application/airtime-boot.php(84): Zend_Application->run()
#9 /usr/share/airtime/php/airtime_mvc/public/index.php(68): require_once('/usr/share/airt...')
#10 {main}
`

Help please solve the problem.

hairmare:
It looks like the podcast isn't returning a valid RSS document. Can you share the URL to the feed?

momai:
[...]

PR #687, opened 2019-01-11
Fix PodcastService PHP error in podcast generation

ciaby:
Fixes #485

PR #810, opened 2019-04-25
Fix podcast service

Robbt:
This replaces #687 as it takes the work @ciaby did and adds additional fixes.


Robbt:
This fixes #485 - and basically allows the podcast to work. Thanks @ciaby for your initial work. Sorry it took 3 months to get around to fully delving into it.

Issue #511, opened 2018-09-05
Problem: The Podcast Datatable initially only shows 25 podcasts regardless of actual #


Robbt:
Basically there is a bug somewhere in the code that makes it so that the podcast-datatable only shows the first 25 podcasts until you select a larger number. It will say 25/25 entries even when there are 26 or more podcasts.

This is a tricky bug to fix as I spent a lot of time looking at the various javascript files to try to diagnose it but it is a messy spaghetti code nightmare and perhaps this will be fixed only when we move away from datatables to a modern javascript framework. I know that it was fixed in airtime.pro from back when we had a subscription to it but I don't know what is causing the error. It is possibly related to #124 but until I get a better grasp of how we are using datatables exactly then I'm pretty sure I will just be wasting my time trying to fix it.

Robbt:
Ok, it appears that the issue is that the dropdown is becoming the maximum number. Ie if you have 51 podcasts and 50 selected it will show Showing 1 to 50 of 50 entries. At this point it will make the maximum number stop at 100 because paging is not enabled. This is probably a hint as to where the problem is occurring now I just need to determine what line of code is incorrect in some code written hastily by someone years ago who is no longer affiliated with the project. I suppose I could ping him and ask him but I'm not sure what the status of the license Airtime.pro is using and if he would even remember or be allowed to share his fix.

frecuencialibre:
where does it say "showing  _____" ? i can't see it. screenshot?

hairmare:
I think It's about the "Show <num>" thing at the bottom right of the screenshot below... It's wrong until the users selects an actual number and then it's fixed for that user. I think this also affects all the other library pagination pages (ie. the screenshot below was made on the Tracks page which IMO is also affected).

<img width="753" alt="screen shot 2019-01-08 at 21 58 00" src="https://user-images.githubusercontent.com/116588/50858476-cc944c00-1390-11e9-8548-deb0370198cc.png">

Robbt:
Actually the # out of # is missing, I'm not sure where it went, evidently some change we pushed through had the side effect of removing this line. I'll look into it figuring it out and see when and what caused it to go away. Removing it fixes the problem of it showing incorrect values but was unintended.

paddatrapper:
Fixed in #813

Issue #660, opened 2018-12-23
Podcast library listing missing search and select-all checkbox UI elements


frecuencialibre:
see screenshots below

**To Reproduce**
Click "Podcasts" in the main menu on the left

**Expected behavior**
that podcasts UI be consistent with established UI as seen in playlists, smartblocks, webstreams

**LibreTime version:**
alpha 6

**Screenshots**

![screenshot from 2018-12-23 14-53-20](https://user-images.githubusercontent.com/36803137/50387503-89b15380-06c3-11e9-954e-899ca5f4e49d.png)
![screenshot from 2018-12-23 14-53-06](https://user-images.githubusercontent.com/36803137/50387504-89b15380-06c3-11e9-8c92-d9620cfbead9.png)

frecuencialibre:
Un-assigning myself here after a quick review of the code. The podcast functionality was added to the code base at a later stage, probably by a different developer, and accomplishes things filling the left-side pane with podcasts in ways that only partially follow the methods used for other content types. Following my recent adventures with #651, I'm deciding that fixing this one would be prohibitively time consuming for now.

Robbt:
In addition the podcast data table currently lacks pagination.
It is created in a separate spot using a widget developed that I suspect was part of a refactoring to make the datatables easier to produced.

Here is where the length is created -  https://github.com/LibreTime/libretime/blob/8ddfe91529b6a454bcb7b44af7a6ef1aca119455/airtime_mvc/public/js/airtime/widgets/table.js#L79

Robbt:
So I looked into adding the search/filter etc and it would time consuming because we'd need to add a filter functionality. Basically the other items use the library datatable whereas podcasts have their own separate datatable and the filtering/search functionality hasn't been added to this.

PR #813, opened 2019-05-01
Fix podcast paging

Robbt:
This fixes an issue where the list of podcasts wasn't correctly being displayed or paging. 
What happens is if you limit the query it will only display the amount of entries that your length is.

There is still more work to be done as this evidently causes it to show all of the podcasts and ignore the paging although the paging is correctly displayed.


Robbt:
So this issue was actually that the $result->count() that was being passed was being shrunk by the filter when what we need for that variable to be assigned is the total outside of the limit (which is actually right according to how data tables does server side pagination etc.

So this fixes the outstanding issue of the total # of podcasts being limited mentioned recently in #660

Robbt:
This fixes #511

PR #814, opened 2019-05-01
Add podcast import date column to podcast page

Robbt:
This adds a new field to the podcast datatable called Import Date which is basically the last time an episode was automatically downloaded by the system. This can be helpful in tracking podcasts that aren't being updated. 

This only updates when the system ingests the podcast and it also doesn't catch failed attempts where there is an issue with the RSS feed. They'll still have the date updated if there are new items even if it fails to download them.

I made it visible by default.


paddatrapper:
Does this manage the migration of the DB?

Robbt:
The information was already in the database, so this was unnecessary, it just needed a join with the ImportedPodcast table which already tracked the last time a specific podcast attempted to download a new episode.

paddatrapper:
Ah Ok. I'm happy to merge once the merge conflicts are sorted

Robbt:
Ok, fixed the conflict.

PR #818, opened 2019-05-08
Smartblock sort tracks by last play date

### LINKED ISSUE #775: Provide a way for smartblocks to return unplayed tracks
I'm trying to program multiple irregularly published podcasts in a single show. For this, I'd like to be able to configure a smartblock to return content that has never been played. Unfortunately none of the modifiers available for the `Last Played` search criteria seem to be able to do this, not even `is not`. 

1. One possibility would be to provide a "never" modifier to the `Last Played` criteria. 
2. Another would be add `null` capable "not after" and "not before" negated modifiers: This would perhaps be more intuitive for folks, but you'd have to reason that in order to get tracks that have never been played you'd have to set it to "not before" 1 minute ago. 
3. both 1 and 2

How's this sound from a UI perspective?

Additionally, the `is` and `is not` modifiers to `Last Played` seem to have no effect. Can anyone reproduce this?

Thanks!

Robbt:
This fixes #775 by adding the ability for smartblocks to be sorted by the most recently played tracks (tracks that were played recently come to the top) or least recently played (tracks that were never played come to the top of the list and it is basically randomly selecting between tracks that match the criteria that have never been played and then if all tracks have been played once based upon the 'lptime' track it'll play the track that was played furthest in the past.

This will allow smartblocks to choose unplayed tracks by setting the sort order to be least recently played.

The docs should also be fixed include this new option but we still need to rewrite the existing smartblock docs page.

### DIFF
--- a/airtime_mvc/application/models/Block.php
+++ b/airtime_mvc/application/models/Block.php
@@ -1762,10 +1762,10 @@ SQL;
             $qry->addAscendingOrderByColumn('utime');
         }
         else if ($sortTracks == 'mostrecentplay') {
-            $qry->addDescendingOrderByColumn('lptime');
+            $qry->addDescendingOrderByColumn('(lptime IS NULL), lptime');
         }
         else if ($sortTracks == 'leastrecentplay') {
-            $qry->addAscendingOrderByColumn('lptime');
+            $qry->addAscendingOrderByColumn('(lptime IS NOT NULL), lptime');
         }
         else if ($sortTracks == 'random') {
             $qry->addAscendingOrderByColumn('random()');
--- a/airtime_mvc/application/models/Block.php
+++ b/airtime_mvc/application/models/Block.php
@@ -1202,7 +1202,6 @@ SQL;
                 // that might contradict itself we group them based upon their original position on the form
                 $criteriaGroup = $i;
                 foreach ($p_criteriaData['criteria'][$critKeys[$i]] as $d) {
-                    Logging::info($d);
                 	$field = $d['sp_criteria_field'];
                 	$value = $d['sp_criteria_value'];
                 	$modifier = $d['sp_criteria_modifier'];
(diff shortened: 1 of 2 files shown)


airtime_mvc/application/models/Block.php
```diff
@@ -1758,6 +1749,8 @@ public function getListofFilesMeetCriteria($showLimit = null)
         // check if file exists
         $qry->add("file_exists", "true", Criteria::EQUAL);
         $qry->add("hidden", "false", Criteria::EQUAL);
+
+        //$qry->addAsColumn('lptime','COALESCE("lptime",\'1970-01-01\')');
```
paddatrapper: Does this comment serve a purpose?

Robbt: Nope it was a failed attempt at accomplishing what I ended up doing in the last commit. I spent many hours digging into Propel attempting to figure out the sort only to realize that it could be done rather easily by modifying the sort statement to sort nulls first with parenthesis. Funny how walking away from a problem and coming at it fresh can sometimes result in instant fixes. I'll push a quick commit to remove this.

Paomas81:
This feature is very interesting, but i think is usefull to implement a filter to choose song not scheduled (i see that in tracks' coloumn this characteristic is signed). This because if i generate a playlist with some same smartblocks, they choose the same songs.

Paomas81:
image

Paomas81:
image

Robbt:
I get what you are saying @Paomas81 but the field in the database tracking files doesn't currently keep track of when a file was "scheduled" only the last time it was actually played out, so this would be a new feature request requiring the database to track when a audio file was added to the schedule anywhere. It might be worth investigating, feel free to suggest it as a new feature request.

Paomas81:
Thanks so much @Robbt . I open a new feature request #923 .

PR #824, opened 2019-06-02
Use official Debian Buster Vagrant image

paddatrapper:
The official Debian Buster Vagrant image is out. We should move to it instead of using an unknown provider

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -60,8 +60,7 @@ Vagrant.configure("2") do |config|
     provision_libretime(os, "debian.sh", installer_args)
   end
   config.vm.define "debian-buster" do |os|
-    # TODO: Replace with generic/debian10 once it is released
-    os.vm.box = "fujimakishouten/debian-buster64"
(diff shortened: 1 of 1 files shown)


Robbt:
I just tried to test debian-buster vagrant image and using this new one failed because they haven't updated the image yet to reflect that Debian 10 is now stable. I am currently testing the bento/debian-10 vagrant image as I don't think the "official images" are as well supported as the bento ones or the generic/debian10 which has over 87,000 downloads vs. the debian/buster one which has 5. The generic one supports libvirt and virtualbox so it maybe preferable.

hairmare:
I ran into some issues (mainly vagrant-vbguest not finding dkms and kernel-headers) with "debian/buster64" as well. I was able to work around them using "generic/debian10". I'm currently on Virtualbox.

paddatrapper:
As much as I prefer to use official images, I guess we should probably use generic/debian10 then. I'll open a PR

PR #837, opened 2019-07-04
added null to try to solve missing arguments bug for Podcast Publish function

### LINKED ISSUE #836: Not able to publish to My Podcast
Why am I not able to publish to My Podcast? Also, publishing to Soundcloud is not working for me either. It seems like I have everything working in Libretime except My Podcast. I'm able to import an external podcas, but if I want to publish a track from the "Tracks" page, the Publish button doesn't do anything besides giving an error:

Failed to load resource: the server responded with a status of 500 (Internal Server Error)
http://--ip-address--/rest/media/883/publish

And for the record:  angular.min.js.map seems to be missing.
Failed to load resource: the server responded with a status of 404 (Not Found)

Robbt:
This might fix #836 - I have not tested it yet.

### DIFF
--- a/airtime_mvc/application/services/PodcastEpisodeService.php
+++ b/airtime_mvc/application/services/PodcastEpisodeService.php
@@ -111,7 +111,7 @@ class Application_Service_PodcastEpisodeService extends Application_Service_Thir
      * @throws Exception
      * @throws PropelException
      */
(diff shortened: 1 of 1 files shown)


Robbt:
This appears to work. I tested it and it solves the bug without introducing any new side effects (that I could find).

PR #841, opened 2019-07-16
Use generic buster boxes

paddatrapper:


paddatrapper:
Haven't had a chance to test this yet though

hairmare:
Thanks @paddatrapper 

It worked like this when I tested it with these exact changes this morning :)

Maybe we should give `debian/buster64` another try given that it would be preferable? 

See: https://github.com/LibreTime/libretime/pull/824#issuecomment-511817244

paddatrapper:
I'm currently at DebConf (the Debian Conference), so I'll try find one of the people who maintain those images and chat to them about their status. The generic boxes "just work", which is definitely a plus. I don't have very strong feelings either way, but I tend to trust official images more than others.

PR #848, opened 2019-08-02
Update index.md

learning-more:
Spelling changes:
From: "The default installer script create a user.."  to: The default installer script createS a user.."
From: "which is far to open.." To: "which is far too open"


paddatrapper:
Thanks!

PR #876, opened 2019-09-18
Updated jquery.cookie to js.cookie

codenift:
Replaced  jquery.cookie with JS Cookie.


paddatrapper:
Haven't tested it, but LGTM

PR #877, opened 2019-09-20
Generate artwork image from audio using ID3.

codenift:
This feature will retrieve album from ID3 and generate thumbnails from audio on upload. Preview of artwork on library and for currently playing tracks on Dashboard header. For previously uploaded tracks (if using on live site before this PR), you can retrieve album artwork by simply re-saving track info (Using Edit tab). 

Todo: Artwork upload via Edit tab.

API Example:
your-site/api/track?id=1521&return=artwork

Other sizes: 32, 64, 128, 256, 512

Example:
your-site/api/track?id=1521&return=artwork-128


Robbt:
So I noticed that your database was using 2.5.17 but we are using libretime3.0 alpha for our database stuff so I'm going to try to push a commit to fix that to change it to airtime_3.0.0-alpha.9.1

codenift:
I wasn't sure. Thanks.

Robbt:
I think this is ready to merge but I noticed that the artwork doesn't show up on the box when the track is playing. It did show up when I went to edit a track.

codenift:
That's odd. If it showed up in the edit window it should have shown in the now playing box as well. Can you see the artwork in the library list as well? There's still room for improvement on this, but should work for now. I have it working on one station with an iPhone app fetching artwork from the LT API already and had no issues so far.

codenift:
I will try to look into it tonight to see if I find anything wrong.

codenift:
I wasn't able to find anything on my side, everything was working fine... although I haven't tested it with the changes you made to the database. I was only able to test it on my Mac right now with the major browsers. Are you able to test my fork directly when you have time to see if you find the same issue on your side?

https://github.com/codenift/libretime

There's 4 branches. 
**js_cookie** (PR already merged to main LT)
**artwork** (PR in main LT)
**media_type** (PR in main LT)
**catalina_support** (PR in main LT)

These are all merged into master in my fork so you're able to test using master.

Robbt:
It looks like you just pulled in a bunch of unrelated commits into this ie the Catalina stuff. I think we need to figure it out how to revert that before we can merge this. I'm not a git expert and I've ran into issues like this before but I think it is related to you merging your branches into your own master ?

You should be forking from the current project master and then not merging anything unrelated to the PR into your branch and I think that would fix it. It's probably doable to revert the last commits that went in.

codenift:
Yes, paddatrapper informed me yesterday and now I have a better understanding on how to do it. I wished Github had an option to just make PR without dragging all other commits allowing one to merge in a fork, although I know I can just clone or pull from a branch it doesn't make it easier for someone else to just look into your fork and clone it. I will make that change in order to reflect this project and from now on I should have clean PRs.

Robbt:
Ok, I am testing this right now and I managed to get it almost working but the lines in  airtime_mvc/public/js/airtime/dashboard/dashboard.js          
```
  var check_current_song = Cookies.get('current_track');
  var loaded = Cookies.get('loaded');
      if (check_current_song != currentSong.name) {

```
are failing because Cookies is undefined. This is under Firefox. 
in Chromium the whole Dashboard and tracks are empty and the site is broken because of the same thing with the error.
Uncaught ReferenceError: Cookies is not defined

So it is related. I had to shift-reload the site to get the latest javascript but it is still failing, but I'm not sure why.

Another thing I noticed is that we need to run composer install to add the new composer php id3 library that is included with this otherwise the site fails.

This should happen if someone runs ./install after they upgrade but we haven't really designed ./install to necessarily be ran to just upgrade the site. So it is a new upgrade challenge related to merging this because our current "upgrade script" just touches the database.

Robbt:
Ok, so for some reason I have the jquery.cookie.js instead of the new cookie.js in my clone of this branch even though that was included with this PR.

Robbt:
So manually altering the cookie.js include and adding file on my computer made everything work. I hadn't updated my master branch where that was included before 

So we just need to revert the unrelated commits and we could probably merge this. @codenift let me know if you need any help removing the unrelated commits. I think you can probably do it with git-revert commit to rollback to the database changes I made.

We will also need to think about how we are going to notify people of the need to run either 'composer install' or the install script.

codenift:
Yes, that was the part I didn't understand at first, especially since the "Artwork" PR was made before the "Media Type" one. I was holding on to those to see what you plan to do with them before I can clean up everything in my fork (get the master back to the original LT and leave it untouched) now that I know how to do the PR correctly.  I haven't had much time to learn everything for Github so I might need help removing unrelated commits, I see how git-revert where you can just add the hash for the commits but not 100% sure since it looks like commits I've made after are showing up here. But, I also have seperate backups on my computer of both branches (Artwork and Media Type) so if I clean up the fork now I can push both branches the right way and we should have clean PRs and I wouldn't take much. For me this maybe easier. Let me know what you would like to do.

I could add a check to see if a function for getid3 exists and just pop out a message. And for people using it on a production server they would use .install to update regardless.

Robbt:
So your master can be different, you just need to base the PR fork off of the project master. My master is far behind the main repo and it doesn't affect me for the most part.
The basic workflow I use is.
`git checkout master`
`git pull origin master` - if you have origin as your master you need to have a remote pointing to the main git repo of the project
`git checkout -b 'branch-name'`
Do my coding, add the commits and then push it to my repository
`git push robbt branch-name` - type in my credentials and when I refresh my view of my repo on github it asks me if I want to open up a new PR.

Once that PR is merged I can pull it back into my project master.
`git pull origin master` and I'll have it on my local server

If you don't have both the project master repo and your repo you can add one or the other with
`git remote add name repoURL`

I'm not sure if you are doing most of your stuff through the github UI, I haven't really used it aside from clicking the "merge pull request" button and commenting on issues etc. But the workflow I wrote up above has worked for me and maintains the separation of concern.

I can try to revert the commits and push it back to your PR (as a maintainer of the project the default permissions give me permissions to push changes to a PR submitted to the project).

Robbt:
So I pushed a revert for the various commits that were unrelated to this PR.
`git revert hashcode` - I did have to clean up a conflict where the previous database upgrade schema popped up again and I had to manually edit and commit that out again' - but otherwise it was pretty straightforward.

It might be a good idea to put a check around the getid3 and see if it exists and pop-up a message or at least keep the software from crashing when that library isn't included. 

I'm not sure we want to wrap every new library with a check to see if it exists and a message though. @paddatrapper do you have any thoughts on this ?

codenift:
Thanks for that. 

We could add the check for the getid3 in the two functions that use them instead (saveArtworkData and resetArtwork) with a conditional to leave it empty which will give them the default image. I should have added this after moving the library to composer. I didn't think about it.

codenift:
So I added a check to see if the library exists on both functions I mentioned, that should prevent the site from failing and just give them default image. So we will just need to inform them. There is a message they can see in the log if it fails to let them know they need to run install, but we could always add a popup if you want. Up to you guys. For all others (as in running it on production server) they will need to run install regardless.

paddatrapper:
I think logging the issue and noting it in the release notes should be enough. The install script should run composer install on upgrade - it it doesn't we need to fix that. That way we can assume that libraries are present in the code and not have to wrap everything in barriers.

codenift:
Yes, the install should run the composer install on upgrade fine, I didn't have any issue doing so. The check/message is basically there for people using Vagrant and haven't upgraded. I think I word the message incorrectly because if they message appears while on a production server they might get confused with the run "composer install" since that alone wouldn't work on production.  I'll update the wording now and keep it simple so they don't get confused. You can change it if you have a better way to word the message.

Robbt:
Ok I'm going to merge this. Thanks for the contribution @codenift

PR #878, opened 2019-09-20
Media Type for tracks

codenift:
Select Media Type on audio tracks (Station ID, Show Intro, Show Outro, Jingles, Sweepers, Commercials, Promo, Interview, Shout Out, News, and Voice Tracking).


codenift:
I'm not sure if I'm doing the separate PR correctly. Does it look good?

paddatrapper:
This PR includes commits from your other PRs. This PR should only include 
b411d6d, and not the others

codenift:
I tried. Not sure what else to tell you. I’m working using Atom made a new branch for each, js_cookie then made PR, artwork then made PR, media_type then made PR. So I guess that’s not the way to do it since every time I create a new PR it drags the previous commits with it. I also noticed it adds a lot of white space I didn’t add on the files. Well, I’m confused with the process.

paddatrapper:
No problem. It shouldn't be a problem as this depends on the previous PR anyway. I don't know how Atom manages branches. I hope to get some time next week to test the changes and then merge them

codenift:
Awesome

Robbt:
Yeah, I'll try to check this one out. In my thinking the media type would actually be a user configurable setting with some reasonable defaults, since every station might want to classify tracks there own way. It would also make sense to integrate it into smartblocks as a selection crtieria. I'm not sure if this does that yet as I haven't played with it yet. This is a good first step either way, thanks contributing.

As far as git stuff goes it can be tricky, to figure out the exact steps to do a clean PR.

codenift:
I was going to add an option since I believe you mentioned it before but
for now I left it like that. Yes, it's in Smart Blocks already. Would be
nice to select in Smart Blocks itself but you will just need to add key
values/codes (eg. MUS, COM, JIN, SWP) so choosing "Media Type - Is - MUS"
for example. Another idea that could be added later is an optional setting
to set Default on Import, pretty much how SamCloud does it where all import
set itself to MUS so if uploading a bunch of songs you won't need to set
them up for each.

Robbt:
So I'm looking at this and I think for my station at least there are too many media types that could probably be lumped together under a single type. For instance Sweepers, Promos, Intros, Outros, Voice Tracking could be considered the same type. https://en.wikipedia.org/wiki/Radio_sweeper

I'm currently mocking up some prototypes of a way to upload tracks directly into a specific folder/media type and as you can see in this mockup, the current types in this PR  presents a large # of options for someone to choose from. In my opinion this is more than we should include by default.
MediaTypeUploadFolder

codenift:
Well, after working for the radio for quite a bit, I can only tell by the way we use it. Like in the page you referenced says "segues between songs", I'd say much less than 10 seconds than 20. Example: "We continue with more music" unless it has some effects to extend it a bit more when overlapping on the crossfade. I just think it needs more control since it can get lost in crossfade, sounding lower in some cases. 

As far as the others, we will need them separate since it actually differentiate from each other. Examples:

Intro: That's something I would only want playing once at the beginning of the show only. 
"Welcome to the ____ show!"

Outro: Only at the end of the show 
"Thanks for listening to ____, we'll be back tomorrow with another episode of ___"

Voice Tracking: I see this as any other voice tracking not related to MT that involves voice. Sometimes you want to construct a show before the show start, but not actually go live. At least you'll be able to separate it.

Promos: Mostly for organizing purposes just like Shout Outs, to use for Station Promotions or Show Promotions and separating them from COM or regular Voice Tracking. I basically use an unused meta tag field such as "conductor" to include the Show where that track will only be played in and make a Smartblock for Promos on each show. Although it would be nice to just have another field with dropdown or simple field that we can link a track to a specific show. This way we don't have promotions playing where they don't belong.

Although I like the idea on the mockup, In my opinion I think it might be cleaner and easier if we keep it the way it is, but having an option on top to switch Media Types. So lets say you have a folder full with new commercials, you can simply switch the option to Commercial (COM) and they will be saved as Commercial. This is basically the way Spacial's Samcloud does it.

Screen Shot 2019-10-05 at 12 31 17 PM

paddatrapper:
@robbt how does that mockup work on mobile? I realise that LibreTime isn't really mobile friendly to begin with, but do we want to go further in that direction?

It may require a UI re-write to make LT mobile friendly, in which case a reactive UI that interacts with the API could be a long term solution and your mockup looks great!

Robbt:
Here is another mockup I made that kind of works like your suggestion. With the idea of being able to assign the track before uploading or after uploading. In addition I considered the idea of being able to select an air date on this form as well but it might be too much information on one screen.

PostUploadSelection

As I mentioned above I would like for media type to be user configurable in the settings with some good defaults. I also think that we should include some kind of documentation or tool tips that explains the thinking behind whatever media types we include by default.

codenift:
Yes, I was going to mention that we need inline editing on uploads. The second mockup is definitely the one to go with.

codenift:
Yes, the documentation should be easy to explain. I think we can provide a few tips such as being able to use it inside with SmartBlocks and also outside of LB. 

Example using API (no specific language):

```
//get media type using api/live-info

if ( mediatype == "COM" ||
      mediatype == "SID" ||
      mediatype == "JIN" ||
      mediatype == "ect.." ) {
          // Show "Commercial Break"
          // don't show any info on other media types if it's not MUS
          // get the Show's Cover instead for commercial break, ect...
} else {
         // is MUS: 
         // Artist and Track Title
         // extra stuff here related to the music track
}
```
For an app or site, I do a check on the upcoming track to see what it's type is. This way if on a commercial and the upcoming track is a commercial, I ignore updating the "Commercial Break" info or show's artwork. There's definitely a lot use with it.

codenift:
One more thing. Your idea with the date range option I think is a excellent idea and would be great to have in LB, I think for all types of tracks and not just commercial. Example, we might have some tracks that would only make sense if played on the week of Christmas or maybe a custom show intros and outros that would only play in specific holidays. I think we may need to put some thought into it to be set as priority when loading on those dates. I mean I guess we can save in the info to 'cc_files' with relation to 'cc_schedule'.

codenift:
I'll close this for now so we can focus on the artwork PR. I might've made a mistake while trying revert some commits on this one. Once artwork is merged I'm just going to redo the fork I have to start fresh and not merge anything into my master except the updates from the main LT. Since I have a backup it will be easy to do another PR.

PR #888, opened 2019-10-10
Catalina support using Vagrant

codenift:
Due to the changes in Catalina with their dedicated system volume we have to change the Vagrantfile in order to point to the install file, else it won't work. NFS sync fix for Mac users updating to the new OS.


paddatrapper:
Please just have changes to the Vagrant file in this PR. The other changes are all already in other PRs and make reviewing and merging this difficult.

codenift:
Ok, that makes more sense to me now.

PR #890, opened 2019-10-12
Moved intro and outro autoplaylist out of repeat until full loop

### LINKED ISSUE #881: Autoloading Playlist: Repeat until full does not work with Outro playlist "fill time remaining" smartblock
**Describe the bug**
Basically if you have a autoloading playlist with "repeat until full" checkbox enabled to try to fill the entire block and also have an outro autoloading playlist with a smartblock set to "fill time remaining"  enabled they will contradict each other.

Basically both are trigger but the outroplaylist fills up the remaining time happens first and then the autoloading playlist is repeated. For instance if you are using the outro playlist to fill in ambient music in the gaps between programs and an autoloading playlist as an autoDJ that should loop until it is full you can find yourself with a single instance of the playlist and then a bunch of repeating ambient tracks. And then another instance of the autoloading playlist after the fact.

**To Reproduce**
1. Go to a show and enable a autoloading playlist that will not fill the entire show and check "repeat until full"
2. Create a smartblock with the limit "time remaining in show" and repeat tracks
3. Add the smartblock to an outro playlist.
2. Go to settings -> general and set the outro playlist to the outro playlist containing the smartblock 
3. Let the show autoschedule
4. See the time remaining smartblock is inserted before the content playlist repeats

**Expected behavior**
In theory if you are repeating a autoloading playlist until full this should happen repeatedly and then the outro playlist should be scheduled. If the outro playlist contains an item that schedules "time remaining in show" it will never schedule because the autoloading playlist should reschedule until full.

The outro autoloading playlist section needs more work and should be better integrated into the software vs. sitting as a small setting under "general" because it can potentially have a huge impact for good or bad (if misconfigured) on the people.

**LibreTime version:**
Recent release
**Installation method:**
 - OS: Ubuntu via Git and ./install

Robbt:
This fixes #881
It removes the intro and outro playlist add from the "repeat until full" content loop to avoid a "time remaining in smartblock" outro from filling a multi-hour show with a repeating outro.

### DIFF
--- a/airtime_mvc/application/common/AutoPlaylistManager.php
+++ b/airtime_mvc/application/common/AutoPlaylistManager.php
@@ -39,22 +39,18 @@ class AutoPlaylistManager {
             $full = false;
             $repeatuntilfull = $playlistrepeat->getAutoPlaylistRepeat();
             $tempPercentScheduled = 0;
+            $si = new Application_Model_ShowInstance($autoplaylist->getDbId());
+            // the intro playlist should be added exactly once
+            if ($introplaylistid != null) {
+                //Logging::info('adding intro');
+                $si->addPlaylistToShowStart($introplaylistid, false);
+            }
             while(!$full) {
-                $si = new Application_Model_ShowInstance($autoplaylist->getDbId());
                 // we do not want to try to schedule an empty playlist
                 if ($playlistid != null) {
                     $si->addPlaylistToShow($playlistid, false);
                 }
                 $ps = $si->getPercentScheduled();
-                if ($introplaylistid != null) {
-                    //Logging::info('adding intro');
-                    $si->addPlaylistToShowStart($introplaylistid, false);
-                }
-                if ($outroplaylistid != null) {
-                    //Logging::info('adding outro');
-                    $si->addPlaylistToShow($outroplaylistid, false);
-                    //Logging::info("The total percent scheduled is % $ps");
-                }
                 if ($ps > 100) {
                     $full = true;
                 }
@@ -62,7 +58,7 @@ class AutoPlaylistManager {
                     break;
                 }
                 // we want to avoid an infinite loop if all of the playlists are null
-                if ($playlistid == null && $introplaylistid == null && $outroplaylistid == null) {
+                if ($playlistid == null) {
                     break;
                 }
                 // another possible issue would be if the show isn't increasing in length each loop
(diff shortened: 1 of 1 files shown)


airtime_mvc/application/common/AutoPlaylistManager.php
```diff
@@ -39,40 +39,45 @@ public static function buildAutoPlaylist() {
...
-                if ($introplaylistid != null) {
-                    //Logging::info('adding intro');
-                    $si->addPlaylistToShowStart($introplaylistid, false);
-                }
-                if ($outroplaylistid != null) {
-                    //Logging::info('adding outro');
-                    $si->addPlaylistToShow($outroplaylistid, false);
-                    //Logging::info("The total percent scheduled is % $ps");
-                }
                 if ($ps > 100) {
```
paddatrapper: Can this check not check the current duration plus the duration of the outro is less that the show duration? That way it won't run over if repeat until full is selected

Robbt: There is no easy way to determine what the effect of adding a playlist to a show will be until it happens. This is because the playlist could have a random smartblock. 
We could "disable" the outro playlist if repeat until full is enabled. There is a basic conflict between repeat until full and the outro playlist that I didn't realize until the software scheduled hours of filler tracks instead of repeating the same random selection.

It's not ideal but I think moving the intro and outro out of the loop is the quickest and most logical fix. The intro and outro playlist are so hidden in the settings that it is unlikely that they are widely in use and they aren't well documented in terms of how they are supposed to work.

I think it makes sense that the intro is added once to the start of the show and the outro once to the end vs. appending them multiple times which would have happened with the repeat until full w/o a fill time remaining in either of them.

I guess the only other potential conflict introduced by this is if there is a "limit to time remaining" smartblock in the intro it could supercede the content, whereas I had coded that intro to be inserted "after" the content playlist was repeated and now it has been moved to happen first.

We don't currently have the tools coded to "check" the contents of the playlist to see whether it contains a "limit to time remaining smartblock".

The whole notion of playlists and smartblocks is pretty confusing from an UI perspective and I'd like to revise it but I think this fix is probably the right way to prevent the scenario I highlighted in the bug.

paddatrapper: Ah right, I had forgotten about that.LGTM then

PR #892, opened 2019-10-17
Track Types Feature

codenift:
I redid the media_type PR with a couple fixes (Now a clean PR).

Here's the link If you need to review the comments from the one I closed:
https://github.com/LibreTime/libretime/pull/878


Robbt:
Ok, so I guess my main question is, should we work on refining the default media types and making them user configurable before merging this ? As I mentioned before I'm a little hesitant to include a large number of media types w/o documentation and tool tips to describe them as I don't want to increase the cognitive load on users getting started with LibreTime.

codenift:
I did put some thought into it. but for now I think the defaults should do as a lot of people who use services like SamCloud for example are used to these already, I forgot the other company that use these as well so they're pretty standard in my opinion. I'll post something with tool tips here later today. 

If you wanted something where users can create their own, My guess would be to have a small link below the drop down saying "Create New Type" that will reveal another input field. Once saved, it can then be included under a division line inside the Media Type drop down for reuse, that would appear just for that user. This might be tricky when it comes down to querying in Smartblocks if other users like the admin set one MT to a track and the user is trying to set their own to the same MT (being that every user is doing their own custom Media Type). Now if it was just made to where everyone agrees and everyone see the media type, like even if it's set where the admin decides what custom Media Type then it will be easier to do. Not saying it's not possible, but might we might need to make sure there's not going to be any confusion later on where users are trying to set one thing, and then it gets ruined by another user.

This is just a quick sample on how it can look.

Screen Shot 2019-10-18 at 9 36 41 AM

After it's saved it will just appear in the list for reuse.
Screen Shot 2019-10-18 at 9 36 31 AM



We could definitely put some more thought into this, but don't think it's something we need right away in my opinion. If there is a type that you might think to be standard and missing we can just add it to the list.

codenift:
Ok, so I have some of information with tips written already, but put this on hold for now since I might need to add more info to it. I just thought of something really cool and going to try and work on it this weekend that way it can be all together.

codenift:
I've added an option to set tracks to a media type on upload. I'm still experimenting with other features for media type but I think we could add them later on, for moving files into their own media type directory on upload and when editing tracks and we just have to make sure we get it right. The custom media type could also be on another PR. For now I think we can pass this PR so that we can at least set media types on our files.

codenift:
- Music (**MUS**) (Default on upload)
   Used for tracks containing music.

- Station ID (**SID**)
   Used for station IDs.

- Intro (**INT**) 
   Used for show intros.

- Outro (**OUT**) 
   Used for your station IDs.

- Sweeper (**SWP**) 
    Used for short segues between tracks.

- Jingle (**JIN**) 
    Used for a short song or tune to be played during commercial break. Contains one or more hooks for your radio station, show, product, service or event. 

- Promo (**PRO**)
  Show or Station promotions

- Shout Out (**SHO**) 
   Shout outs from artists for the station 

- News (**NWS**)
   Used for news reporting

- Commercial (**COM**) 
   This is used for commerical advertising.

- Interview (**ITV**)
   Used for interviews
 
- Voice Tracking (**VTR**)
   Used for general voice tracking

Used media type ID for Smart Blocks. (eg. **MUS** for Music)

Robbt:
Well we could merge this as is but I do think we should work on creating a new section in settings where stations can modify the media types their station uses. I know that these terms are pretty universal in commercial broadcasting but I think it would be better to allow different stations to modify it to fit their broadcast niche or organization. And I don't want to just add more types by default but for instance my station would need "Public Service Announcements" and "Underwriting Spots".

 (I'd also prefer possibly track type vs. media type) since they are all audio files and media type has connotations of different types of media like photo, video, audio etc.

I also think vs. requiring hand-typing the media type short code for smart blocks we could do something like a drop-down list. I haven't even tested what you have written so far, so I need to do this and since you are the one contributing the code at the moment so feel free to pass on coding any of my suggestions and we can review what you have done thus far when you are ready.

codenift:
Ok. I'll do that. I had something I was working on, but I think for now I'll add it on an admin level so they can set custom types. Also, I think we can change media type to track type. That was actually the first thing I was going for but went for the way SamCloud does it since I was already used to their way. But we can do that.

Robbt:
Thanks for considering the idea. My preference would be for admins to be able to add and remove (or at least disable) track/media types. We can go with your list as the default unless someone wants to suggest an alternative default list. The idea of custom types would be better but it would not allow stations to simplify the choices. 

This is probably too much to code but the idea of having different track type bundles would be useful. My goal once this is in in core is for us to add a filter to the library page that will just show tracks of a certain type so that DJs can more easily browse the library.

codenift:
Working on it. Hmmm, I thought I added the filter on library already, I'll make sure to add it. Right now you're able to check on column "Type" where you can see all the types in order.

Robbt:
So I finally got around to reviewing this and it appears to be working as is. 

There are a few more features that I'd like to see added.

- a drop down menu for smart block media type selection vs requiring the user to input the 3 letter shortcode.
- ability to modify the media types via settings
- add media type to columns of the library so that library can be filtered by media type
- add documentation for end-users

The question at this point is if we should merge this and then add the desired features in another PR or if @codenift plans on working on it more since there hasn't been a commit for a couple of months. I might tinker with it on my own box and see what I can accomplish in terms of meeting these desired features.

codenift:
Since I was working on other projects I didn't really have the time to get back to this, but will get back to it after New Years. I did get to include a drop down on Smart Block that mostly work, just need to get back to it. Also did work the other 3. Give me like 3 to 4 days until I push another update.

Robbt:
Sounds good. Thanks for the update.

paddatrapper:
I think the documentation is a blocker for merging this, while the other points could be added in follow-up PRs

codenift:
I will have some time to work on this today. Expect update real soon.

codenift:
Hey guys, I apologize for the delay. Time is not my friend. I was able to get some work done for this PR, just need to do some cleaning up and run a couple tests. Here's a little preview of the additions: https://youtu.be/mSU0_jBaBmk

I'm think it would make more sense to have checkboxes where you can select which roles can see it instead of the dropdown you see in the video for visibility. I'll see if I add it, but for now I just want to get this cleaned up and pushed here.

codenift:
Ok guys, I'm down to just one small issue for the Smart Block (in **smart_blockbuilder.js**). if someone wants to tinker with it if you don't want to wait a little longer, everything else is ready. Basically, all it needs is to keep the criteria modifier from changing when the type track select menu is changed. It's a bit tricky to get working properly. although at one point I did have it working, not sure what happened. I'll try to see if I find what I had in a backup, but If someone can beat me to it that would be great, if not then I'll try to get to it this week. You can compare them with the previous file and go by that.

There's definitely room for improvement when working with track types, but don't think it's necessary at the moment. We have the basics here. If we do the inline editing in the future, then I can modify so that if choosing various types, the selector changes on top from what I have now. 

We can add the folder organizing feature depending on track type later unless someone wants to work on it right now. For that, we may also have to move files around when type is changed from the track edit tab. Ill add this in my todo list.

The track type select menu on upload window can stay like that for now, I was thinking of making a custom select box later on where you can choose the types and also playlist.

If someone manage to fix that javascript issue with smart block then this is ready to get merged, unless you find something else. Test it and make any changes accordingly.

codenift:
How can I revert back two commits? I think i screwed up trying to resolve a conflict.

paddatrapper:
git reset --hard HEAD~2

git push --force origin media_type

codenift:
I just noticed that I reverted 1 commit too far, removing the Smart block commit where the file that needs attention is (**public/js/airtime/playlist/smart_blockbuilder.js**). Good thing I made a backup. I just pushed it back here. Forgot to mention, I'm not good with documentation so this is something that can be improved.

codenift:
Ok, I figured out the little javascript issue i was having, forgot to add the 'sp_criteria_value' on line 321, it should be go to go now.

Robbt:
Hey @codenift I'll try to get around to reviewing this soon. Thanks for your work on this. It's a major accomplishment and will dramatically improve the LibreTime workflow in my opinion.

Robbt:
Ok, this is great. Finally got around to testing it and it looks like everything works.

One question, what is the reason for using short codes vs. just displaying the full type name in the library ? I'm not sure we need short codes and I think that it'd probably be easier for people to see the full name of the track type displayed.

I see that you allow the choice of track type via drop down for smart blocks and under advanced search. This is great.

As far as feedback regarding the documentation I saw you put it under how-to. I think it'd be good to actually update the rest of the documentation to integrate this feature. I can work on that, for instance we might want to update the Tracks, Upload and Smart Blocks sections to include information about Track Types.

I also really like the customization functionality for custom Track Types. Deleting and disabling track types didn't break anything in the UI as far as I could tell so that is good. Tracks with the deleted or disabled type kept their type but needed to be changed.

I'm still not sure that the 12 types you included are optimal for a default choice to install. For instance the track settings page only shows 10 types, so you have to go to page 2 to see the last two types. I suppose people can customize this after installation and the choices are well labelled. So it's definitely not a blocker.

An idea for future development would be to allow users to select a package of track types during installation to match their station type. But that is unnecessary before we merge this.

codenift:
I did put some thought on it and just decided to go this route of having codes. Most radio softwares have it this way using codes instead (COM for Commercial), so not sure if they're just keeping it numbered which I also thought of... but in something like Samcloud it seems like they're using it this way since the code will appear listed but also in full name in other areas such as uploads. In their API you call it by code.

If they want to change the name, add characters, ect.. at least it won't break in an external application or website using the API. If I had it numbered, then of course we could have still achieved it and able to add different languages in the future for it without breaking anything in LT or external script, but this way it's just as easy as remembering a screen name. For some it may not be needed, but helps for others such as devs to identify it quickly. Also for users, being that you can do a short code, you're able to identify it quickly in the library list without taking up to much space in a row. If you notice other radio softwares, they have it in many areas such as the library since it's only using 3 or 4 letters..  I was actually thinking of adding a color option to it as well later on.

Robbt:
Those are good reasons to have codes ie you can change the label without messing up API calls or the like. I still think it might be preferable in the Library to display the full name vs. the code, maybe we could display both. Commercial (COM) or make it an option. I just think that a DJ browsing the library will be more likely to recognize a track by its full name and possibly confused by the short code. Keeping a code for a machine name persistent ID makes sense. 

How do you feel about changing just the library display under tracks ?

paddatrapper:
It may end up being too long as the name if listed in the library. Especially for smaller screens

codenift:
@paddatrapper Yes, I was going to say the same thing. In the radio world, djs are already used to the codes in capital letters to identify it quickly. They would welcome it. It keeps everything organized. If someone was to add a long track type name, it would take too much space in the library.

Robbt:
Hmm maybe the reverse code first then full display would be better. Example (MUS) Music. I don't think its a big deal either way, people will adapt and learn the question from a HCI perspective is what is easiest for a person to understand and they will need to learn the codes whereas the full text is more self explanatory. Also  if we are going to have just the codes we need to have a legend somewhere for them to look them up.

paddatrapper:
That would work. Could also do short in the list and long in the edit/detail view for the item

Robbt:
A tool tip/hover pop up with the full name could also work but if we did the code first and then full name if the column is too small it would still display the short code. I don't think non admins will be able to see the admin track types screen and would only get an overview in edit tracks.

codenift:
@Robbt Yes, they'll adapt to it very quickly, in Samcloud just in documentation was enough, but since we have option to use custom types then maybe add the short code next to the name found in edit tab for tracks. With a quick view on the track they would see if they need to remember. But the full name is in mostly everywhere (advanced search, edit window, smart blocks) i think the only place I kept it short was displaying it in library tracks for space saving. 

@paddatrapper yes, that's how it is right now. So when you go to edit it will show long version already selected in the dropdown box.

codenift:
I was thinking of tooltips but not sure how it would look there. Yes, right now through there and If they were to also search Music in Advanced Search, all tracks they'll notice they all have MUS on it. I'll see what I can do.

Robbt:
Tooltips could work, or even just a hover JavaScript would provide more detail for curious users. I find that users vary in their knowledge and understanding. Many users are unaware that advanced search exists because they don't notice the drop down arrow, as a separate issue maybe we could make it more visible.

codenift:
Yes, I agree with the advanced search. I even wrote a note to try and improve how it's design one day.  

Yes, i just don't want to make it where the short code are not shown, since it's just a way a lot of people in radio are used to, and this being a radio application it's good to have, i think it could be even a standard in broadcast since most use COM for Commercial and MUS for music for example. Once they view it for a couple days, they'll get accustomed to it. I'll see what I can do with tool tips, just don't know if it's going to place nice with touch devices, I might have another solution that will work.

codenift:
Ok, so I added the tooltip. We can probably tweak it up to give it a different look if we want to, doesn't have to be right now. I did set it up the way I wanted and made it into a box just incase I add the option to add colors to the types later on, plus they can click on it if using touch devices. Let me know if it needs anything else.

Robbt:
The tooltip works great. I think that this is ready to merge. We can modify the existing documentation in a separate PR since this one at least includes a basic how-to.

 Also on a side note I'll also be talking about LibreTime this weekend at LibrePlanet online - https://libreplanet.org/2020/program/ - only had my talk accepted about a week ago and since then the conference went virtual.

paddatrapper:
Sounds good to me.

Congratulations! I'm looking forward to the talk. Do you know what timezone the schedule times are in?

Robbt:
The schedules are in EST. Recordings should be available afterwards as well.

codenift:
Cool. I'll tune in.

Robbt:
Ok, I'm going to merge this, thanks for the contrib @codenift

kmahelona:
Hey, so this changes the database right? Is there a proper way to upgrade with this? Are database changes handled with migrations automatically?

Robbt:
With regards to the database Zend should handle the upgrades automatically if the update code is working and we try to test to make sure that it works. Each time a database update is merged we increment the versioning on any existing PRs.

I don't think we should hardcode a default type to watched folders. This might be a situation where we should give the user the option of say assigning a type to each watched folder and then during the import process that type is assigned to the tracks. I could see this being useful in a lot of ways actually.

codenift:
I'll be making a couple more updates regarding types. We still need to do something for inline editing multiple tracks.  I'll make sure to hide the type button from the list if value is null, no real need to show it... this way it's not shown in playlists, smart blocks as well. I did notice we need to find a way to separate the way we view columns in playlists and smart blocks from tracks since they don't use the same fields Tracks is using. So being that it wasn't just the Types I didn't touch them. Should be in a separate PR. 

Making an option to assign a type to each watched folder sounds like a good idea.

Issue #434, opened 2018-02-03
Admin request: Lets get some help from probot


hairmare:
I'd like to activate some [probot](https://probot.github.io/) apps on this repo.

As a first step let's activate the following apps to get started.

## [WIP](https://probot.github.io/apps/wip/)
Blocks merge of PRs with WIP in title.

![wip](https://raw.githubusercontent.com/gr2m/wip-bot/master/assets/wip.gif)

## [Lock Threads](https://probot.github.io/apps/lock/)
Locks issues and prs for further edits a year after they have been closed.

![Lock Threads](https://raw.githubusercontent.com/dessant/lock-threads/master/assets/screenshot.png)

## [Invite Contributors](https://probot.github.io/apps/invite-contributors/)
Automatically invites people that got their first pull request merged to the Contributors group so they can tag issues and self assign issues.

Since the last one needs configuration I'll do a PR on it anyway, so we can discuss it's merits over there.

----

There are [more apps](https://probot.github.io/apps/) available on the probot page. If everyone is on board, I see us adding [stale](https://probot.github.io/apps/stale/), [support](https://probot.github.io/apps/support/) and [Triage New Issues](https://probot.github.io/apps/triage-new-issues/) when we are ready for them asap.

Robbt:
This all looks good to me.

hairmare:
WIP and Lock Threads activated, I'll close this when I do the PR for Invite Contributors.

hairmare:
invite-contributors PR created: #436

PR #896, opened 2019-10-18
Config for lock bot

hairmare:
I activated the [lock probot](https://probot.github.io/apps/lock/) after our discussion in #434.

Before the the bot every had any work to do it was refactored to mandate needing a config file in `.github/lock.yml`.

This PR finally adds such a configuration and configures it to post a comment that points people finding such issues our discourse and chat.

### DIFF
--- a/.github/lock.yml
+++ b/.github/lock.yml
@@ -1,0 +1,30 @@
+# Configuration for Lock Threads - https://github.com/dessant/lock-threads
+
+# Number of days of inactivity before a closed issue or pull request is locked
+daysUntilLock: 365
+
+# Skip issues and pull requests created before a given timestamp. Timestamp must
+# follow ISO 8601 (`YYYY-MM-DD`). Set to `false` to disable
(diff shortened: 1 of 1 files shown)


.github/lock.yml
```diff
@@ -0,0 +1,30 @@
...
+# Label to add before locking, such as `outdated`. Set to `false` to disable
+lockLabel: false
+
+# Comment to post before locking. Set to `false` to disable
+lockComment: >
+  This thread has been automatically locked since there has not been
+  any recent activity after it was closed. Please open a new issue for
+  related bugs.
+
+  Please chat us up on [discourse](https://discourse.libretime.org/) or
```
paddatrapper: "please chat to us on" - otherwise sounds like we're asking people to flirt with us :)

hairmare: I wasn't aware of that connotation, fixed and rebased into 09d7397.

PR #902, opened 2019-10-21
Add OpenAPI doc of the current API

paddatrapper:
This is an attempt to document the current API. The functions I have found to be broken are marked as such in the description of the function

### DIFF
--- a/docs/api/openapi.yaml
+++ b/docs/api/openapi.yaml
@@ -1,0 +1,1343 @@
+---
+openapi: "3.0.0"
+info:
+  title: LibreTime API overview
+  version: 1.1
+paths:
+  /live-info:
+    get:
+      summary: Retrieve the currently playing show as well as upcoming shows
+      parameters:
+        - name: api_key
+          in: path
+          description: The API key to use for authentication
+          required: false
+        - name: type
+          in: path
+          description: |-
+            endofday retrieves the info for shows up until the end of the day,
+            while interval will return shows in the next 48 hours
+          schema:
+            enum:
+              - 'endofday'
+              - 'interval'
+            default: 'interval'
+          required: false
+        - name: limit
+          in: path
+          description: the number of shows to retrieve
+          schema:
+            type: integer
+            default: 5
+          required: false
+      responses:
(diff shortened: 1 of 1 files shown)


docs/api/openapi.yaml
```diff
@@ -0,0 +1,963 @@
...
+                "nextwednesday": [],
+                "nextthursday": [],
+                "nextfriday": [],
+                "nextsaturday": [],
+                "nextsunday": [],
+                "AIRTIME_API_VERSION": "1.1"
+              }
+  /station-metadata:
+    get:
+      summary: BROKEN: Retrieve the schedule for the week
```
jooola: This string need to be quoted, it breaks when trying to parse it with swagger-ui or redoc.
Using `:` in the string can lead to weird behavior.

There might be some other strings that need to be fixed.

paddatrapper: Thanks. Fixed

paddatrapper:
Seems many API endpoints are broken due to setting `$this->view->msg` presumably as the response, which doesn't work. Just returns a 200 response with no content.

This is fixed by replacing that with `$this->_helper->json->sendJson(...)`. I'll open a separate PR for that

paddatrapper:
This should now reflect the current state of the API. There are places where I am not sure what the current behaviour is, but I have tried to make them clear

Robbt:
I'm going to commit this now since it doesn't actually affect the codebase yet and it should be useful.

paddatrapper:
Well it is the current API, not a potential new one, so it reflects the current codebase. New API doc will follow once $dayjob settles down in the new year

Robbt:
Cool that makes sense, should we also include the REST modules under airtime_mvc/applications/modules/rest/controllers - this is another API style functionality that it would be good to document I think.

paddatrapper:
We should. I didn't know they existed

Robbt:
Yeah it's part of the fragmented codebase that makes working on LibreTime so challenging. Merging these two different API points conceptually and coming up with a logical path forward would be really rewarding I suspect.

PR #907, opened 2019-11-04
PypoMessageHandler reliable RabbitMQ consumption

### LINKED ISSUE #906: MVC/Liquidsoap Synchronization : Hardening Exception Handling?
**Describe the bug**
Desynchronization between the MVC and Liquidsoap playout.
This is probably a duplicate of #699 and might be related to #232 and #125 or even #307
Opening my own issue so I can make a whole of it and work towards a solution.

**To Reproduce**
For instance, adding a track after the current one, and the removing the latter, should result in changing the currently playing track ("take over"). Instead, ON AIR turns off, and cannot catch up until pypofetch times out. 

**LibreTime version:**
I'm running *almost bleeding-edge LibreTime (last merge the 7th Oct - 50ca96156ef1962ea79ef1172ca81084529ff7a4)
Liquidsoap is 1.3.3 and I have applied the patch from https://github.com/radiorabe/libretime/tree/feature/liquidsoap-1.3-for-3.0.0-alpha.7

**Installation method:**
Running on a custom docker container, based on Ubuntu Xenial inspired by this thread #624

**Additional context**

I can confirm the work-around is to restart *airtime-playout*, but the bug will reappear after some hours.

After staring at the logs my impression is *pypopush* dies at some point, and that causes this desynchronization, as the only schedule update is obtained when pypofetch times out.
This could explain why the issue overcomes after a period of silence.

On the other hand, I suspect* pypopush* dies just because of a poor exception handling.
Indeed, RabbitMQ regularly throws broken pipe errors, although it reconnects immediately after.
Other connections have been seen to 'misbehave' similarly, such as telnet to liquidsoap.

And this could be happening not only to _pypopush_. 
Hardening exception handling in pypo could be the real issue here?

numericalatina:
This fixes #906 

Finally I found out the main issue causing desynchronization between the MVC and Liquidosap was that the RabbitMQ got stale after a while (less than an hour), then schedules were only updated on PypoFetch time out.

The gotcha here is we were using a SimpleQueue, which's 'python-like' *get*  method apparently doesn't handle recovery well. Kombu Consumers are the way to go instead.

### DIFF
--- a/python_apps/pypo/pypo/pypomessagehandler.py
+++ b/python_apps/pypo/pypo/pypomessagehandler.py
@@ -7,14 +7,31 @@ import sys
 from threading import Thread
 import time
 # For RabbitMQ
-from kombu.connection import BrokerConnection
+from kombu.connection import Connection
 from kombu.messaging import Exchange, Queue
 from kombu.simple import SimpleQueue
 from amqp.exceptions import AMQPError
 import json
 
+from kombu.mixins import ConsumerMixin
+
 logging.captureWarnings(True)
 
+
+class RabbitConsumer(ConsumerMixin):
+    def __init__(self, connection, queues, handler):
+        self.connection = connection
+        self.queues = queues
+        self.handler = handler
+
+    def get_consumers(self, Consumer, channel):
+        return [
+            Consumer(self.queues, callbacks=[self.on_message], accept=['text/plain']),
+        ]
+
+    def on_message(self, body, message):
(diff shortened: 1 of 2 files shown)


numericalatina:
It might be useful to tell that my RabbitMQ service runs on a separate Docker container.
Therefore, we cannot rely on the connection (through an overlay network) as much as when they're on the same host.

paddatrapper:
Thanks! This works for me to

Issue #921, opened 2019-11-24
UI broken after trying to save Live Broadcasting settings


matias-tecnosoul:
In a fresh install via git done today I finally made everything work (after trying a couple of times). But **after making some changes in the Live Streaming section I cannot save changes again in the Streams section and lost the Scheduler from the Dashboard**.
However the show and streams are still working, I can still hear the output.

 When I got to the "streams" settings I went straight to "Output Streams" part since I already had some strange behaviour in a previous install. 
First I tried with the "default streaming" with the local Icecast working ok. Then I added an external output stream located in another server and both streams worked great.

But then I tried to set the "Master Source" in "Live Broadcasting" section and when tried to save it just stayed there loosing the "Connected to the streaming server" message and changing to "Getting information from the server". 
**Since then I cannot save changes again in the Streams section and lost the Scheduler from the Dashboard, subelements (tracks, playlists etc) and Calendar.**

**Expected behavior**
Change the login data for the master input

**LibreTime version:**

55d5b40

**Installation method:**
 - OS: Ubuntu
 - OS Version 16.04.5 LTS (Xenial Xerus)] 
 - Method:  `./install` script
 - Details: git clone https://github.com/libretime/libretime.git

**Screenshots**
![libretime-broken-ui1](https://user-images.githubusercontent.com/19720454/69496896-e447c680-0eb5-11ea-823b-e192b5f22074.png)
![libretime-broken-ui2](https://user-images.githubusercontent.com/19720454/69496904-f32e7900-0eb5-11ea-84a2-91485d8bb510.png)
![libretime-broken-ui3](https://user-images.githubusercontent.com/19720454/69496907-f6c20000-0eb5-11ea-9cfd-326fba9a8349.png)

**Client (please complete the following information if applicable):**
 - OS: Arch Linux
 - Browser: chromium
 - Version 78.0.3904.97

**Additional context**
Fresh Ubuntu install. Already have an older Libretime install working on another server like this one.

matias-tecnosoul:
Some more info:
After a couple of hours I logged in again and the Dashboard was there and the streams showed up again  as "connected to the server" :)
but there weren't any shows scheduled...
I first went to try and set up the "live broacasting" master input and it worked fine! the settings showed a "succes" message and the streams were still "connected"... Still without any scheduled shows.
So I added a show a few minutes ahead and it seems that when it started the issue was back again... No dashboard, no calendar, no users listing, and the streams back on "Getting information from the server..."

[...]

Issue #919, opened 2019-11-23
MVC breaks with input streams (ports 8001 & 8002)


numericalatina:
**Describe the bug**
When live streaming to LibreTime on ports 8001 or 8002, the MVC interface would break (500)
I've found this is because of this line introduced recently on ScheduleController L305

**To Reproduce**
streaming to LibreTime on ports 8001 or 8002

**Expected behavior**
MVC working (showing ON AIR or the orange string)

**LibreTime version:**
latest

I'm opening a pull request with the fix immediately

numericalatina:
The above commit fixes the issue when Streaming to Input ports, however, it produces an other bug when Scheduling a Webstream: once the webstream goes on air, the MVC interface freezes.

I can confirm the error disappears when reverting this commit. So we're in the situation 

So I'm asking to please re-open this issue. We need to find a fix that solves both bugs, instead of one or the other.

![LT_blank](https://user-images.githubusercontent.com/39980286/69575939-2b01f300-0f80-11ea-84e9-087d6e467b88.png)

Note how this could be related to #921 as the screens look similar

paddatrapper:
Damn... I wonder what is causing that. @numericalatina please open a new issue describing the problem and reference this issue and the PR.

PR #925, opened 2019-12-04
Fix Dashboard when current instance has no metadata

### LINKED ISSUE #924: Dashboard fix when current audio has no metadata
**Describe the bug**
when Scheduling a Webstream: once the webstream goes on air, the dashboard freezes.

**To Reproduce**
1. Schedule a webstream

**Expected behavior**
update the playbar

**Screenshots**
dashboard

**Additional context**
Check https://github.com/LibreTime/libretime/issues/919

numericalatina:
The issue in the dashboard was when scheduling elements which do not have metadata (for instance, a webstream, or a live input stream). 

This basic not null check was missing in javascript .

fixes #924

may be related to #919 and #921


paddatrapper:
Thanks for the fix!

PR #930, opened 2019-12-28
fix rasbian debian 10 install

### LINKED ISSUE #929: “Raspbian GNU/Linux 10 (buster)” is not supported with “raspbian-10”
It seems that is not possible to get the install working on Raspberry Pi's OS Raspbian 10 (Buster)
Im using a Raspberry Pi 3b v1.2

This is probably easily fixable by changing the install script to accept raspbian-10 ?

Please assist

Robbt:
This should fix #929 but I haven't tested it.

### DIFF
--- a/install
+++ b/install
@@ -633,6 +633,15 @@ case "${dist}-${code}" in
 	is_debian_stretch=true
 	;;
     #End of fix
+    #Fix for Raspbian 10 (buster)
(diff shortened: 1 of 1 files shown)


Robbt:
Just remembered while testing debian-buster that we need to implement the liquidsoap 1.3 fix in the installer for this to work for anyone.

PR #931, opened 2019-12-31
add liquidsoap 1.3 support while retaining legacy script for liquidsoap 1.1.1 compatibility

### LINKED ISSUE #192: Problem: We want to upgrade to Liquidsoap >=1.3.x
## Issue
The thing is, [liquidsoap 1.3.0](https://github.com/savonet/liquidsoap/releases/tag/1.3.0) was recently released and it contains some fixes that are highly relevant to us. There were changes to liquidsoap that need adopting our code to in a way that breaks with the liquidsoap <1.3.0 (like the 1.1.1 we currently have on Debian). I did the required changes on https://github.com/LibreTime/libretime/compare/master...radiorabe:feature/liquidsoap-1.3.0 and they are rather minimal but break badly with older liquidsoap releases.

See [below](#user-content-workaround) for how to patch your install to use 1.3.x now.

## ~Proposed~Available Solutions
The liquidsoap package that was previously maintained by @toots on Debian but has been orphaned in favor of opam so we need to consider a couple of options going forward.
* Take over the packaging torch and bring the packages up to date (I think this means removing the modularity and making one large binary)
* Try submitting an NMU (non maintainer update) against the packages in the hopes that it can get bumped and stay unmaintained for a while (same as option 1 but doable without a Debian contributor on board)
* Offer our own packages based on the NMU or even completely separate from Debian and Ubuntu (ie. built on build.opensuse.org)
* Change our docs and the install script to install liquidsoap from opam
* Use packages from deb.liquidsoap.info (possibly `liquidsoap>=1.4.0`)

## Details for opam Solution
Since the last one in this list [is recommended in the liquidsoap docs](http://liquidsoap.fm/download.html) I'm documenting what is needed to get it working on the Ubuntu Trusty vagrant box (I started running these as root, the user to use is mostly noted where a switch is needed).

```bash
# remove distro liquidsoap
apt-get remove liquidsoap
# prepare system for opam install
add-apt-repository ppa:avsm/ppa
apt-get update

# install system packages like opam (the make tools are for the 1.3.0+scm install below)
apt-get install ocaml ocaml-native-compilers camlp4-extra opam autotools-dev automake
mkdir /usr/local/opam
chown liquidsoap:liquidsoap /usr/local/opam /usr/share/liquidsoap/

# we need to switch to the liquidsoap, some things do not like being installed as root
usermod -s /bin/bash liquidsoap
# switch to now active liquidsoap user
sudo --login -u liquidsoap
# run as liquidsoap user
opam init --root=/usr/local/opam

# To setup the opam/ocaml you initialized in the current shell, you need to run:
# Until you install a profile you need to do this each time you want liquidsoap on the $PATH
eval `opam config env --root=/usr/local/opam`

# I installed the deps as root before re-installing as user liquidsoap
# run this as root after running an additional eval `opam config env --root=/usr/local/opam`
opam depext alsa cry fdkaac lame liquidsoap mad opus taglib vorbis

# install liquidsoap and deps we need (run this as user liquidsoap again)
# run the git instruction instead of this for now (or re-install from git after pinning below)
opam install alsa cry fdkaac lame liquidsoap mad opus taglib vorbis    

# run this as root to make liquidsoap your default on the whole system (extremely hacky)
echo "eval \`opam config env --root=/usr/local/opam\`" > /etc/profile.d/liquidsoap-opam.sh
ln -s /usr/local/opam/system/bin/liquidsoap /usr/bin/liquidsoap
```

~~This gives us liquidsoap 1.2.1 and will also work for 1.3.0 once it is available from opam.~~ The above installs the current 1.3.x version of liquidsoap. Switching to the git install that might have additional patches works as follows.

```bash
# as long as opam does not have 1.3.0 we switch back to the liquidsoap user to install from git:
opam install alsa cry fdkaac lame mad opus taglib vorbis
sudo --login -u liquidsoap
git clone https://github.com/savonet/liquidsoap.git
cd liquidsoap
git submodule init
git submodule update
opam pin add liquidsoap .
```

Having this the vagrant boxes for Ubuntu and Debian might be an option as they are geared toward developers anyway. I'm not convinced that it's something we can support for users though. CentOS is already on 1.3.0. I believe I'm the only user of LibreTime on CentOS at the moment (please correct me if I'm wrong) so I'll just live with applying `feature/liquidsoap-1.3.0` manually or during packaging until we can figure out how to best get the branch merged. 

I think most of this has previously been discussed (ie. in the debian packaging context) and this issue was created to have a tracker issue for everything regarding the liquidsoap bump.
<a name=workaround>

## Workarounds 
Should you want to test liquidsoap `1.3.1` you can apply this patch manually. Depending on how you got your copy of LibreTime there are different ways to do so.

You need to apply the patch before running the installer (or you need to rerun it after applying the patch).
### tarball install
An up to date patch is available through GitHub and can be applied to an unpacked tarball as follows.
```
cd libretime-3.0.0-alpha.2/
curl -L https://github.com/LibreTime/libretime/compare/master...radiorabe:feature/liquidsoap-1.3.0.patch | patch -p1
```
#### For LibreTime >=3.0.0-alpha.7

```
cd libretime-3.0.0-alpha.7/
curl -L https://github.com/LibreTime/libretime/compare/master...radiorabe:feature/liquidsoap-1.3-for-3.0.0-alpha.7.patch | patch -p1
```
### git working copy
To merge the liquidsoap-1.3.0 branch to the local working copy you can run the following. You should probably create a new branch based on `master` before you do this. If you merge to your local `master` your working copy might need some work during upgrades going forward.
```
cd libretime/
git remote add radiorabe https://github.com/radiorabe/libretime.git
git pull radiorabe feature/liquidsoap-1.3.0
# or for LibreTime >=7.0.0-alpha.7
git pull radiorabe feature/liquidsoap-1.3-for-3.0.0-alpha.7
```

Robbt:
This fixes #192 and pulls the fixes @hairmare added for liquidsoap 1.3 compatibility while creating a legacy version of the scripts that will be used if the user has liquidsoap 1.1.1 (ie if they are running any Ubuntu version before 19.10 since 14.04 and probably Debian stretch). So this should allow us to start supporting Liquidsoap 1.3 immediately without breaking thing for people using Ubuntu. If we decide to quit supporting 16.04 & 18.04 in the future we can easily remove this.

I kept the changes in liquidsoap_auth.py but I haven't been able to confirm that this doesn't cause an issue with connecting on a older system as I couldn't get my local test system to connect via the show or master mount and I'm not sure if this is an issue caused by this script or is unrelated at this point like possibly an issue such as #921

### DIFF
--- a/python_apps/pypo/liquidsoap/__main__.py
+++ b/python_apps/pypo/liquidsoap/__main__.py
@@ -5,6 +5,7 @@ import argparse
 import os
 import generate_liquidsoap_cfg
 import logging
+import subprocess
 
 PYPO_HOME = '/var/tmp/airtime/pypo/'
 
@@ -21,8 +22,12 @@ def run():
         logging.basicConfig(level=getattr(logging, 'DEBUG', None))
     
     generate_liquidsoap_cfg.run()
-    script_path = os.path.join(os.path.dirname(__file__), 'ls_script.liq')
-    
+    ''' check liquidsoap version if less than 1.3 use legacy liquidsoap script '''
+    liquidsoap_version=subprocess.check_output("liquidsoap --version", shell=True)
+    if "1.1.1" not in liquidsoap_version:
+        script_path = os.path.join(os.path.dirname(__file__), 'ls_script.liq')
+    else:
+        script_path = os.path.join(os.path.dirname(__file__), 'ls_script_legacy.liq')
     if args.debug:
         os.execl('/usr/bin/liquidsoap', 'airtime-liquidsoap', script_path, '--verbose', '-f', '--debug')
     else:
--- a/python_apps/pypo/liquidsoap/liquidsoap_auth.py
+++ b/python_apps/pypo/liquidsoap/liquidsoap_auth.py
@@ -15,7 +15,9 @@ elif dj_type == '--dj':
 
 response = api_clients.check_live_stream_auth(username, password, source_type)
 
-if 'msg' in response:
+if 'msg' in response and response['msg'] == True:
     print response['msg']
+    sys.exit(0)
 else:
     print False
+    sys.exit(1)
--- a/python_apps/pypo/liquidsoap/ls_lib_legacy.liq
+++ b/python_apps/pypo/liquidsoap/ls_lib_legacy.liq
@@ -266,7 +266,7 @@ def input.http_restart(~id,~initial_url="http://dummy/url")
   source = audio_to_stereo(input.http(buffer=5.,max=15.,id=id,autostart=false,initial_url))
 
   def stopped()
-    "stopped" == list.hd(server.execute("#{id}.status"), default="")
+    "stopped" == list.hd(server.execute("#{id}.status"))
   end
 
   server.register(namespace=id,
@@ -321,7 +321,7 @@ def cross_http(~debug=true,~http_input_id,source)
   cross_d = 3.
 
   def crosser(a,b)
-    url = list.hd(server.execute('#{id}.url'), default="")
+    url = list.hd(server.execute('#{id}.url'))
     status = list.hd(server.execute('#{id}.status'))
     on_m([("source_url",url)])
     if debug then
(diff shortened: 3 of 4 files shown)


paddatrapper:
I've been trying to test this, but my vagrant setup is playing tricks with me. Hopefully I'll be able to test properly this week

paddatrapper:
I can connect successfully on a Debian Stretch vagrant box

Robbt:
Cool, well then in theory this should be ready to go.

PR #933, opened 2020-01-02
change default icecast password automatically during install

### LINKED ISSUE #86: Problem: Add dialog to prompt user to change the default IceCast password upon install
Since many users will be just rolling this out without much expertise, we should at least try to encourage secure installations.

Taking a queue from the how the setup dialog prompts the user to change their rabbitmq password, we should also include something that asks them to change their icecast password or perhaps handle this via the web interface and then saves it automatically. 
Or perhaps it can autogenerate a secure password and seed it into the web interface/icecast config.

Ideally the rabbitmq password prompt would be handled better in the same way as well since most people will likely ignore the prompt and just use the default password thus creating a potential vulnerability.

Robbt:
This is my almost working attempt to update the default icecast password.
It fixes #86 by automagically generating a password for icecast.

I am running into an issue where the prepared PDO SQL statement to insert the password into the database doesn't appear to actually run and so there is a bug somewhere in there. 

Once I can figure that out it just needs more testing but should be ready to rock. I'm pushing it right now in case anyone wants to review it and point out what I'm doing wrong, otherwise I'll come back to it later and figure it out.

### DIFF
--- a/airtime_mvc/public/setup/database-setup.php
+++ b/airtime_mvc/public/setup/database-setup.php
@@ -185,7 +185,6 @@ class DatabaseSetup extends Setup {
             throw new AirtimeDatabaseException("The Icecast Password file was not accessible", array());
        };
        $icecast_pass = file_get_contents(LIBRETIME_CONF_DIR . '/icecast_pass', true);
-       error_log($icecast_pass);
        $statement =  self::$dbh->prepare("UPDATE cc_stream_setting SET value = :icecastpass WHERE keyname = 's1_pass'");
        $statement->bindValue(':icecastpass', $icecast_pass, PDO::PARAM_STR);
        try {
@@ -193,7 +192,6 @@ class DatabaseSetup extends Setup {
            }
        catch (PDOException $ex) {
            print "Error!: " . $ex->getMessage() . "<br />";
-           die();
            }
        $statement =  self::$dbh->prepare("UPDATE cc_stream_setting SET value = :icecastpass WHERE keyname = 's1_admin_pass'");
        $statement->bindValue(':icecastpass', $icecast_pass, PDO::PARAM_STR);
@@ -202,7 +200,6 @@ class DatabaseSetup extends Setup {
            }
        catch (PDOException $ex) {
            print "Error!: " . $ex->getMessage() . "<br />";
-           die();
            }
        $statement =  self::$dbh->prepare("UPDATE cc_stream_setting SET value = :icecastpass WHERE keyname = 's2_pass'");
        $statement->bindValue(':icecastpass', $icecast_pass, PDO::PARAM_STR);
@@ -211,7 +208,6 @@ class DatabaseSetup extends Setup {
            }
        catch (PDOException $ex) {
            print "Error!: " . $ex->getMessage() . "<br />";
-           die();
            }
        $statement =  self::$dbh->prepare("UPDATE cc_stream_setting SET value = :icecastpass WHERE keyname = 's2_admin_pass'");
        $statement->bindValue(':icecastpass', $icecast_pass, PDO::PARAM_STR);
@@ -220,7 +216,6 @@ class DatabaseSetup extends Setup {
            }
        catch (PDOException $ex) {
            print "Error!: " . $ex->getMessage() . "<br />";
-           die();
            }
 
        $statement =  self::$dbh->prepare("UPDATE cc_stream_setting SET value = :icecastpass WHERE keyname = 's3_pass'");
@@ -230,7 +225,6 @@ class DatabaseSetup extends Setup {
            }
        catch (PDOException $ex) {
            print "Error!: " . $ex->getMessage() . "<br />";
-           die();
            }
        $statement =  self::$dbh->prepare("UPDATE cc_stream_setting SET value = :icecastpass WHERE keyname = 's3_admin_pass'");
        $statement->bindValue(':icecastpass', $icecast_pass, PDO::PARAM_STR);
@@ -239,7 +233,6 @@ class DatabaseSetup extends Setup {
            }
        catch (PDOException $ex) {
            print "Error!: " . $ex->getMessage() . "<br />";
-           die();
            }
        $statement =  self::$dbh->prepare("UPDATE cc_stream_setting SET value = :icecastpass WHERE keyname = 's1_admin_pass'");
        $statement->bindValue(':icecastpass', $icecast_pass, PDO::PARAM_STR);
@@ -248,7 +241,6 @@ class DatabaseSetup extends Setup {
            }
        catch (PDOException $ex) {
            print "Error!: " . $ex->getMessage() . "<br />";
-           die();
            }
        $statement =  self::$dbh->prepare("INSERT INTO cc_pref (keystr, valstr) VALUES ('default_icecast_password', :icecastpass )");
        $statement->bindValue(':icecastpass', $icecast_pass, PDO::PARAM_STR);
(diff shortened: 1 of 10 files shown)


Robbt:
So I managed to get the database statements working and the database is created but the webpage is getting stuck and not automatically going to the rabbitmq setup screen. 
Also I noticed that when I first went to settings->streams the default stream did not connect so evidently it isn't pulling the new password. When I switch to custom stream and then back to manual stream it connects. So my suspicion is that it is perhaps grabbing the default icecast pass from the config file even though I have added the appropriate variable to cc_pref that should override that one.
It requires further trouble shooting on both of these issues before it is a fix.

Summary of remaining todos
- [ ] Fix install website hang
- [ ] Fix default stream authentication error upon startup

Robbt:
Ok, so the issue appears to be that the default /etc/airtime/liquidsoap.cfg that gets generated has an extra carriage return in the password the file has this line ```
s1_pass = "BpAAUJCpLrqP
"
```
 and then when I save the stream settings and recreate it it is fixed and doesn't have this extra carriage return.

**Comment by Robbt** [comment]
Ok, just realized php was adding a new line when it read the temporary icecast password file. ```/etc/airtime/icecast_pass``` 

Because systemd gives php its own special sandboxed temp directory I had to write the icecast_pass to /etc/airtime - but it might be good to delete it at some point in the setup but doing so at any step would make it impossible to run through the setup again successfully. 

So while this is an improvement in security over the previous setup and should be mergeable at this point. It might be worth it at some point to create another setup screen specifically for icecast that allows them to enter a custom password and that pulls the password from the file.

File: airtime_mvc/public/setup/database-setup.php [hunk 1]
```diff
@@ -175,5 +176,83 @@ private function checkDatabaseEncoding() {
...
+    * @throws AirtimeDatabaseException
+    */
+    private function updateIcecastPassword() {
+       if (!file_exists(LIBRETIME_CONF_DIR . '/icecast_pass')) {
+            throw new AirtimeDatabaseException("The Icecast Password file was not accessible", array());
+       };
+       $icecast_pass_txt = file(LIBRETIME_CONF_DIR . '/icecast_pass');
+       $icecast_pass = $icecast_pass_txt[0];
+       $icecast_pass = str_replace(PHP_EOL, '', $icecast_pass);
+       error_log($icecast_pass);
```
paddatrapper: Do we want to log the password?

Robbt: Good catch, yeah that was just left there by my last debugging process and so there is no real point. The question of whether leaving the icecast_pass in plain text under /etc/airtime is the other concern but considering I also made it visible via the settings -> stream settings page I wouldn't consider it a huge security risk, especially considering we have been shipping alphas with the default password. I'll push a fix to remove the error_log line.

paddatrapper:
Just tested it and looks good! The only think I noticed is that the password under **Additional Settings** still shows as `hackme`, we should probably replace that with the actual value

Robbt:
Weird on my instance, they have been showing the generated password. Also I just realized that I don't know how to add packages to the centos version of the app since this version did require the CLI xml processer to change the password and I don't see a package list for centos.

paddatrapper:
This was on a debian-buster vagrant image. 

There are a list of packages to install in `installer/vagrant/centos.sh`, but according to the installer, Centos dependency installation is not supported - https://github.com/LibreTime/libretime/blob/master/install#L712

Robbt:
So I tested it under debian-buster with vagrant and I saw the updated password. Did you do it from a completely new instance ie ```vagrant destroy debian-buster``` and ```vagrant up debian-buster``` - I am adding the xmlstarlet file to the centos vagrant package list and I'm going to test it on our centos image just to see.

paddatrapper:
Good point, I think it was an existing image. Does liquidsoap pull the config from the DB or the config file? I.e. does the 'hackme' showing in the stream settings when I run this on an existing box mean that the DB is not updated? If so, then there will be an inconsistency between icecast always changing the password on running `./install` and the DB not being updated to reflect the change

Robbt:
Yeah, this could be an issue if people are running ```install``` to update their instances, which is pretty much or only existing easy upgrade path.

paddatrapper:
We don't create `airtime.conf` on install the first time, so maybe only update the icecast password in `install` if `airtime.conf` does not exist

paddatrapper:
It shows correctly in a  clean install

Robbt:
I'm working to only change the icecast password if /etc/airtime/airtime.conf doesn't work and getting it to work on centos (different icecast.xml location). Once I push the next commit it should be ready for final merge. I'll spin up and test on another ubuntu box just to make sure it is hitting all 3 targets.

Robbt:
I think this works but I noticed what I think is a bug in our install script that is unrelated but was causing errors in verifying that my script wouldn't change the password upon install. 
```    OLD_CONF=$(grep "media-monitor" /etc/airtime/airtime.conf)``` - doesn't provide the intended result because we will always assume that an existing conf is an upgrade because the [ ] are brackets in grep and thus they will match if any of the letters match. 

So when I take that out and run install it doesn't assume it is an upgrade and nothing happens. So I think rerunning install to faciliate upgrading is actually an unintended feature of this bug. But it hangs trying to stop media-monitor when I run it. So yeah, I think I should open a bug for that and perhaps add a --upgrade flag to ./install that will facilitate upgrading w/o assuming the user had an instance of airtime or always assuming it is an upgrade when they run install.

**Comment by paddatrapper** [comment]
I'm going to merge this and open an issue about the `--upgrade` stuff then

**Comment by paddatrapper** [comment]
This still causes issues when reprovisioning a vagrant instance:

```
$ vagrant destroy debian-buster
$ vagrant up debian-buster
$ vagrant provision debian-buster
```
Then go to the stream settings and it reports a 401 error: `Can not connect to the streaming server 401, Authentication Required (HTTP/1.0)`

Robbt:
Hmm provisioning should happen automatically when you run vagrant up debian-buster so perhaps this is happening from it being ran twice ? 

I never really reprovision an instance, and I'm not sure what exactly gets ran just on the reprovision, does this run the install script again ? I ran into issues trying to run ./install inside of vagrant regarding it trying to reinstall everything and getting hung up but thought this was related to the old instance detection not working. 

I guess the next thing to do is figure out what exactly is happening, I suspect that the password is getting changes and the database doesn't set the new instance. 

What is the intended purpose of running provision again ? Should it basically just install all of the packages again or should it wipe the database etc ?

paddatrapper:
I use it as it is faster than re-creating over a slow internet connection. From what I can gather it runs four scripts:

1. `installer/vagrant/debian.sh` or `installer/vagrant/centos.sh`
1. `cd vagrant; ./install --force --in-place --verbose --postgres --apache --icecast --web-port=8080`
1. `docs/scripts/install.sh`
1. `docs/scripts/serve.sh`

I wonder if the `--force` is causing the issue

Robbt:
So in theory it shouldn't be changing the icecast_pass because /etc/airtime/airtime.conf should already exist...but what happens is the install script changes airtime.conf to airtime.conf.tmp because it things it is an old_install because of #940 

When I try to run vagrant provision I get hung up on 
    ubuntu-xenial: Failed to stop airtime-media-monitor.service: Unit airtime-media-monitor.service not loaded.

But I guess before when I was testing this I got impatient vs. allowing it to timeout and run through again.

Ok I figured it out...

Basically I only copied the icecast_pass to /etc/airtime if /etc/airtime didn't exist assuming that it was a new install, when we reprovision we move /etc/airtime/airtime.conf to airtime.conf.tmp and do the whole install again, so the proper thing to check for is whether /etc/airtime/airtime.conf

I just pushed a fix #946 that should fix this issue.

PR #936, opened 2020-01-06
[WIP] Remove notes about manual intervention required for Liquidsoap 1.3.3

paddatrapper:


Robbt:
Maybe we should hold of on merging this until we release alpha.9 because I think a lot of people install from the release vs. the latest git.

paddatrapper:
True, good point

paddatrapper:
These are just notes around vagrant setup though, the documentation doesn't actually talk about liquidsoap versions at all. That seems to only in the release notes

Issue #917, opened 2019-11-13
Set CORS URLs in web installer


paddatrapper:
The CORS Urls setting is required when running LT behind a proxy (Containerisation is a common use case). The only documentation currently is [here](http://libretime.org/manual/general/), but it only contains one line:

> The Allowed CORS URLs is intended to deal with situations where you want a remote site with a different domain to access the API.

This should be elaborated on and highlight proxy use-cases. Maybe a "Common Setups" section in the Getting Started section, or Tutorials.

fauust:
Hi @paddatrapper!
I agree, I had the same problem when installing LT behind haproxy.
If you have a bit of time why no trying to submit a PR for the documentation? I will be happy to help you to review this PR.

numericalatina:
Hello.
I think this is important, as it makes LibreTime unusable on most browsers, unless you can find this issues on GitHub and apply the configuration.
This from https://github.com/LibreTime/libretime/issues/822#issuecomment-546609291 worked for me (replacing *domain* and *subdomain* with the real ones, of course)

```
https://sub.domain.org
http:/sub.domain.org
http://sub
http://localhost
https://sub
https://localhost
```

But i'm not really sure why, and what is the relation with the proxy.
In my case LibreTime is installed in a Docker container, behind an Nginx reverse proxy.

May be you can explain why this is necessary so we can document it?
Thanks

numericalatina:
Also, it's not clear to me if this can be configured in **airtime.conf** (or any other config file) rather than in the user interface

paddatrapper:
So preferences are accessed from the Database in `airtime_mvc/application/models/Preferences.php`. The CORS settings are stored there. The config file values are read/manipulated from `airtime_mvc/application/configs/conf.php`. There seems to be no relation between the two, so it would seem that it cannot be configured from anywhere outside of the UI

paddatrapper:
> But i'm not really sure why, and what is the relation with the proxy.
Cross-Origin Resource Sharing (CORS) is a mechanism that uses additional HTTP headers to tell browsers to give a web application running at one origin, access to selected resources from a different origin. More information is here - https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS

paddatrapper:
Please review #938 and let me know if that is clearer

fauust:
Hi @paddatrapper !
#938 looks good to me but I believe that there is a workaround that may have it's place in the documentation.

[...]

PR #939, opened 2020-01-07
[WIP] Move CORS URLs configuration into the config file

paddatrapper:
From #917, it turns out that you need the CORS URLs settings to access the CORS URLs settings. So this needs to be set before the user is able to log into LibreTime. This PR moves the configuration into the config file which means the system administrator can set it properly before running through the install guide

To do:

- [x] Move references to DB to config file
- [x] Write existing values to config file
- [ ] Fix bugs
- [ ] Document new way of setting CORS URLs

### DIFF
--- a/airtime_mvc/application/common/CORSHelper.php
+++ b/airtime_mvc/application/common/CORSHelper.php
@@ -29,7 +29,7 @@ class CORSHelper
     {
         $allowedCorsUrls = array_map(
             function($v) { return trim($v); },
-            explode(PHP_EOL, Application_Model_Preference::GetAllowedCorsUrls())
+            explode(',', Config::getConfig()['allowed_cors_urls']))
         );
 
         // always allow the configured server in (as reported by the server and not what is i baseUrl)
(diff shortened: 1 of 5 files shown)


Robbt:
I think that adding this during the web installer would make more sense than requiring someone editing a config file. I think that if easily feasible we should support CORS from both the DB and the config file. I could probably help add a new page to the setup for CORS specifically since I've been playing around with that already.

paddatrapper:
Would work if we can verify that users can reach the web installer without CORS URLs being set

Robbt:
I think they should be able to. It is pretty much a straight-forward apache + php based setup w/o any heavy duty back-end javascript going on.

paddatrapper:
Just need to check if there are any API calls I think. If there are none, then we are good to go

Robbt:
Yeah, there couldn't be any API calls because the web installer assumes the database isn't setup yet. Feel free to work on the install page if you want, I'll post an update on here if I start coding it, not sure when I'll find the time at the moment.

paddatrapper:
I'm going to close this in favour of doing it in the web-installer

Issue #928, opened 2019-12-14
no jack output on alpha 8


refutationalist:
Hello!   I'm working with Alpha 8 under 18.04.3 LTS, via the install script.   I'm trying to fit it into an existing broadcast chain that leans pretty heavily on JACK.  I went into settings and turned on JACK output support after getting jackd running for the www-data user.    No connections appear to be made from liquidsoap.   I also turned off my jackd to see if liquidsoap start up one of its own, which it does not appear to.   The jack module for liquidsoap is installed.

I took a look at the liquidsoap config, and it does not appear to have a stanza for jack output.  I tried adding it something that should work in, but that failed.

Looking for pointers, including the most recent startup from ls_script.log, which is after I removed my jack stanza after it failed.
[output.txt](https://github.com/LibreTime/libretime/files/3963868/output.txt)

Robbt:
Yeah this isn't currently something that we have put the time into figuring out. I think the place to start would be looking at the output.jack API here - https://www.liquidsoap.info/doc-1.4.0/reference.html#source-output
and then adding a stanza to this file using the alsa output as a reference - https://github.com/LibreTime/libretime/blob/1c2906f5aba167c398865a1ee93febd4e96ed0e2/python_apps/pypo/liquidsoap/ls_script.liq#L345-L389

You will also need to enable it more than just adding it. I'm going to say this is less of a bug and more of a feature request for us to support JACK. If you figure it out and want to share your code feel free to do so. If you need more assistance with troubleshooting liquidsoap you might get more reliable information upstream from the liquidsoap developers as we are not experts on it.

refutationalist:
Thanks for the background info, that's really helpful.

I tried a basic liquidsoap script that just output a file via jack and got it working.  I must've gotten it wrong the first time I tried it, as I added it back in and jack is now functional.  It's a very trivial change, but I'll include the patch.

Relatedly, is all dev discussion done in GitHub issues, or is there a forum or freenode or something?   I can't find somewhere else to ask this.

[readd_jack.patch.txt](https://github.com/LibreTime/libretime/files/3964022/readd_jack.patch.txt)

Robbt:
Yeah, we have a forum at https://discourse.libretime.org/ and a real-time chat mattermost instance at https://chat.libretime.org

Thanks for contributing, it might make sense to open a PR or add this as a feature request or we can just close this and hope that someone else finds it if they want to use jack.

paddatrapper:
[...]

PR #941, opened 2020-01-08
Add Jack support to liquidsoap output

paddatrapper:
Fixes: #928. I haven't tested it, but applied the patch from #928

### DIFF
--- a/python_apps/pypo/liquidsoap/ls_script.liq
+++ b/python_apps/pypo/liquidsoap/ls_script.liq
@@ -396,6 +396,13 @@ if output_sound_device then
 	end
 	%endif
 
+    %ifdef output.jack
+    if output_sound_device_type == "Jack" then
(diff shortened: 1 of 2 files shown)


Robbt:
I want to test this but it would mean diving into jack, which I assume will take some time based upon my past experiences with it. Or we could just merge it and say "jack support is experimental and untested." 

What do you think ?

paddatrapper:
My basic testing seemed to work, but jack on the command line is not my strong point either. Yeah, I think let us mark it as experimental and get the people in #928 test and see if they have issues

Robbt:
Sounds good to me, I'll test to confirm that this doesn't break anything obvious and then merge it and we can add that it was added as experimental to the release notes. I imagine that Jack could be very useful for a station in terms of doing software based audio compression and/or signal manipulation etc.

Issue #879, opened 2019-09-26
Statistics


hopfrogsalerno:
Statistics don't work:  _a: Please make sure admin user/password is correct on Settings->Streams page.

RememberForgottenHits:
Hi,

Check again to see if you still have the correct passwords. Admin and stream passwords must be different.

paddatrapper:
Also please provide more information as follows:

**Describe the bug**
A clear and concise description of what the bug is.

**To Reproduce**
Steps to reproduce the behavior:
1. Go to '...'
2. Click on '....'
3. Scroll down to '....'
4. See error

**Expected behavior**
A clear and concise description of what you expected to happen.

**LibreTime version:**
Version from the upgrade popup if you can reach it.

**Installation method:**
 - OS: [e.g. Ubuntu]
 - OS Version [e.g. 16.04.5 LTS (Xenial Xerus)] 
 - Method: [e.g. `./install` script or packages]
 - Details: [how did you call the install script, where did you get packages from]

**Screenshots**
If applicable, add screenshots to help explain your problem.

**Client (please complete the following information if applicable):**
 - OS: [e.g. Fedora]
 - Browser [e.g. chrome, safari]
 - Version [e.g. 22]

**Additional context**
Add any other context about the problem here.

Robbt:
I think the bug is pretty straight forward and is related to the way that we don't initially setup a custom icecast password see #86 - so the default streaming is insecurely hard coded and you need to switch to custom and change the password for statistics to work and be captured in the first place.

This should be a high priority in terms of improving the initial setup to be more secure. We need to ensure that we change the default icecast password and then save it so that the statistics page picks it up automatically. 

If you change to custom and enter a custom icecast2 password in your stream settings "Statistics" will start working.

paddatrapper:
With #933, we update the admin password in  icecast, however I just tested it and we still get the following error: `Please make sure admin user/password is correct on Settings->Streams page.` I am not sure if we correctly update the admin password in the DB. In the stream settings, the admin username is also blank

paddatrapper:
Ah, we set the password correctly, however the `s1_admin_user` is empty in the DB. This should default to `admin`

Robbt:
Makes sense, we should probably set s2_admin_user & s3_admin_user since I put the password into those as well and it'll save someone the trouble of inputting it when they enable stream 2 and 3.

PR #944, opened 2020-01-08
Add default icecast admin username

paddatrapper:
Fixes: #879


Robbt:
Can we add one for s1_admin_user and s3_admin_user so that it doesn't need to be added if people enable the additional streams.

paddatrapper:
Added for streams 2-4

PR #953, opened 2020-01-22
Fix existing config check in install

paddatrapper:
`-e` is the way to check if a file exists in bash


Robbt:
I think the issue with fixing this is that it would break the ability to run install to upgrade a libretime instance. I need to do more testing but if my memory is serving me correctly the broken check is what makes LibreTime go forward with installing everything even though it already exists as discussed in #933

paddatrapper:
So this was introduced in #933 and #947. At the moment, the error means that the icecast password is changed every time the script is run, resulting in inconsistent passwords between LT interface and Icecast (I seem to recall, been a while since I worked with it).

Robbt:
Ohh ok, I thought we had fixed that issue, I'll work on testing this. Our install/upgrade process has always been the least developed part of the project.

Robbt:
Ok, so I just did a fresh install off the master and "hackme" was still the default password. 
When I try to run "sudo ./install -fiap" from /vagrant with this script it freezes at 
"Failed to stop airtime-media-monitor.service: Unit airtime-media-monitor.service not loaded."
I think I ran into this issue before. 

After about 30 seconds or however long it took me to write this it resumed and now on vagrant at least the installer no longer has apache running on port 8080 and when I go to port 80 - I can't really test the web-based installer because I am using lynx and it doesn't work without javascript. Might just need to run "vagrant provision" to test it.

Robbt:
Ok, ugh, I tried to push a fix to add -e to line 909 so that the icecast fix would be added as well and it pulled all of the rest of the commits into here as well, but I think this is working.

Robbt:
But yeah I think this is ready to merge based upon my testing. It no longer overwrites the install file and the media-monitor is still an issue but it is separate one already noted in the comments of the install file.

Issue #185, opened 2017-04-23
Has anyone succeeded in geting aac streaming to icecast?


ghost:
I seem to have all the right codecs installed. The LT status page says it is connected to aac, even when the stream is disabled. The streams portion in config just says 'Getting information from server'.
I have a vague recollection of having to do a special compile of LiquidSoap to do this in AT 2.5x.
Is this still needed?

hairmare:
If you are on Debian or Ubuntu the recompile is probably still needed, the package on those distros hasn't been updated in ages :( I think the [Rabe Liquidsoap Distribution](https://build.opensuse.org/project/show/home:radiorabe:liquidsoap) (RaBe LSD) CentOS package has working aac but I haven't tested that yet.

Robby262:
I am on Ubuntu Trusty. Daniel James in AT (actually he is part of the LS team) did a compile that worked with 2.5.2a on Ubuntu 120.7 at least. But I can't find it now.

hairmare:
I'd assume that that binary is on http://apt.sourcefabric.org/ with the related build scripts being in https://github.com/Airtime/airtime-packaging/tree/master/liquidsoap. I think this is something we need to port off of that mirror.

Robby262:
Update. The stable LS binaries in Ubuntu & Debian distribution omitted to dynamically load external decoders are startup. So the fkaac encoder codec will not load and that is why it does not work without a recompile of Liquidsoap. I, and many others have found this difficult/impossible in my distribution.

**Workaround**
I suggest that for the next release that this should be the default. At least it works.

Edit: /usr/local/lib/python2.7/dist-packages/airtime_playout-1.0-py2.7.egg/liquidsoap/acc.liq
Change the FDK-AAC section to look like this:

# FDK-AAC is the only good AAC encoder. libvoaac is deprecated and aacplus is subpar. 
# The difference in compression quality is clearly audible. -- Albert
      %ifencoder %aac
       if type == "aac" then
         %include "aac.liq"
        end
       %endif

#      %ifencoder %aacplus
#       if type == "aacplus" then
#          %include "aacplus.liq"
#        end
#        %endif

#       %ifencoder %fdkaac
#        if type == "aac" then
#           %include "fdkaac.liq"
#       end
#        %endif
    else
        user_ref = ref user
        if user == "" then
            user_ref := "source"
        end

Robby262:
**Further notes.**
[...]

PR #967, opened 2020-03-03
[WIP] Ubuntu 19.10 Eoan Ermine Support

kmahelona:
I'm not sure if we want this since this isn't an LTS version of Ubuntu but I'm providing this because I do want to work on aac streaming support with liquidsoap ~1.3. and presumably this will make the Ubuntu 20. support easier when that's release.


paddatrapper:
I think the more the merrier, so adding 19.10 is a good idea. We just need to deprecate it when upstream support is dropped. `libfdkaac-ocaml` should add ACC support to liquidsoap. After that it should just be a matter of adding the format to the liquidsoap calls made my LibreTime

kmahelona:
Cool.

Yea I installed `libfdkaac-ocaml` but still got this error in `/var/log/airtime/pypo-liquidsoap/ls_script.log`,
`[dynamic.loader:3] Error while loading dynamic fdkaac encoder at /usr/lib/ocaml/fdkaac`

paddatrapper:
Maybe create a forum post and we can debug it?

kmahelona:
More issues possibly with 19.10 and liquidsoap 1.3.7,

```
2020/03/09 02:17:19 [decoder:3] Method "AAC" accepted "audio/aacp".
2020/03/09 02:17:19 [http:3] Decoding...
2020/03/09 02:17:20 [http:2] Feeding stopped: Faad error: Unexpected channel configuration change
```

I'm having issues decoding AAC web streams. MP3 web streams are playing fine.

kmahelona:
I've managed to get liquidsoap 1.4.1 built and working in hopes that it would fix my `faad` error above. It didn't appear to fix it but I also ran into a bunch of other issues where the liquidsoap scripts need to be updated for 1.4.1 and I kinda guessed it and got something that runs but basically can't play files but can now play an aac webstream (I think I broke things with the `cross()` functions).

In 1.4.1 I still get a bunch of the same errors above but after about 6 of those I get `Channel coupling not yet implemented`,
```
[http:2] Feeding stopped: Faad error: Unexpected channel configuration change
[decoder:3] Method "AAC" accepted "audio/aacp".
[http:3] Decoding...
[http:2] Feeding stopped: Faad error: Unexpected channel configuration change
[decoder:3] Method "AAC" accepted "audio/aacp".
[http:3] Decoding...
[http:2] Feeding stopped: Faad error: Channel coupling not yet implemented
[decoder:3] Method "AAC" accepted "audio/aacp".
[http:3] Decoding...
```
And then it plays the aac stream without error.

I think what I'll try now is to build liquidsoap 1.3.7 using opam and see if that fixes my issues. If not... maybe the liquidsoap scripts we have are causing this faad error?

Is anyone else having trouble playing (decoding) aac streams?

kmahelona:
Ok, so sorry for all the unnecessary posts. It turns out the icecast streams I'm trying to play are so old that their aac codec isn't liked by new versions of liquidsoap (probably a faad version thing). I've tested this by streaming my own he-aac streams to both an icecast2 and a shoutcast server and I am able to play them. It means I have to convince the people, AVC, who run the icecast streams for 21 stations to update their stuff!

paddatrapper:
Is it possible to do those two things in Ubuntu 18.04?

kmahelona:
I'm not sure, but I could give it a try if that would be useful. I should probably be using vagrant and then I could test these things much more quickly ;).

Robbt:
Vagrant is super helpful for development and relatively straight forward to get setup - the manual is on http://libretime.org/vagrant/ 

In terms of the streaming support this doesn't seem like a good reason to deny support if anything we should open up a separate issue for this and try to isolate the cause and test the streams. 

Are there public streams where we can test different versions of LibreTime to see if they support HE-AACv2 ? 

I also don't think we have ever support AAC streaming out of the box, this has always required some tweaking to get it working as far as I know see #185 

It would be nice to support it but I think it maybe more patent encumbered and less free than other solutions such as opus and/or mp3 (now that the patents for it expired).

kmahelona:
Ok. This pull request is basically ready only I haven't tested the vagrant aspects of it (just didn't get around to it, I do use vagrant for other local dev).

I've got some public HE-AACv2 links that we can use e.g. http://apple.irirangi.net:8000/I_Waatea.aac (and more here: https://github.com/TeHikuMedia/whakaruruhau-website/blob/master/src/assets/radio_streams.ts but please don't do any heavy loading).

Robbt:
Looks like Ubuntu 20.04 utilizes Php 7.4 which deprecates a whole new subset of features. I think once we merge python 3 support then we can test that and see if it works out of the box or requires code changes. If we could support 20.04 then perhaps we can exclusively support LTS releases especially if the package selection for things like silan and liquidsoap aren't broken.

Robbt:
So now that 20.04 is out I think that we might as well close this. I'm working on a PR for 20.04 and we could make it a point to only support LTS versions of Ubuntu. Feel free to re-open this if you want me to reconsider merging 19.10 support.

PR #985, opened 2020-03-22
Russian translation updated

farggus:
Full revision of existing translation;

- Updated russian translation of the LibreTime software interface;
- This includes a re-reading of previous translations, with a good amount of typos and  grammar errors now corrected;
- The meanings of certain words and phrases which did not convey the essence of the LibreTime functionality were corrected;
- Fixed numerals, conjugations and other grammatical forms;
- Added translation of new words and phrases;
- Language and regional adaptation.

I also adjusted the display of some interface elements, which depend on the appearance of the translated phrases.

But there is still work - some blocks have not been translated, for example, such as the new section “Track types”.


paddatrapper:
Thanks for the translations!

PR #993, opened 2020-04-01
Fix source disconnection with recent Liquidsoap

jeromelebleu:
The `kick` command has been removed since Liquidsoap [1.3.2](https://github.com/savonet/liquidsoap/blob/master/CHANGES.md#132-02-09-2017) as it was a duplicate of `stop`. Thus, source disconnection is broken and especially the DJ one - which is quite problematic for scheduled shows.


paddatrapper:
Is `stop` introduced in 0.3.7, so supported in all our liquidsoap versions

PR #1001, opened 2020-04-18
changed pages to nav due to deprecation of pages in mkdocs

### LINKED ISSUE #986: Docs: Links are broken on site
The links on the docs site under tutorials and howtos (for example http://libretime.org/manual/tutorials/upload-tracks ) return 404s. However under `mkdocs serve` they are found correctly. I think it is something to do with how the web-server is configured on the site server.

Robbt:
This might work for #986 
Since pages was deprecated and navs is now preferred, I am wondering if switching to navs will somehow fix the broken links for tutorials.

It is a difficult thing to troubleshoot because it works on local mkdocs instances but so far the github deployed version is suddenly broken. Anyways, this shouldn't hurt and might the lack of inclusion.

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -14,7 +14,7 @@ repo_url: https://github.com/LibreTime/libretime/
 extra_css:
   - '_css/term.css'
 
-pages:
+nav:
   - 'Home': index.md
   - 'What is LibreTime?': manual/index.md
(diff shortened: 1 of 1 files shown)


paddatrapper:
I'll merge and we can see what happens. I have edited your description so that it will not close that issue so we can manually close it after testing if it works

PR #1007, opened 2020-04-23
Make travis CI use 1.0 explicitly

### LINKED ISSUE #986: Docs: Links are broken on site

Robbt:
This is an attempt to fix #986 - as the build logs are showing warnings related to the beta travis CI 2.0 or dpl v2 command line

For instance 
```
 Build config validation
deploy: deprecated key skip_cleanup (not supported in dpl v2, use cleanup)
deploy: deprecated key skip_cleanup (not supported in dpl v2, use cleanup)
deploy: deprecated key skip_cleanup (not supported in dpl v2, use cleanup)
deploy: deprecated key skip_cleanup (not supported in dpl v2, use cleanup)
deploy: deprecated key skip_cleanup (not supported in dpl v2, use cleanup)
deploy: deprecated key skip_cleanup (not supported in dpl v2, use cleanup)
root: key matrix is an alias for jobs, using jobs
deploy: key api_key is an alias for token, using token
deploy: key github_token is an alias for token, using token
env: key matrix is an alias for jobs, using jobs
root: key matrix is an alias for jobs, using jobs
deploy: missing strategy, using the default git
root: missing os, using the default linux
deploy: key api_key is an alias for token, using token
deploy: key github_token is an alias for token, using token
env: key matrix is an alias for jobs, using jobs 
```

I don't know how to test this PR without merging it and I'm not sure it will fix the docs issue, but it might and should be easy to revert if it breaks anything.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,8 +38,8 @@ jobs:
   - env: PYTHON=true
   include:
   # having the latest run as php 7.3 to do the python tests
-    php: 7.3
   - env: PYTHON=true
+    php: 7.3
 addons:
   apt:
     packages:
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,13 +33,14 @@ jobs:
   # there are some issues with phpunit as well as some deep in zf1 on 7.3
   - env: PYTHON=false
     php: 7.3
-  exclude:
-  # by excluding all of python we make way to just run python tests in one seperate instance
-  - env: PYTHON=true
   include:
   # having the latest run as php 7.3 to do the python tests
   - env: PYTHON=true
     php: 7.3
+  exclude:
+  # by excluding all of python we make way to just run python tests in one seperate instance
+  - env: PYTHON=true
+
 addons:
   apt:
     packages:
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,6 +33,7 @@ matrix:
   # there are some issues with phpunit as well as some deep in zf1 on 7.3
   - env: PYTHON=false
     php: 7.3
+jobs:
   exclude:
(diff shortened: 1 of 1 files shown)


Robbt:
Ok I was thinking of merging this as requested by @paddatrapper at #986 I am not sure that this will make dpl v1 happen by default I am just using the suggestion from https://docs.travis-ci.com/user/build-config-validation/

But I don't think it will have any change due to https://config.travis-ci.com/explore not showing anything different using version 1.0 - it is likely still using dplv1 but they changed their build parser slightly and this affected how it interpreted the jobs we want to run.

So I suspect that we need to do something else to get it to run the python script separately.

Robbt:
I decided to try to have php 7.0 run the python tests because it is the latest version of php that still can pass the phpunit tests and I suspect that travisCI parser quit creating an independent job. Also going to merge this and see if it fixes things.

Robbt:
Ok, so I had to deep dive into the build matrix concept of travis cI and  I realized what was happening. We previously could have the python tests only run on php 7.3 by excluding all jobs in the matrix that had python=true and then manually including a job with php 7.3 and python=true.

This quit working for some reason and anytime I excluded python=true it would also exclude the include job.

So I had to basically explicitly exclude every php version other than php 7.3 where python=true and I got rid of the include because it was causing an additional job to run. 

Not sure why this changed but I'm going to merge this and we can see if the docs are updated.

PR #951, opened 2020-01-16
Port to Python 3

paddatrapper:
Replaces python 2 support with python 3 support. Not backwards compatible

Closes #257

PR #1009, opened 2020-04-23
test using bionic VM for travisCI to test mkdocs fix

Robbt:
This is an attempt to fix the issue with hidden pages no longer being added to mkdocs navigation only when deployed from travisCI.


Robbt:
Going to close this and figure that the issue can be resolved with #951

Issue #990, opened 2020-03-29
Track Type database migrations not applied


itcrowdsource:
**Describe the bug**
After clicking the login button this message appears:

Oops!
Something went wrong!

**To Reproduce**
1. Install
2. Login

**Expected behavior**
Be able to login

**LibreTime version:**
Latest version from GIT

**Installation method:**
 - OS: Ubuntu 
 - OS Version 18.04 Bionic
 - Method: ./install`

**Screenshots**
![image](https://user-images.githubusercontent.com/20809792/77837220-888c9300-715e-11ea-8f72-64ef7101b874.png)

**Additional context**
Zendphp.log

```
2020-03-28T22:38:33+00:00 ERR (3): <url> [ErrorController.php:54 - errorAction()] - An internal application error has occurred.: PDOException: SQLSTATE[42P01]: Undefined table: 7 ERROR: relation “cc_track_types” does not exist
LINE 1: SELECT id, code, type_name, description FROM cc_track_types …
^ in /usr/share/airtime/php/airtime_mvc/application/common/Database.php:24
Stack trace:
#0 /usr/share/airtime/php/airtime_mvc/application/common/Database.php(24): PDOStatement->execute()
#1 /usr/share/airtime/php/airtime_mvc/application/models/Tracktype.php(124): Application_Common_Database::prepareAndExecute(‘SELECT id, code…’, Array, ‘all’)
#2 /usr/share/airtime/php/airtime_mvc/application/models/Tracktype.php(98): Application_Model_Tracktype::getTracktypesData(Array, NULL)
#3 /usr/share/airtime/php/airtime_mvc/application/controllers/plugins/PageLayoutInitPlugin.php(146): Application_Model_Tracktype::getTracktypes()
#4 /usr/share/airtime/php/airtime_mvc/application/controllers/plugins/PageLayoutInitPlugin.php(85): PageLayoutInitPlugin->_initTranslationGlobals(Object(Zend_View))
#5 /usr/share/airtime/php/airtime_mvc/application/controllers/plugins/PageLayoutInitPlugin.php(47): PageLayoutInitPlugin->_initGlobals()
#6 /usr/share/airtime/php/vendor/zf1s/zend-controller/library/Zend/Controller/Plugin/Broker.php(260): PageLayoutInitPlugin->routeShutdown(Object(Zend_Controller_Request_Http))
[...]

PR #1011, opened 2020-04-24
changed track type id to serial for auto increment

Robbt:
This should fix the issue with auto upgrades. I tested it on Ubuntu Xenial vagrant box and the use of serial worked whereas the previous code was failing to create the DB.


Robbt:
This fixes #990

PR #1019, opened 2020-05-04
Track Type Default

codenift:
Option to set a default value for Track Type (set to none or any visible track type). I will do a follow up to this soon with option to change the timing to leave track type selection on upload (None, By session, 15 minutes, 1 hour, a day, ect..). At the moment it's just set to session, when it expires it goes back to default.


codenift:
Ok, so I do have an update coming for adding a track type default for podcast but I will put it in a separate PR instead of here. It will just be adding an extra option for Podcast feeds, so if a track type is selected, it will set a track type to the imported audio on that feed (every feed with it's own default). I should have a PR by tomorrow after I test it further.

Robbt:
Ok, tested this and it works great.

One question though, from a UX perspective it probably makes more sense for the default track type to be under the track type settings page. It is hard to find under the general settings page. 

How hard do you think it would be to move it over to that page vs. the general settings page ?

codenift:
I thought about it. Although it's possible and it's related to Track Type, I want to see if we can keep things a little similar to the User page (not the best layout, but the best we have now) to make it easier for development down the road. That option would still save into General settings even if we decide to change it to another page without breaking the site and doesn't involve updating the database.

The way I see it, Users and Track Types work pretty much like custom post types, where the options for that type (in this case, General Settings) are apart. I found that In Libretime there's no need to mess with the database just to add another setting in General so it made sense to me. Should be simple to move to any page later, just need to make sure we do it right so we don't screw the layout. Pretty much work like options in WP. I think we can leave it where it is for now since there are other Default settings for tracks there too (Fade In, Fade Out, Crossfade, ect..). 

The Track Type branch (which i still have) can serve as a template for now and is based on Users incase adding another feature that needs such data management. And if the layout is ever changed in the future, it will be easy to keep these kind of pages with the same look.

Robbt:
Ok, I'm cool with leaving where it is for now, you are right that there is a lot of reworking that could be done. I'm going to merge this now.

Issue #516, opened 2018-09-14
Change airtime_analyzer to airtime-analyzer.service


AEK-BKF:
Hi,
I've installed Libretime on Ubuntu Server 16.04 (64bit) via Virtual box.
I got this error when I run : **sudo service airtime-analyzer start** 

> Failed to start : airtime-analyzer.service : Unit airtime-analyzer.service not found

![error](https://user-images.githubusercontent.com/16412739/45532264-51a89380-b7eb-11e8-8f21-619bbec7a038.png)

Please any help

AEK-BKF:
Ah Ok ! Never mind !
I wrote wrong command : _sudo service airtime-analyzer start_
The correct one is : **sudo service airtime_analyzer start**

Thanks :)

Robbt:
Actually this is a point of annoyance for me personally because I have ran into this. I think we should use the same string format for services. Does this seem like a good idea to anyone else, either changing the old ones to use _ or modify analyzer to use the - instead. If we don't want to break any scripts/past muscle memory we could

AEK-BKF:
+1
:+1:

hairmare:
This has annoyed me as well 😉The `_` is pythonic and the `-` is how everyone else usually does it. We should probably just rename the service file with the pythonic `_` to a `-` and live with the actual script have a different file that the unit.

hairmare:
I vote for `-` since it seems more natural in systemd and a majority of the services we already have use `-`. My decision isn't based on hard fact but more of a gut feeling so it doesn't really count though.

Robbt:
Since there is a consensus here I'm going to re-open it as an issue. I just edited the title to reflect the change.

hairmare:
Do we want to combine this with the overdue change of switching to `libretime-*` service names?

Robbt:
We could do certainly do that, it would make sense, is there a easy way to maintain an alias to the old ones ?

paddatrapper:
Looks like you can alias units (using `Alias=` under `[Install]`), but this is only relevant for installing units (`systemctl enable` and `systemctl disable`). We could symlink back to the old units, which should work as expected.

paddatrapper:
The recommended way in the [docs](https://www.freedesktop.org/software/systemd/man/systemd.unit.html#Description) is to use a symlink

Issue #940, opened 2020-01-08
Correctly support upgrade using install script


paddatrapper:
**Is your feature request related to a problem? Please describe.**
There is a bug in the install script where `OLD_CONF=$(grep "[media-monitor]" /etc/airtime/airtime.conf)` doesn't provide the intended result because we will always assume that an existing conf is an upgrade because the `[ ]` are brackets in grep and thus they will match if any of the letters match. This means that the script will always assume that it is an upgrade if `/etc/airtime/airtime.conf` exists.

So when I take that out and run install it doesn't assume it is an upgrade and nothing happens. So I think rerunning install to facilitate upgrading is actually an unintended feature of this bug. But it hangs trying to stop media-monitor when I run it. So yeah, I think I should open a bug for that and perhaps add a --upgrade flag to ./install that will facilitate upgrading w/o assuming the user had an instance of airtime or always assuming it is an upgrade when they run install.

**Describe the solution you'd like**
We should either add a `--upgrade` flag to `install`  that will facilitate upgrading w/o assuming the user had an instance of airtime or always assuming it is an upgrade when they run install.

**Additional context**
Found by @Robbt [here](https://github.com/LibreTime/libretime/pull/933#issuecomment-571680862)

Robbt:
This is related to and could help fix #623

paddatrapper:
Fixed by #953

PR #1023, opened 2020-05-05
Rename all systemd, sysvinit and upstart files to libretime

paddatrapper:
Also cleans up all documentation and website references to sysvinit calls to use systemd

Fixes: #516


Robbt:
Ok, this seems to work. Noticed one spot where you forgot to change airtime-liquidsoap to libretime-liquidsoap as well - in build/airtime-setup/forms/finish-settings.php

It seems like a good time to also strip out some of the references to welcome to airtime etc.

Not sure if we should go as far as renaming /etc/airtime to /etc/libretime/libretime.conf and the like or if that would potentially cause headaches to upgraders but it might be a good idea to do all of this stuff at once now that we are switching to python3.

Thanks for starting this process.

paddatrapper:
I think moving the config file would be a good idea for 3.0, but needs a migration plan. Symlink to airtime.conf for upgrades perhaps? I also think that the systemd symlinks made here should only happen on upgrade.

Robbt:
Yeah, we really should work on having a functional upgrade or at least making it not rely upon a misconfiguration that treats every upgrade as a new install as noted in #940

Robbt:
So I suggest we fix https://github.com/paddatrapper/libretime/blob/a254a3222ecd668241eeae453d78814709505ebf/airtime_mvc/build/airtime-setup/forms/finish-settings.php#L8-L16 and change the airtime-liquidsoap to libretime-liquidsoap and then we could merge this (unless we want to do further changes regarding the aliases only being used on upgrades).

paddatrapper:
I've fixed the references, and made the symlink only occur when the old systemd unit exists and is not already a symlink.This should ensure that the migration only occurs once on systems that have existing installs and not at all on new installs.

Robbt:
I tried testing it up on a upgrade of my Debian Buster virtual machine and I don't think the aliases are working after running the install script. So that code might need some work.

paddatrapper:
Vagrant upgrades seem to be weird... They scrap the airtime.conf file too.

What is the output of `ls /etc/system/systemd/`?

Robbt:
Actually I think it is our upgrade process that scraps the /etc/airtime/airtime.conf as referenced above, it always thinks that it is a new install, but maybe it is specific to vagrant.

paddatrapper:
Interesting. To be fair, I have never run subsequent install steps in production. I tend to just destroy the LXC container and create a new one for upgrades

paddatrapper:
@Robbt fixed the issue. It was due to on upgrade, the installer happily removing all systemd units that were installed previously

PR #1034, opened 2020-05-12
Cors in setup

### LINKED ISSUE #917: Set CORS URLs in web installer

Robbt:
This fixes #917 
Needs tested on a completely new install but appears to be working.
Didn't pull existing CORS values from the database if for some reason the people see the install screen again it won't overwrite them. Tried to add that to text. Hopefully most people will ignore it unless they know otherwise that they need to input CORS values due to additional documentation etc that we write about setup for Docker/reverse proxy.


paddatrapper:
Works in a new install with proxy in front. The only thing I would suggest is including instruction to place each URL on a new line

Robbt:
Ok done. Wasn't sure if you needed to include the http:// prefix as well as https://URL, so if this isn't clear or we don't need http in there and just want the host name we can clarify that as well.

paddatrapper:
I'll text just using the hostname without protocol and see what happens

paddatrapper:
Just tested - you need to include the protocols

PR #1030, opened 2020-05-10
Replacing mkdocs with Jekyll

zklosko:
Picking up #590 and as a response to #1029, I'm attempting to migrate the Libretime site from mkdocs to Jekyll via Github Pages. The current changes of my repo can be viewed [here](https://zklosko.github.io/libretime/).

Fixes #1029

PR #1036, opened 2020-05-12
if the screen is small enough to have a scrollbar always go to the top

### LINKED ISSUE #984: Poor layout on Install screen for rabbitmq error message
**Describe the bug**
The error message returned on the install screen for rabbitmq sometimes displays only the second line "server is running and your credential are correct". But the full error message is "Couldn't connect to RabbitMQ server! Please check if the server is running and your credentials are correct." which is completely different in meaning.

**To Reproduce**
1. During install, enter invalid values for rabbitmq server
2. Click on "Next"
3. Screen refreshes, and displays only second line of error message

**Expected behavior**
The error message should clearly indicate the problem, even when truncated. 
Suggestion: "Couldn't connect to RabbitMQ server! Perhaps the server is not running or your credentials are incorrect."

**LibreTime version:**
V8 Alpha (I think... - It would be useful to have version number displayed on every install screen)

**Installation method:**
 - OS: Ubuntu
 - OS Version: 18.04.03 LTS bionic
 - Method: download .tar.gz, run ./install.sh
 - Details: 

**Screenshots**

**Client (please complete the following information if applicable):**
 - OS: Debian 10 GNU/Linux (buster), MATE desktop
 - Browser: Chromium
 - Version: Version 80.0.3987.132 (Developer Build) built on Debian 10.3, running on Debian 10.3 (64-bit

**Additional context**

Robbt:
This should fix #984 by moving the scrollbar to the top on any new page load. Thus preventing the top of the message from being cut-off as indicated in the bug report.

### DIFF
--- a/airtime_mvc/public/js/setup/setup-config.js
+++ b/airtime_mvc/public/js/setup/setup-config.js
@@ -117,6 +117,7 @@ function submitForm(e, obj) {
     e.preventDefault();
     var d = $(e.target).serializeArray();
     addOverlay();
+    $(".viewport").scrollTop(0);
     // Append .promise().done() rather than using a
     // callback to avoid call duplication
     $("#overlay, #loadingImage").fadeIn(500).promise().done(function() {


paddatrapper:
The travis failure here is thanks to `dev_addr: '0.0.0.0:8888'` in `mkdocs.yml`. I believe that this is required for our dev setup because vagrant doesn't come up with a constant IP for every image on every platform... I'm not sure if there is a way to disable this check until #1030 is merged

Robbt:
Yeah it appears that our Travis CI setup is breaking down as PHP deprecates more functions included in various libraries, such as 7.3 PHP unit tests failing I think because the phpUnit version we are using that is compatible with Zend uses a deprecated functionality. 

The real question is why did the mkdocs issue just start appearing. I don't think we should be running mkdocs on our TravisCI tests so maybe we can easily eliminate it there.

Issue #957, opened 2020-01-28
Libretime behind Nginx Reverse Proxy w/ LetsEncrypt URL issues


guythnick:
I have been having a hard time getting Libretime to run correctly behind my reverse proxy.  I have added the CORS as mentioned in https://github.com/LibreTime/libretime/issues/822#issuecomment-546609291 

The reverse proxy is working with other sites, but I am missing something I think either in the /etc/airtime/airtime.conf  file, or the apache2 airtime.conf file.  Can someone point me in the right direction?

I am able to login and change/view settings.  But the playout page, previewing music, etc. is not displaying.  It appears to be forcing http://mysite, even though the reverse proxy is set to always re-direct to https.  

 Example log:
```
an 28 11:42:34 TerryRadio airtime-playout[9038]: 2020-01-28 11:42:34,846 [pypoliqqueue] [INFO ]  waiting indefinitely for schedule
Jan 28 11:42:40 TerryRadio airtime-playout[9038]: 2020-01-28 11:42:40,771 [api_client] [ERROR]  HTTP Error 500: Internal Server Error
Jan 28 11:42:40 TerryRadio airtime-playout[9038]: 2020-01-28 11:42:40,846 [api_client] [ERROR]  HTTP Error 500: Internal Server Error
Jan 28 11:42:40 TerryRadio airtime-playout[9038]: 2020-01-28 11:42:40,912 [listenerstat] [ERROR]  Exception: HTTP Error 500: Internal Server Error
Jan 28 11:44:41 TerryRadio airtime-playout[9038]: 2020-01-28 11:44:41,501 [api_client] [ERROR]  HTTP Error 500: Internal Server Error
Jan 28 11:44:41 TerryRadio airtime-playout[9038]: 2020-01-28 11:44:41,580 [api_client] [ERROR]  HTTP Error 500: Internal Server Error
Jan 28 11:44:41 TerryRadio airtime-playout[9038]: 2020-01-28 11:44:41,640 [listenerstat] [ERROR]  Exception: HTTP Error 500: Internal Server Error
Jan 28 11:45:01 TerryRadio CRON[11095]: (root) CMD (command -v debian-sa1 > /dev/null && debian-sa1 1 1)
Jan 28 11:46:56 TerryRadio airtime-playout[9038]: 2020-01-28 11:46:56,750 [listenerstat] [ERROR]  Exception: <urlopen error [Errno -3] Temporary failure in name resolution>

```
A default apache conf:  

```
[...]

PR #1037, opened 2020-05-13
Add reverse proxy documentation

### LINKED ISSUE #1025: Document steps for reverse proxy configuration
There are several issues with running LibreTime behind a reverse proxy where the proxy terminates SSL. These are described in #957. However there are additional steps that the administrator must perform for the reverse proxy to work (assuming the code issues are fixed). These include:

* Reverse proxy configuration
* CORS URLs configuration (in web installer when #917 is fixed)
* Icecast Hostname set (https://github.com/LibreTime/libretime/issues/957#issuecomment-625530020)

paddatrapper:
Fixes #1025

Mentions #957, so will need updating when that is fixed. Also will need including in #1030


zklosko:
I added it to #1030 👍

paddatrapper:
Thanks!

PR #1038, opened 2020-05-13
Removed mkdocs build from TravisCI tests

Robbt:
I'm not sure if this will disable updates of mkdocs but evidently our current mkdocs settings are not building any ways so it should fix the tests in the meantime.

### DIFF
--- a/travis/python.sh
+++ b/travis/python.sh
@@ -14,7 +14,3 @@ pushd python_apps/api_clients
 pip3 install -e .
 nosetests
 popd
-
-echo "Building docs..."
-mkdocs build --clean -q > /dev/null
-echo -n "done"


paddatrapper:
This will disable doc deploys. The deploy is done on Travis using the artifacts from the python build. See https://github.com/LibreTime/libretime/blob/8498029cd983536a6e3016780d0b1693985a8e01/.travis.yml#L107

Robbt:
Hmm, then there must be a different way of building w/o 0.0.0.0 as the IP address that will allow this to work ? Or we could just ignore it for now as the Jekkyl transition seems to be coming along quickly.

paddatrapper:
I think long term we need to overhaul our travis CI setup to act more like a pipeline:

1. Concurrently run all tests (PHP and Python)
2. Deploy docs on master
3. Deploy release artefacts on tags

Each stage should be completely separate from each other and not try use the artefacts from the previous stage. I have no idea how to set that up in Travis (GitLab CI built this way by default, so that is all I have used)

For the moment, We can probably ignore it, provided it is fixed before 3.0

paddatrapper:
Opened  #1039

PR #1046, opened 2020-05-18
Force fixed strings on grep for media-monitor

paddatrapper:
Fixes removing the config file on re-running install. This somewhat addresses LT handling upgrades badly when using `./install`. The issue was that it was grepping for `[media-monitor]`, which was being interpreted as a regular expression, and so matching lines with any of those digits in them. Thus, it always thought it needed to migrate from media-monitor and scrap the old config file

### DIFF
--- a/install
+++ b/install
@@ -768,7 +768,7 @@ loudCmd "./build.sh"
 if [ -f /etc/airtime/airtime.conf ]; then
     # TODO use VERSION or some other way to check for updates and handle
     # media-monitor case on it's own
-    OLD_CONF=$(grep "[media-monitor]" /etc/airtime/airtime.conf)
+    OLD_CONF=$(grep -F "[media-monitor]" /etc/airtime/airtime.conf || true)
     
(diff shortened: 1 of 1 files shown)


Robbt:
Going to test this but I think this is what helped upgrading happen unintentionally.

paddatrapper:
Yeah, now upgrading happens without removing a bunch of files first (mainly init system files and config)

paddatrapper:
@Robbt have you had a chance to look at this? It doesn't break upgrading, and ensures that files are persisted over updates

hairmare:
We should at some point get rid of grepping for `media-monitor` as noted in the TODO a couple of lines above what I'm merging.

I don't see any harm in merging it for a couple of reasons: 

a) the diff between fgrep and grep is irrelevant in this case since we don't use any patterns 
b) the amount of users doing a direct migration from airtime is shrinking and we should recommend they manage their ini file outside of triggering this code during a migration anyway 
c) I'm a fan of `|| true` in this case 😉

PR #1050, opened 2020-05-20
Fixing copyright, organization as per @robbt

zklosko:
Fixing #1030


Robbt:
This looks good. Maybe we should mention that the docs are GPLv2 and a mixture of copyright LibreTime community & SourceFabric. Just deleting the copyright is probably fine but we could also just include a variation of the booktype page I linked to before. What do other people think ?

paddatrapper:
The GPL requires that we retain copyright attributions, so we should include their and our copyright statements somewhere

zklosko:
@Robbt could you please link to that booktype doc here?

Robbt:
Here's the link to the copyright notice from the original source material for the mkdocs site - http://sourcefabric.booktype.pro/airtime-25-for-broadcasters/about-this-manual/

zklosko:
@Robbt Please review. Would hate to have incorrect copyright information on the site for too long.

Robbt:
Looks good. It's fine to link to the original source especially while it is still live.

Issue #1066, opened 2020-06-16
Error 500 when no show is filled with files


kaya-karaca:
**Describe the bug**
I get an 500 Error on the live-info API, whenever I have different shows setup but none of those has any file / mp3 loaded into them.

**To Reproduce**
1. Go to Calendar, setup some shows but don't load any tracks into them
2. Go to yourdomain.com/api/live-info/?callback
3. See error

**Expected behavior**
Still want to see the upcoming shows even if there is no file in them.

**LibreTime version:**
16bff4ff9.

**Installation method:**
 - OS: Ubuntu
 - OS Version: 18
 - Method: ./install
 - Details: Clean Install

paddatrapper:
What are the contents of `/var/log/airtime/zend.log`?

caveman99:
I've had the same issue and made a quick fix. The code is trying to access parts of the object that simply dont exist.

PR #1070, opened 2020-07-11
Prevent 500 errors when no content is available

caveman99:
this addresses #1066 (please edit out the whitespace change at the end of the file)


paddatrapper:
Thanks!

PR #1073, opened 2020-07-18
Corrupt Cover Art prohibits mp3 upload

caveman99:
Don't throw an exception on corrupted cover art image data in upload. The order of processing needs to be the other way around.

### DIFF
--- a/airtime_mvc/application/common/FileDataHelper.php
+++ b/airtime_mvc/application/common/FileDataHelper.php
@@ -394,16 +394,18 @@ class FileDataHelper {
               $im = @imagecreatefromjpeg($get_cont);
           }
 
-          if ($size){
-            $im = imagescale($im , $size);
-          }
-
-          if(!$im) {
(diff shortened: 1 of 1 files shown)


Robbt:
Thanks for the fix. This looks pretty simple, so switching the order of those two if statements fixes the problem because it doesn't try to access $im to scale it if it has an error ?

caveman99:
Yes, exactly - imagescale bombs out if $im is false. 

2020-07-18T15:54:16+02:00 ERR (3): [MediaController.php:161 - postAction()] - imagescale() expects parameter 1 to be resource, boolean given 

I fixed another problem in the same commit, The variable $imgname is accessed in the error message but does not exist.

2020-07-18T16:22:53+02:00 ERR (3): [MediaController.php:161 - postAction()] - Undefined variable: imgname 

The effect was a red bubble on the uploader and the file not imported. The patch fixed it on our live station immediately.

PR #1074, opened 2020-07-20
Add a switch to force 'assume SSL' mode behind a reverse proxy

caveman99:
Like in many other software projects, there are issues when you reverse-proxy from SSL to a non-ssl backend. The elegant but error-prone way would be auto-detecting proxy headers in the HTTP request. (What if a misconfigured or prior-unknown reverse proxy is used?) I have used another proven approach and added a force_ssl parameter to the general section of airtime.conf, The patch is non-invasive as with a working configuration it is ignored. Python is not my home turf but i'll try to wrap my head around it.


paddatrapper:
This is related to Issue #957. I hope to get to testing this later in the week

paddatrapper:
Thanks!

PR #1092, opened 2020-09-28
Ensure that the python version under test is installed

paddatrapper:
Fixes the current travis failures for Python 3.7

Issue #1111, opened 2020-11-18
Docs fail to build in Vagrant


paddatrapper:
The docs install script produces the following error:

```
==> ubuntu-bionic: Running provisioner: build-site-jekyll (shell)...
    ubuntu-bionic: Running: /tmp/vagrant-shell20201118-171947-ilj452.sh
    ubuntu-bionic: Installing Ruby
    ubuntu-bionic: Reading package lists...
    ubuntu-bionic: Building dependency tree...
    ubuntu-bionic: Reading state information...
    ubuntu-bionic: build-essential is already the newest version (12.4ubuntu1).
    ubuntu-bionic: zlib1g-dev is already the newest version (1:1.2.11.dfsg-0ubuntu2).                                                                                                          
    ubuntu-bionic: zlib1g-dev set to manually installed.                                                                                                                                       
    ubuntu-bionic: The following additional packages will be installed:                                                                                                                        
    ubuntu-bionic:   fonts-lato javascript-common libgmp-dev libgmpxx4ldbl libjs-jquery
    ubuntu-bionic:   libruby2.5 rake ri ruby ruby-dev ruby-did-you-mean ruby-minitest
    ubuntu-bionic:   ruby-net-telnet ruby-power-assert ruby-test-unit ruby2.5 ruby2.5-dev
    ubuntu-bionic:   ruby2.5-doc rubygems-integration
    ubuntu-bionic: Suggested packages:
    ubuntu-bionic:   gmp-doc libgmp10-doc libmpfr-dev bundler
    ubuntu-bionic: The following NEW packages will be installed:
    ubuntu-bionic:   fonts-lato javascript-common libgmp-dev libgmpxx4ldbl libjs-jquery
    ubuntu-bionic:   libruby2.5 rake ri ruby ruby-dev ruby-did-you-mean ruby-full ruby-minitest
    ubuntu-bionic:   ruby-net-telnet ruby-power-assert ruby-test-unit ruby2.5 ruby2.5-dev
    ubuntu-bionic:   ruby2.5-doc rubygems-integration
    ubuntu-bionic: 0 upgraded, 20 newly installed, 0 to remove and 36 not upgraded.
    ubuntu-bionic: Need to get 8,366 kB of archives.                                                                                                                                           
    ubuntu-bionic: After this operation, 48.1 MB of additional disk space will be used.                                                                                                        
    ubuntu-bionic: Do you want to continue?                                                                                                                                                    
[...]

PR #1112, opened 2020-11-18
Fixing issues with Travis CI

zklosko:
Hopefully fixes #1092


.travis.yml
```diff
@@ -32,15 +36,9 @@ jobs:
     - language: php
       php: 7.2
       stage: test
-    - language: php
-      php: 7.1
```
paddatrapper, line 36: PHP 7.1 is provided in Ubuntu Bionic, so we need to retain this test

.travis.yml
```diff
@@ -32,15 +36,9 @@ jobs:
...
-      stage: test
     - language: php
       php: 7.0
       stage: test
-    - language: php
-      php: 5.6
```
paddatrapper, line 42: PHP 5.6 is provided in CentOS 7, which we currently support. We need to have the discussion about continuing that support, but I'll open an issue for that

paddatrapper: #1113 1113

paddatrapper:
Please create a new PR for the docs stuff

zklosko:
It's not docs stuff, it's getting Jekyll to run in our CI tools

paddatrapper:
I was referring to #1111, which you talk about in the PR body

zklosko:
Okay. Can they be together just this once, since they're similar issues?

paddatrapper:
It is very useful to have 1 PR for each bug or feature. That way we can easily roll-back changes if there are issues without affecting other things. Additionally it makes creating the changelog much easier as we can go through the PRs that have been merged since the last release and include them

paddatrapper:
By the way, if you haven't seen it yet and would like to join, there will be a virtual project meet up on 25 November 2020 @ 19:00 UTC - https://discourse.libretime.org/t/online-libretime-meetup-25-november-2020-19-00-utc/713

paddatrapper:
Depending on how things go, we can look at replacing PHP 7.3 tests with PHP 7.4. May be a bit too soon though (too many test errors that make the run useless to us anyway)

zklosko:
As far as I understand, Xenial uses PHP 7.0 and Bionic uses PHP 7.2. Focal uses PHP 7.4? Why test the others?

paddatrapper:
The PHP 7.4 test would, in theory, give us an indication if things were going to break before a stable release of a distro we support ships with it. Obviously in our case this doesn't happen as the tests are broken without a chance of fixing them... The only real solution is to move off Zend and we're back at the core issue of LibreTime maintenance - the large steaming pile of technical debt we have to try get rid of.

The current PHP support matrix is this:

| Distro   | Release         | PHP Version |
|-----------|--------------------|------------------|
| Debian | Buster (10)     | 7.3               |
| Ubuntu | Xenial (16.04) | 7.0               |
| Ubuntu | Bionic (18.04) | 7.2               | 
| CentOS| 7.3                  | 5.4              |

Source:

https://packages.ubuntu.com/xenial/php

https://packages.debian.org/buster/php

Also looking at that I realise that my earlier review was a mistake, and you can remove the 7.1 test, sorry.

zklosko:
It's all good. It sounds like we're really hankering for a rewrite.

paddatrapper:
Looks like there are two issues at the moment:

Python - failing because Bionic has Silan version 0.3.3, which is known to segfault and not work. We probably need to install it from the libretime PPA instead. The docs on how are here: https://docs.travis-ci.com/user/installing-dependencies/#installing-packages-from-a-custom-apt-repository

PHP - failing because Bionic doesn't include the required PHP versions. The supported versions on Travis are (https://docs.travis-ci.com/user/reference/bionic#php-support):

- 7.1
- 7.2
- 7.3
- 7.4

Out of those versions, only 7.1 will actually run our tests correctly. I'm in two minds about what the solution should be. We could just drop all the other versions and only test on 7.1, or we could include all and ignore failures. The former would make tests run faster, however the latter is more for a potential future where our tests run on PHP >= 7.2... The chance of that happening, however is extremely thin

zklosko:
Would it be fair to assume we'd drop support for Xenial in the near future?

Also, I'll add the PPA to Travis. The PHP tests will probably be irreverent after our rewrite, no? It also seems like using a Javascript-based framework would make it easier on all of us.

paddatrapper:
We'll drop Xenial support in April 2021 in line with Ubuntu's support for it.

They will be irreverent eventually, yeah. The issue is that the rewrite may take a while

zklosko:
The Python 3.5 test passed! According to my Ubuntu Bionic instance, I'm running Python 3.6.9.

zklosko:
I'm still getting `pyenv: python version not installed` errors when the prior `python --version` command clearly shows the correct version of Python is installed...

paddatrapper:
Have a look at #1092 for how I was trying to address that

paddatrapper:
If you've got the correct python versions already installed you can remove all the pyenv commands from travis/install.sh

paddatrapper:
Python builds are failing: https://travis-ci.org/github/LibreTime/libretime/jobs/744893394

zklosko:
My guess is that you're missing these under *addons:apt:packages*:

      - gcc
      - gir1.2-gtk-3.0
      - python3-gi
      - python3-gi-cairo
      - python-cairo
      - pkg-config

paddatrapper:
That build was from this PR with [those lines included](https://github.com/LibreTime/libretime/pull/1112/files#diff-6ac3f79fc25d95cd1e3d51da53a4b21b939437392578a35ae8cd6d5366ca5485R30-R35)

zklosko:
Sorry, I was confused. The test you're referencing appears to be before I added your code to this PR, in which Python works but the plugins don't install.

paddatrapper:
That build is of commit https://github.com/LibreTime/libretime/pull/1112/commits/ada9dec45aa921dc185f6550d92f53a2783968ca which is 'Removing pyenv instructions'. Travis matrix: https://travis-ci.org/github/LibreTime/libretime/builds/744893391?utm_source=github_status&utm_medium=notification

Unfortunately we can't test this locally. Travis is painfully slow... My solution would be to move to GitLab! (I am joking, not serious)

zklosko:
But we're just testing the plugins' installation, right? Are we making them run any other tests?

paddatrapper:
We can run the tests locally, yes, but the issue isn't the tests, it is the infrastructure set up to run the tests that we need to fix here

zklosko:
I think the main issue we're having is the right packages aren't being installed on Travis when we run the tests. I've been getting "missing package" errors when I test locally; adding the missing packages lets the tests run.

zklosko:
It appears only PHP 5.6 and 7.0, and Python 3.5 will pass on Xenial, and I'm speculating only Python 3.6 will pass on Bionic (due to PHP issues with Zend/Propel). Is there a point to run the other tests if we know they won't pass due to OS compatibility?

I know we're not ready for 20.04 LTS yet, but I do at least think we should press on to support Bionic for the 3.0 release and drop support for Xenial. This way, we're running two tests (PHP 7.2 and Python 3.6) instead of seven, and every test we run is relevant to the OS we're supporting.

paddatrapper:
If we set up tests on Bionic, we only have the 1 working test (thanks to PHP failing). So I think, until we can replace Zend, we're stuck testing on Xenial. Yeah, I think we can drop the other tests that are known failure cases. The failures are just extending run time without giving us any extra insight.

I have never tried Debian on Travis... Could see how that goes, if it is possible to run under Debian Buster...

zklosko:
Are we allowed to test on multiple OSes on Travis for free?

zklosko:
If only... but it doesn't look like Debian testing is an option with Travis: https://docs.travis-ci.com/user/for-beginners/#infrastructure-and-environment-notes

zklosko:
@paddatrapper Is this a good spot to merge?

paddatrapper:
LGTM!

PR #1115, opened 2020-11-19
Removing legacy Selenium tests

zklosko:
This PR removes files in `tests/` and `airtime_mvc/tests/`, both used with legacy testing with Selenium.


paddatrapper:
You've got the Travis commits in here too

PR #1116, opened 2020-11-19
Removing Selenium tests

zklosko:
This PR removes the legacy testing done with Selenium.


airtime_mvc/tests/application/models/database/ScheduleDbTest.php
```diff
@@ -1,138 +0,0 @@
-<?php
```
paddatrapper, line 1: This isn't a selenium test...

Issue #1041, opened 2020-05-13
Airtime Analyzer ReplayGain fails with Python 3.5


paddatrapper:
Calling `subprocess.check_output()` with `text=True` was only added in Python 3.7:

https://github.com/LibreTime/libretime/blob/0c5c8fae5238eabe7da73a2490535976205f2d3a/python_apps/airtime_analyzer/airtime_analyzer/replaygain_analyzer.py#L23-L24

Before then `universal_newlines` was used: https://docs.python.org/3.7/library/subprocess.html#subprocess.check_output

rjhelms:
Per the python docs, both `text` and `universal_newlines` are present in Python 3.7 and up as aliases of each other, so changing this call to use `universal_newlines` instead should be compatible with newer versions while resolving this issue in python 3.5 and 3.6.

I don't have a proper Libretime dev environment set up to test thoroughly, but making this change and reinstalling Libretime on my staging environment seems to resolve this issue with Ubuntu 18.04 LTS / Python 3.6.

paddatrapper:
@rjhelms can you submit a PR with that fix? I'm happy to review, test and merge

rjhelms:
@paddatrapper Apologies for the delay in getting to this - PR has been submitted.

paddatrapper:
Thanks!

PR #1117, opened 2020-11-19
Fix replaygain_analyzer in Python 3.5

rjhelms:
'text' argument was added in Python 3.7 as an alias of 'universal_newlines'. Using 'universal_newlines' works on Python 3.5/3.6 and should be forward-compatible.


rjhelms:
Fix for issue #1041

Issue #988, opened 2020-03-27
iframe tag attribute has no double quotes


gagzzz:
**Describe the bug**
It's a tiny bug, on the public page of libretime, the player seems to be loaded in an iframe, but the src attribute doesn't use double quotes as it should.

**To Reproduce**
1. install a fresh libretime
2. fetch its frontpage
3. look at its source

**Expected behavior**
src=url should be src="url"

**LibreTime version:**
3 alpha 8

**Installation method:**
 - OS: ubuntu
 - OS Version 18.04
 - Method: `./install` script
 - Details: [how did you call the install script, where did you get packages from]

**Screenshots**

valerio-bozzolan:
(GitHub ghosts? Apologies for such rows.)

Here to say that with [Border Radio](https://border-radio.it/) we had enough time to fix this issue with a pull request. Let's see when it will be merged. Happy hacking!

https://github.com/LibreTime/libretime/pull/1119

PR #1119, opened 2020-11-28
Fix HTML iframe tag in EmbedController(s) and adopt relative URLs

valerio-bozzolan:
This changeset fixes this issue:
* https://github.com/LibreTime/libretime/issues/988

This changeset contributes in fixing this issue
* https://github.com/LibreTime/libretime/issues/957

### DIFF
--- a/airtime_mvc/application/views/scripts/index/index.phtml
+++ b/airtime_mvc/application/views/scripts/index/index.phtml
@@ -34,7 +34,7 @@
     <?php endif; ?>
 
     <div id="tab-1" class="schedule tab_content current">
-        <iframe onLoad="autoResize('schedule_iframe');" id="schedule_iframe" height="300px" scrolling="yes" frameborder="0" src="<?php echo $this->stationUrl."embed/weekly-program?style=premium"?>"></iframe>
+        <iframe onLoad="autoResize('schedule_iframe');" id="schedule_iframe" height="300px" scrolling="yes" frameborder="0" src="/embed/weekly-program?style=premium"></iframe>
     </div>
 
 
@@ -299,7 +299,7 @@
     }
     ?>
 
-    <iframe id="player_iframe" frameborder="0" width="100%" style="bottom:0px; left:0px; position:fixed; right:0px;" src=<?php echo $this->stationUrl."embed/player?stream=auto&style=premium";?>></iframe>
+    <iframe id="player_iframe" frameborder="0" width="100%" style="bottom:0px; left:0px; position:fixed; right:0px;" src="/embed/player?stream=auto&style=premium"></iframe>
 
 </div>
 
--- a/airtime_mvc/application/controllers/EmbedController.php
+++ b/airtime_mvc/application/controllers/EmbedController.php
@@ -25,18 +25,18 @@ class EmbedController extends Zend_Controller_Action
 
         $request = $this->getRequest();
 
-        $this->view->playerhtml5_js = Application_Common_HTTPHelper::getStationUrl() . "js/airtime/player/playerhtml5.js?".$CC_CONFIG['airtime_version'];
-        $this->view->jquery = Application_Common_HTTPHelper::getStationUrl() . "js/libs/jquery-1.10.2.js";
-        $this->view->metadata_api_url = Application_Common_HTTPHelper::getStationUrl() . "api/live-info";
+        $this->view->playerhtml5_js = "/js/airtime/player/playerhtml5.js?".$CC_CONFIG['airtime_version'];
+        $this->view->jquery = "/js/libs/jquery-1.10.2.js";
+        $this->view->metadata_api_url = "/api/live-info";
         $this->view->player_title = json_encode($this->view->escape($request->getParam('title')));
-        $this->view->jquery_i18n = Application_Common_HTTPHelper::getStationUrl() . "js/i18n/jquery.i18n.js?";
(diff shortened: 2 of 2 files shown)


paddatrapper:
This PR doesn't include a fix for the quoting of the iframe: 

```
<iframe id="player_iframe" frameborder="0" width="100%" style="bottom:0px; left:0px; position:fixed; right:0px;" src=http://localhost:8080/embed/player?stream=auto&style=premium></iframe>
```

paddatrapper:
The other thing to consider here is that the embeddable widgets, like the schedule, can be embedded into sites that aren't the LibreTime instance. How are relative URLs handled within iframes? Also `getStationUrl()` obeys 'force_ssl` in the config file, so that can be used to ensure it uses SSL even if the config has a different port configured

valerio-bozzolan:
> This PR doesn't include a fix for the quoting of the iframe:
> 
> ```
> <iframe id="player_iframe" frameborder="0" width="100%" style="bottom:0px; left:0px; position:fixed; right:0px;" src=http://localhost:8080/embed/player?stream=auto&style=premium></iframe>
> ```

Uhm. This also fixes the double-quotes in the iframe's `src=""`. Probably it does not fix what do you mean and I've missed something. What do you mean instead?

valerio-bozzolan:
Edited: I missed something. I will git commit --force to avoid two commits.

airtime_mvc/application/views/scripts/index/index.phtml
```diff
@@ -57,7 +57,7 @@
         <div id="tab-4" class="podcast tab_content">
             <div class="rss-title-wrapper">
                 <div id="rss-subscribe">
-                    <a target="_blank" href="<?php echo $this->stationPodcastRssUrl ?>" class="rss_icon"><span><?php echo _("Subscribe") ?></span><img src="css/radio-page/img/rss.png"></a>
+                    <a target="_blank" href="<?php echo $this->stationPodcastRssUrl ?>" class="rss_icon"><span>Subscribe </span><img src="css/radio-page/img/rss.png"></a>
```
paddatrapper, line 60: Why did you remove the translation call here?

airtime_mvc/application/views/scripts/index/index.phtml
```diff
@@ -70,7 +70,7 @@
             <div id="jp_container_1" class="jp-audio">
                 <div class="jp-type-playlist">
 
-                    <div class="current-track track-overflow"><?php echo _("Now Playing:") ?><span class="track-overflow"></span></div>
+                    <div class="current-track track-overflow">Now Playing: <span class="track-overflow"></span></div>
```
paddatrapper, line 73: And the translation call here?

airtime_mvc/application/views/scripts/index/index.phtml
paddatrapper, line 60: Why did you remove the translation call here?

airtime_mvc/application/views/scripts/index/index.phtml
paddatrapper, line 73: And the translation call here?

airtime_mvc/application/views/scripts/index/index.phtml
```diff
@@ -355,4 +355,14 @@
 
     });
 
-</script>
+</script><link href="/css/radio-page/custom.css" rel="stylesheet" type="text/css" />
```
paddatrapper, line 358: Did you mean to add this? Considering it is Google Analytics, please remove it from the PR

paddatrapper (changes requested):

valerio-bozzolan:
The problem is I'm editing directly from your production Docker container to see if everything works (`/usr/share/airtime`, not versioned) and committing in top of your master assuming they were the same version.

I will fix ASAP.

valerio-bozzolan:
OK I'm fine filling GitHub with trash.
Now this commit is OK from my perspective.

Thanks for your review!

airtime_mvc/application/views/scripts/index/index.phtml
valerio-bozzolan: Unintended.

Fixed in https://github.com/LibreTime/libretime/pull/1119/commits/b8f4a3392716b5be2ab89139cc5b65b973dec523.

airtime_mvc/application/views/scripts/index/index.phtml
valerio-bozzolan: Unintended.

Fixed in https://github.com/LibreTime/libretime/pull/1119/commits/b8f4a3392716b5be2ab89139cc5b65b973dec523.

airtime_mvc/application/views/scripts/index/index.phtml
valerio-bozzolan: Unintended.

Fixed in https://github.com/LibreTime/libretime/pull/1119/commits/b8f4a3392716b5be2ab89139cc5b65b973dec523.

paddatrapper:
LGTM, thanks!

hairmare:
Thanks for cleaning up this ❤️

proto+host relative URLs make much more sense.

In the long run we'd also like to make it easy to be hosted on subpath without using a base tag but that's gonna be part of the refactor to vue & a drf based api first approach!

valerio-bozzolan:
> Thanks @valerio-bozzolan for cleaning up this heart
> 
> proto+host relative URLs make much more sense.
> 
> In the long run we'd also like to make it easy to be hosted on subpath.

Good luck.

Issue #1109, opened 2020-11-16
LibreTime support for Ubuntu 20.04 LTS (Focal Fossa)


valerio-bozzolan:
At the time of writing the LibreTime's installer supports Ubuntu 18.04 (Bionic Beaver) but not Ubuntu 20.04 LTS (Focal Fossa).

About Ubuntu 20.04 LTS (Focal Fossa): this is the latest long-term supported Ubuntu distribution and it will have official support for 5 years until April 2025.

This issue is about landing to this new distribution (where _new_ means 7 months ago at the time of writing).

## What to do

1. [ ] Try Propel v.2 and fix migration issues
2. [ ] package Liquidsoap in Ubuntu (bonus point!)

The issues with 20.04 are listed here: #1024.

## Notes

Another reason is that Liquidsoap and its plugins are not in Ubuntu's official repositories. This is maybe caused by lack of support of these packages since Debian GNU/Linux 10 (buster). The Liquidsoap's plugins were dropped since Debian GNU/Linux 9 (stretch):

* https://packages.debian.org/jessie/liquidsoap-plugin-gstreamer (Debian 8: OK)
* https://packages.debian.org/stretch/liquidsoap-plugin-gstreamer (Debian 9: OK)
* https://packages.debian.org/buster/liquidsoap-plugin-gstreamer (Debian 10: missing)

Anyway I think we can overcome this lack and we may be able to land on Ubuntu 20.20 LTS (Focal Fossa) before 2025.

paddatrapper:
The issues with 20.04 are listed here: #1024. The biggest problem is PHP 7.4 in 20.04, which Propel (which we use for database stuff) doesn't support. We need to migrate to Propel v2 to fix that

zklosko:
What if we moved towards a NoSQL database, like CouchDB or MongoDB? I'm not too familiar with MongoDB, but I've been poking around in CouchDB and it uses API calls and JSON for pretty much all database operations.

valerio-bozzolan:
This is still strategic until 2025. No need to close.

paddatrapper:
I've pinned it so that it doesn't get marked as stale

PR #1024, opened 2020-05-06
[WIP] Adding support for Ubuntu 20.04 and updating zf1s for Php 7.4 support

Robbt:
So I'm in the process of testing this. It looks like we need to update our zf1s libraries to use 1.13 where the maintainer did some fixes so that it will work with Php 7.4. We had already been using master for a couple of them. 

I'm still in the process of testing this via vagrant. And I'll remove the WIP when I can confirm it is working without further changes.

PR #1120, opened 2020-11-28
Add unofficial support for Ubuntu 20.04 LTS (Focal Fossa)

valerio-bozzolan:
Please note that LibreTime currently does *not* support Ubuntu 20.04 LTS but this may be a good starting point.

Read the official website to read official news and discover when LibreTime will really work on Ubuntu 20.04 LTS (Focal Fossa).

Additional details:
* https://github.com/LibreTime/libretime/issues/1109
* https://github.com/LibreTime/libretime/pull/1024


paddatrapper:
PHP 7.4 isn't supported, and judging by the progress upstream in Propel, it won't be. So I'm not sure how useful this PR would be. Installing PHP 7.3 from a PPA could work (depending on the quality of the PPA)

zklosko:
@paddatrapper Do we know Libretime works with PHP 7.3?

paddatrapper:
> Do we know Libretime works with PHP 7.3?

Yes it does. That is what Debian Buster ships

install
```diff
@@ -738,10 +744,24 @@ if [ "$ignore_dependencies" = "f" ]; then
         if [ "$in_place" = "t" ]; then
             loudCmd "DEBIAN_FRONTEND=noninteractive apt-get -y -m install git"
         fi
+
+        # Ubuntu 20.04 Focal Fossa is missing some dependencies
+        # https://github.com/LibreTime/libretime/issues/1109
+        if [[ "$is_ubuntu_focal" -eq "true" ]]; then
+            echo
+            echo "ERROR: This distribution probably does not package Liquidsoap."
```
zklosko, line 752: Ubuntu Focal does package Liquidsoap: https://packages.ubuntu.com/focal/liquidsoap

installer/lib/requirements-ubuntu-focal.apt
```diff
@@ -0,0 +1,55 @@
...
+python3-cairo
+rabbitmq-server
+silan
+sysvinit-utils
+unzip
+vorbisgain
+vorbis-tools
+vorbis-tools
+xmlstarlet
+zip
```
zklosko, line 55: We had to add a few extra packages to make the Cairo package install work:

```
gcc
gir1.2-gtk-3.0
libcairo2-dev
libgirepository1.0-dev
python3-cairo
python3-gi
python3-gi-cairo
pkg-config
```

install
zklosko: Sadly, Focal does not appear to include these in its repo:

```
liquidsoap-plugin-alsa
liquidsoap-plugin-ao
liquidsoap-plugin-faad
liquidsoap-plugin-flac
liquidsoap-plugin-icecast
liquidsoap-plugin-lame
liquidsoap-plugin-mad
liquidsoap-plugin-ogg
liquidsoap-plugin-portaudio
liquidsoap-plugin-pulseaudio
liquidsoap-plugin-taglib
liquidsoap-plugin-voaacenc
liquidsoap-plugin-vorbis
```

zklosko (changes requested):

install
zklosko:
```
sudo apt install liquidsoap-plugin-vorbis
Reading package lists... Done
Building dependency tree       
Reading state information... Done
Package liquidsoap-plugin-vorbis is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source
However the following packages replace it:
  liquidsoap

E: Package 'liquidsoap-plugin-vorbis' has no installation candidate
```

Does this mean we no longer need the extra plugins?

paddatrapper: Yes. The plugins are no longer separate packages

zklosko:
@paddatrapper for the upstream PPA: https://launchpad.net/%7Eondrej/+archive/ubuntu/php/+index?batch=75&memo=150&start=150

hairmare:
I guess we'll want to support Focal Fossa for a while since it's an LTS release. AFAICT An update to the liquidsoap scripts is also needed for Focal Fossa.

The liquidsoap change in #1138 targets 1.4.3 and Focal Fossa has 1.4.1 so we might need to tweak the scripts some more if we are affected by the fallibility change on `cross`/`crossfade` in 1.4.2.

paddatrapper:
To be honest I think we should close this without merging - it doesn't add support, just reports the lack thereof. Support won't come any time soon unfortunately (unless the API and UI rewrite happens is miraculously fast). Lack of support should be announced in the documentation, not in the installer in my opinion

valerio-bozzolan:
> To be honest I think we should close this without merging - it doesn't add support, just reports the lack thereof. Support won't come any time soon unfortunately (unless the API and UI rewrite happens is miraculously fast). Lack of support should be announced in the documentation, not in the installer in my opinion

Well, while the documentation should note this lack of support, the installer can try to go in the right direction, even if partial.

I'm unsubscribing from this since sadly I have not time to contribute anymore on this goal. I've adopted a Docker image based on a legacy supported distribution. Have a nice day and thanks!

hairmare:
IMO I'd prefer we merge it. It isn't quite finished but we can for sure use it as a development environment to get working on Focal Fossa. I've always seen the Vagrant images as more of a bunch of development environments that do not perfectly indicate what distros we support but rather what distros are being used to work on LibreTime.

paddatrapper:
Extended with some more error reporting, a Vagrant image and PHP 7.3 from Ondrej's PPA in #1168

Issue #1123, opened 2020-11-30
Bug: Looming Rate Limit on Travis-CI


hairmare:
**Describe the bug**
We will most likely be affected by [travis-ci only offering a limited free tier](https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing) going forward.

**To Reproduce**
1. Use travis-ci with the new free tier for a bit
2. Run into limit
3. Be frustrated

**Expected behavior**
Not to run into rate limiting when contributing to LibreTime.

**LibreTime version:**
CI/CD related.

**Additional context**
Travis recently introduced these limits.

IMHO the obvious solution is a refactor towards using [GitHub Actions](https://github.com/features/actions).

I put the 3.0-release-blocker tag on this since we don't have a generic release blocker label and I'd like to get this fixed before we do the next alpha or beta point release.

**Related**
* #1112 (ooff, I didn't even realize)

paddatrapper:
+1 for Github actions. They also run much more quickly, so we can get feedback without waiting the 24hra or so that Travis takes to run.

Robbt:
Sounds like a good plan to me.

zklosko:
It turns out Github Actions' free plan isn't much better than Travis's. It might actually be worse: https://docs.github.com/en/free-pro-team@latest/github/getting-started-with-github/githubs-products.

paddatrapper:
That document says 2000 action minutes for an organisation, which is double the 1000 for travis (as I understand it). The major advantage I see is the faster run times

zklosko:
Ah, okay. I see now.

PR #1122, opened 2020-11-30
feat(ui): results of vue create ui using vue2 defaults

hairmare:
For now these are the raw results of `vue create ui` in the root while choosing the new Vue 3 default.

## Tasks
* [x] rework to actual LibreTime specific scaffolding
* [ ] update docs proper to embed in LibreTime proper
* [x] yarn vs. npm decision (currently uses yarn because ¯\_(ツ)_/¯)
* [ ] ensure renovate or dependabot for `ui/`
* [x] add css and js assets to legacy mvc stack as part of "strangling" the old views
* [ ] #518 as PoC widget 
* [ ] implement playouthistory page in vue
* [ ] replace favicon/assets 😛 
* [ ] squash this PR for clean history
* [x] storybook scaffolding
* [x] jest scaffolding
* [x] working storybook-docs
* [ ] get rid of `[Vuetify] Unable to locate target [data-app]` warning in tests
* [ ] run jest tests through GitHub Actions
* [ ] finalize api and rebase onto final api variant (needs some cleanup in the playouthistory & playouthistorytemplate endpoints)
* [ ] working i18n conversion script (most packages I found don't do exactly what we need)

Fixes #1121

PR #1125, opened 2020-12-01
Moving from Travis CI to Github Actions

zklosko:
Addresses @hairmare's request in #1123. I have scripts for testing on both Ubuntu Xenial and Bionic; testing on Debian or CentOS with Github Actions is not available at this time. 🙄

Note: it only runs when pushed to the master branch or a PR is opened/marked for review relative to the master branch.

Fixes: #1123


paddatrapper:
Will this run on updates to a PR too?

.travis.yml
```diff
@@ -1,97 +0,0 @@
...
-            dos2unix
-      script:
-        ./travis/release.sh
-      deploy:
-        - provider: releases
-          token: $GITHUB_TOKEN
-          file: build/libretime-*.tar.gz
-          edge: true
-          on:
-            tags: true
```
hairmare, line 74: I think we'll also want to add a workflow that covers these release bits which where responsible for uploading a fat tarball (ie. with packages php deps) to the GitHub releases page.

.github/workflows/test-bionic.yml
```diff
@@ -0,0 +1,30 @@
+name: test-bionic
+on:
+  push:
+    paths-ignore:
+    - 'docs/**'
+  pull_request:
```
zklosko, line 6: @paddatrapper It should. That can be customized here.

.travis.yml
zklosko: Of course. How often should it run?

paddatrapper: On tags on master

.github/workflows/test-bionic.yml
paddatrapper: OK cool

.travis.yml
zklosko: On *any* tag, or just tags that include "v." or versioning information?

paddatrapper: Any tags or a regex for `\d+\.\d+\.\d+(~\s+\d+)*` (I think, that is untested)

zklosko: > \d+\.\d+\.\d+(~\s+\d+)*

I don't understand that at all. Sorry. 😢 

Do tags *only* get applied to releases?

paddatrapper: Actually it should be `\d+.\d+.\d+(~\s+\.\d+)*`. It should match things like 3.0.0~alpha.5, but yeah, we only creat tags for releases. It also matches things like 3.0.0

.github/workflows/package-release.yml
```diff
@@ -13,6 +14,12 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@v2
+      - name: Read file contents
```
paddatrapper, line 17: We don't have a release notes file, we use the draft releases page to add the changes and release notes

zklosko: I don't think we can automate that.

zklosko: I don't think we can automate it that way. What this will do is make a release *and* compile the gzp archive every time we push with a tag that matches your specifications.

paddatrapper: Ok.What format should the RELEASENOTES file take?

zklosko: Markdown

paddatrapper: And the entire content is copied into the release notes for the tag?

zklosko: That's my understanding, yes

.travis.yml
hairmare: We're not currently using a v prefix so just running it on every tag would be fine IMO.

If we change anything related to tagging we can refactor the workflow to use a pattern then.

.github/workflows/package-release.yml
paddatrapper: Ok. Then I think lets put that file somewhere out of the way, maybe under `.github`. @hairmare, @Robbt what are your thoughts?

hairmare: `action/create-release` does seem to create the actual release which is already done the way I've been releasing in the past. AFAICR there are some existing standards regarding how to format `CHANGELOG.md` that might make sense for us.

I would prefer to keep the actual releasenotes human written though. I've been thinking of moving them to the wiki to make way for autogenerated notes (PR based) in the actual github release. 
A RELEASENOTES.md would work as well but I don't like having to nag people that they should also mention their contribution in the file... maybe we should do consider something this [nag](https://github.com/ponylang/release-notes-reminder-bot-action) action?

We could also use semantic commits and integrate with a semantig release tool that automatically creates tags when we merge to master but IMO this is something to look into after we reach at least beta.

paddatrapper: I also like the handwritten release notes. Perhaps the nag action is a good idea. Or adding the current milestone to all merged PRs if that can be automated. Then we can do the same as now and go through the merged PRs to create the release notes

zklosko: We can still write our own release notes under the automated method; the Github Action itself doesn't generate the RELEASENOTES file, just pulls from it (like `cat file > newfile` or `myvariable=${cat file}`).

I'm not aware of a plugin that would pull from a Github Wiki but I can keep digging. Keeping basic explainers in static files instead of a CMS or database sounds better to me.

zklosko:
@paddatrapper @hairmare Any more thoughts on moving to GH Actions?

.github/workflows/package-release.yml
paddatrapper: I'm happy to merge this with release notes moved to .github/worrkflows. The nag bot is useful, but doesn't need to be part of this PR

zklosko: @paddatrapper Well, this PR also removed the files we need to use Travis CI with. While the tests are running, the Python plugins do have failing tests (which GH Actions still labels as a successful test). Below output is from a test on Ubuntu 18.04:

```
Run sudo bash ./.github/scripts/python-pkg-test.sh
  sudo bash ./.github/scripts/python-pkg-test.sh
  shell: /bin/bash -e {0}
  env:
    pythonLocation: /opt/hostedtoolcache/Python/3.6.12/x64
    LD_LIBRARY_PATH: /opt/hostedtoolcache/Python/3.6.12/x64/lib
..........................................................FFFFFF...
======================================================================
FAIL: tests.replaygain_analyzer_tests.test_mp3_utf8
----------------------------------------------------------------------
  File "/usr/local/lib/python3.6/dist-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/runner/work/libretime/libretime/python_apps/airtime_analyzer/tests/replaygain_analyzer_tests.py", line 33, in test_mp3_utf8
    check_default_metadata(metadata)
  File "/home/runner/work/libretime/libretime/python_apps/airtime_analyzer/tests/replaygain_analyzer_tests.py", line 19, in check_default_metadata
    assert abs(metadata['replay_gain'] - expected_replaygain) < tolerance
AssertionError: 
-------------------- >> begin captured stdout << ---------------------
5.67

--------------------- >> end captured stdout << ----------------------

======================================================================
FAIL: tests.replaygain_analyzer_tests.test_mp3_dualmono
----------------------------------------------------------------------
  File "/usr/local/lib/python3.6/dist-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/runner/work/libretime/libretime/python_apps/airtime_analyzer/tests/replaygain_analyzer_tests.py", line 38, in test_mp3_dualmono
    check_default_metadata(metadata)
  File "/home/runner/work/libretime/libretime/python_apps/airtime_analyzer/tests/replaygain_analyzer_tests.py", line 19, in check_default_metadata
    assert abs(metadata['replay_gain'] - expected_replaygain) < tolerance
AssertionError: 
-------------------- >> begin captured stdout << ---------------------
5.66

--------------------- >> end captured stdout << ----------------------

======================================================================
FAIL: tests.replaygain_analyzer_tests.test_mp3_jointstereo
----------------------------------------------------------------------
  File "/usr/local/lib/python3.6/dist-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/runner/work/libretime/libretime/python_apps/airtime_analyzer/tests/replaygain_analyzer_tests.py", line 43, in test_mp3_jointstereo
    check_default_metadata(metadata)
  File "/home/runner/work/libretime/libretime/python_apps/airtime_analyzer/tests/replaygain_analyzer_tests.py", line 19, in check_default_metadata
    assert abs(metadata['replay_gain'] - expected_replaygain) < tolerance
AssertionError: 
-------------------- >> begin captured stdout << ---------------------
5.67

--------------------- >> end captured stdout << ----------------------

======================================================================
FAIL: tests.replaygain_analyzer_tests.test_mp3_simplestereo
----------------------------------------------------------------------
  File "/usr/local/lib/python3.6/dist-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/runner/work/libretime/libretime/python_apps/airtime_analyzer/tests/replaygain_analyzer_tests.py", line 48, in test_mp3_simplestereo
    check_default_metadata(metadata)
  File "/home/runner/work/libretime/libretime/python_apps/airtime_analyzer/tests/replaygain_analyzer_tests.py", line 19, in check_default_metadata
    assert abs(metadata['replay_gain'] - expected_replaygain) < tolerance
AssertionError: 
-------------------- >> begin captured stdout << ---------------------
5.66

--------------------- >> end captured stdout << ----------------------

======================================================================
FAIL: tests.replaygain_analyzer_tests.test_mp3_stereo
----------------------------------------------------------------------
  File "/usr/local/lib/python3.6/dist-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/runner/work/libretime/libretime/python_apps/airtime_analyzer/tests/replaygain_analyzer_tests.py", line 53, in test_mp3_stereo
    check_default_metadata(metadata)
  File "/home/runner/work/libretime/libretime/python_apps/airtime_analyzer/tests/replaygain_analyzer_tests.py", line 19, in check_default_metadata
    assert abs(metadata['replay_gain'] - expected_replaygain) < tolerance
AssertionError: 
-------------------- >> begin captured stdout << ---------------------
5.67

--------------------- >> end captured stdout << ----------------------

======================================================================
FAIL: tests.replaygain_analyzer_tests.test_mp3_mono
----------------------------------------------------------------------
  File "/usr/local/lib/python3.6/dist-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/runner/work/libretime/libretime/python_apps/airtime_analyzer/tests/replaygain_analyzer_tests.py", line 58, in test_mp3_mono
    check_default_metadata(metadata)
  File "/home/runner/work/libretime/libretime/python_apps/airtime_analyzer/tests/replaygain_analyzer_tests.py", line 19, in check_default_metadata
    assert abs(metadata['replay_gain'] - expected_replaygain) < tolerance
AssertionError: 
-------------------- >> begin captured stdout << ---------------------
5.69

--------------------- >> end captured stdout << ----------------------

----------------------------------------------------------------------
Ran 67 tests in 5.804s

FAILED (failures=6)
...........
----------------------------------------------------------------------
Ran 11 tests in 0.157s

OK
```

paddatrapper: Ah, this looks like Silan issues. Maybe install Silan from the PPA?

paddatrapper: Also failing tests should equal a failed CI run...

zklosko: The install script should be grabbing Silan from the PPA... I asked it to...

> Also failing tests should equal a failed CI run...

Exactly. Maybe we need to force the test to exit with a different exit code upon an error or failure.

zklosko:
Okay, so now I have package-release function working, but I'm having problems with the database in testing.

```
psql: error: could not connect to server: No such file or directory
	Is the server running locally and accepting
	connections on Unix domain socket "/var/run/postgresql/.s.PGSQL.5432"?
```

GH Actions uses Docker containers for testing. I'm not familiar with Docker testing with Postgre, so the commands to create the libretime user don't appear to be running. It doesn't appear to make the tests fail. More tests are running than before.

I still haven't found a way to raise an error or a non-zero exit code upon failures in nosetests. We might have to make a custom bash script for it.

paddatrapper:
Typically docker doesn't have an init system, so it could be that you need to run things manually? Not sure though

zklosko:
Yeah, I didn't think I'd have to do that. Upon checking the logs, it declined to install postgres because it was already installed... But then the service wasn't running..? The VMs are supposed to start fresh each time...

As nice as Docker can be, I prefer testing on vanilla VMs 🤷🏻‍♂️

paddatrapper:
If it didn't start that is most likely because GH installed the packaged version with a systemd/sys v init script, which won't work because no init system.

zklosko:
How would it be initialized then?

paddatrapper:
Seems you create a service container: https://docs.github.com/en/free-pro-team@latest/actions/guides/creating-postgresql-service-containers

zklosko:
Do our testing scripts know they need to be looking for an external Docker container?

paddatrapper:
Probably not. The connection details would need to be modified in the `/etc/airtime/airtime.conf` file for tests I would expect

zklosko:
That file doesn't appear to be populated by the testing scripts.

zklosko:
@paddatrapper I have ~two~ three questions: does the testing even pull from a database at all? Also, what is the point of installing the PHP Composer dependencies on each test? Do the Python packages need external access to the PHP Composer dependencies?

Also, we're getting a lot of errors testing on Xenial; would it be okay to drop Xenial tests moving forward? I know I asked about this earlier but it just doesn't seem needed. We just seem to be doing a lot of unnecessary testing.

paddatrapper:
>  does the testing even pull from a database at all?

I don't know. I don't think the python ones do, but I could be wrong. If all the tests pass without a DB, then I would assume that they do not.

> Also, what is the point of installing the PHP Composer dependencies on each test? Do the Python packages need external access to the PHP Composer dependencies?

The PHP tests will need the PHP dependencies, but the Python ones do not

> Also, we're getting a lot of errors testing on Xenial; would it be okay to drop Xenial tests moving forward? I know I asked about this earlier but it just doesn't seem needed. We just seem to be doing a lot of unnecessary testing.

What errors are you getting? If they are errors that need fixing so that users can use LT on Xenial, then we need to fix the errors and not drop the tests

zklosko:
The file for the PHP testing may or may not have been deleted in a previous PR...

zklosko:
@paddatrapper The version of PHPUnit we're using to test our PHP setup (v.4) isn't compatible with PHP 7.0 or 7.2. I'm not entirely sure what needs to be done to upgrade everything, but we're going to need to rewrite our testing setup in order to get this to work.

paddatrapper:
Why did this not cause issues on PHP 7.0 on Travis I wonder...?

zklosko:
No idea.

zklosko:
Some slightly good news: the database initializes correctly now. The PHP testing configuration still needs work and possibly a new test suite.

.github/RELEASE.md
```diff
@@ -0,0 +1,177 @@
...
+
+The full tarball for the 3.0.0-alpha.8 release of LibreTime is available here.
+
+Since this is an alpha release there will be bugs in the code.
+
+Please report new issues and/or feature requests in the issue tracker. Join our discourse or chat us up on Slack if you need help and for general discussion.
+
+## Table of Contents
+
+    Features
```
hairmare, line 11: Should all of these be internal links or should we remove the ToC?
```suggestion
    [Features](#Features)
```
(I didn't test linking as in the suggestion)

.github/RELEASE.md
```diff
@@ -0,0 +1,177 @@
...
+## Deprecated Features
+
+    We are phasing out support for Debian 8 (Jessie). See the announcement in the 3.0.0-alpha.7 release notes for more details.
+    Liquidsoap 1.1.1 support. 3.0.0-alpha.8 is most likely the last version to support liquidsoap 1.1.1 out of the box. The number of distros that install a current version of liquidsoap are gaining the majority and once Ubuntu releases a version of their distro that has liquidsoap 1.3.x we will switch to native liquidsoap 1.3.x support. Users still on liquidsoap 1.1.1 will need to apply a patch to their liquidsoap scripts (or update liquidsoap).
+
+## Contributors
+
+The LibreTime project wants to thank the following contributors for adding PRs to this release:
+
+    @ciaby
```
hairmare, line 53: Maybe this list can be automated?

.github/scripts/install-bionic.sh
```diff
@@ -0,0 +1,35 @@
+#/bin/bash
```
hairmare, line 1: Maybe these should be their own PR to make this one smaller.

.github/scripts/python-pkg-install.sh
```diff
@@ -0,0 +1,6 @@
+#/bin/bash
+
+pip3 install -e python_apps/airtime_analyzer/.
+pip3 install -e python_apps/airtime-celery/.
+pip3 install -e python_apps/api_clients/.
+pip3 install -e python_apps/pypo/.
```
hairmare, line 6: are we gonna add more `python_apps`? Maybe it make sense to automate installing them on GH?
```suggestion
for app in `ls python_apps`; do
  pip3 install -e python_apps/$app
done
```

.github/workflows/package-release.yml
hairmare: btw... I can live with the release parts not working after we merge this, for now our focus should probably be on running tests... I can do the next release manually and/or fix what's not working then if that makes sense.

hairmare (commented):
I took a quick look at this and am braindumping my findings... There are some tasks in #1122 that I'd rather look into after we merge this ;)

.github/scripts/python-pkg-install.sh
zklosko: Automating sounds good to me. I was just copying over what we had so far. With only four apps, it's just one more line of code to write. 🤷

.github/scripts/install-bionic.sh
zklosko: I disagree. Getting dependencies right is part of setting up testing.

.github/RELEASE.md
zklosko: Specifically the contributors?
If automated in general, we could look into something like this: https://github.com/marketplace/actions/release-notes-generator.

.github/RELEASE.md
zklosko: This was a copy/paste of our previous release notes document; links probably won't be needed if we edit the content down to something like 300 words max.

zklosko:
```
../../vendor/bin/phpunit: line 2: ?php: No such file or directory
../../vendor/bin/phpunit: line 3: /bin: Is a directory
../../vendor/bin/phpunit: line 4: application: command not found
../../vendor/bin/phpunit: line 5: application: command not found
../../vendor/bin/phpunit: line 6: syntax error near unexpected token `c'
../../vendor/bin/phpunit: line 6: ` * (c) Sebastian Bergmann <sebastian@phpunit.de>'
Error: Process completed with exit code 2.
```

paddatrapper:
Where are the logs for the run?

zklosko:
https://github.com/zklosko/libretime/runs/1571149085?check_suite_focus=true

paddatrapper:
It looks like something is interpreting a php file as bash...

zklosko:
@paddatrapper Duh 🤦 

All tests are running now. We're getting two errors: one upstream with Nosetests and a second with our PHP testing script. The PHP test appears to be creating the same database at least 10 times, and on one run it fails.

Results here: https://github.com/zklosko/libretime/actions/runs/430690438

paddatrapper:
The last [run] (https://github.com/zklosko/libretime/runs/1577830268?check_suite_focus=true) looks to me like three code failures.

The first is a non numeric value encountered in testIsFileScheduledInTheFuture. 

 The second failure with the date time objects looks like it should compare the absolute value of the difference between them to ensure that it is within a range (|a-b|<0.1 or something)

The third looks like it is not inserting the expected values into the DB for a repeated show

zklosko:
I was kinda hoping someone more well versed in PHP could take a look at it...

hairmare:
```
There was 1 error:

1) ScheduleUnitTest::testIsFileScheduledInTheFuture
A non-numeric value encountered

/home/runner/work/libretime/libretime/vendor/zf1s/zend-log/library/Zend/Log.php:636
/home/runner/work/libretime/libretime/airtime_mvc/application/models/Preference.php:768
/home/runner/work/libretime/libretime/airtime_mvc/application/controllers/ApiController.php:859
/home/runner/work/libretime/libretime/airtime_mvc/tests/application/models/unit/ScheduleUnitTest.php:60
```
This looks like it's triggering an edge case that indicates an actual bug.

```

There were 2 failures:

1) PreferenceUnitTest::testSetShowsPopulatedUntil
Failed asserting that two DateTime objects are equal.
--- Expected
+++ Actual
@@ @@
-2020-12-21T09:15:21.000000+0000
+2020-12-21T09:15:21.908977+0000

/home/runner/work/libretime/libretime/airtime_mvc/tests/application/models/unit/PreferenceUnitTest.php:18
```
PHP 7 introduced higher precision timestamps per default. The current test only worked because it was generating the date twice in the same second and there where always edge cases where it could fail due to the time being generated before ad after rolling over to the next second.
```
2) ShowServiceDbTest::testCreateWeeklyRepeatRRShow
Failed asserting that 
+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+
| cc_show                                                                                                                                                                                                                                                                                                                                                                                              |
+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+
|          id          |         name         |         url          |        genre         |     description      |        color         |   background_color   | live_stream_using_ai | live_stream_using_cu |   live_stream_user   |   live_stream_pass   |        linked        |     is_linkable      |      image_path      |   has_autoplaylist   |   autoplaylist_id    | autoplaylist_repeat  |
+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+
|          1           |      test show       |         NULL         |         NULL         |         NULL         |        ffffff        |        364492        |                      |                      |         NULL         |         NULL         |                      |          1           |                      |                      |         NULL         |                      |
+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+
+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+
|   '4' != actual 6    | '2044-01-08 00:00:00 | '2044-01-08 01:00:00 |          1           |   '1' != actual 0    |   '0' != actual 1    |   NULL != actual 4   |                      |
+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+
|   '5' != actual 7    | '2044-01-09 00:00:00 | '2044-01-09 01:00:00 |          1           |   '0' != actual 1    |   '1' != actual 0    |  '4' != actual NULL  |                      |
+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+
|   '6' != actual 8    | '2044-01-10 12:00:00 | '2044-01-10 13:00:00 |          1           |          0           |          1           |   '4' != actual 7    |                      |
+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+
|   '7' != actual 9    | '2044-01-15 00:00:00 | '2044-01-15 01:00:00 |          1           |   '1' != actual 0    |   '0' != actual 1    |   NULL != actual 7   |                      |
+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+
|   '8' != actual 10   | '2044-01-16 00:00:00 | '2044-01-16 01:00:00 |          1           |   '0' != actual 1    |   '1' != actual 0    |  '7' != actual NULL  |                      |
+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+
|   '9' != actual 11   | '2044-01-17 12:00:00 | '2044-01-17 13:00:00 |          1           |          0           |          1           |   '7' != actual 10   |                      |
+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+
|  '10' != actual 12   | '2044-01-22 00:00:00 | '2044-01-22 01:00:00 |          1           |   '1' != actual 0    |   '0' != actual 1    |  NULL != actual 10   |                      |
+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+
|  '11' != actual 13   | '2044-01-23 00:00:00 | '2044-01-23 01:00:00 |          1           |   '0' != actual 1    |   '1' != actual 0    | '10' != actual NULL  |                      |
+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+
|  '12' != actual 14   | '2044-01-24 12:00:00 | '2044-01-24 13:00:00 |          1           |          0           |          1           |  '10' != actual 13   |                      |
+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+
|  '13' != actual 15   | '2044-01-29 00:00:00 | '2044-01-29 01:00:00 |          1           |   '1' != actual 0    |   '0' != actual 1    |  NULL != actual 13   |                      |
+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+
|  '14' != actual 16   | '2044-01-30 00:00:00 | '2044-01-30 01:00:00 |          1           |   '0' != actual 1    |   '1' != actual 0    | '13' != actual NULL  |                      |
+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+
|  '15' != actual 17   | '2044-01-31 12:00:00 | '2044-01-31 13:00:00 |          1           |          0           |          1           |  '13' != actual 16   |                      |
+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+
|  '16' != actual 18   | '2044-02-05 00:00:00 | '2044-02-05 01:00:00 |          1           |   '1' != actual 0    |   '0' != actual 1    |  NULL != actual 16   |                      |
+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+
|   '17' != actual 1   | '2044-02-06 00:00:00 | '2044-02-06 01:00:00 |          1           |   '0' != actual 1    |   '1' != actual 0    | '16' != actual NULL  |                      |
+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+
|   '18' != actual 2   | '2044-02-07 12:00:00 | '2044-02-07 13:00:00 |          1           |          0           |          1           |   '16' != actual 1   |                      |
+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+----------------------+


+----------------------+----------------------+----------------------+----------------------+
| cc_show_rebroadcast                                                                       |
+----------------------+----------------------+----------------------+----------------------+
|          id          |      day_offset      |      start_time      |       show_id        |
+----------------------+----------------------+----------------------+----------------------+
|          1           |        1 days        |       00:00:00       |          1           |
+----------------------+----------------------+----------------------+----------------------+
|          2           |        2 days        |       12:00:00       |          1           |
+----------------------+----------------------+----------------------+----------------------+


+
| cc_show_hosts |
+
|
+

.

/home/runner/work/libretime/libretime/airtime_mvc/tests/application/services/database/ShowServiceDbTest.php:485
```
This does smell like it's related to how the tests initialize the database. It looks like the `id` field on `cc_show_instances` isn't getting reset to start with 1 at some point before this test.

I'll take a closer look later this week to see if I can figure out what is going on with the DB tests.

hairmare:
`ScheduleUnitTest::testIsFileScheduledInTheFuture ` can be fixed by using a fixed date rather than "now" as in https://github.com/radiorabe/libretime/commit/7020c7d17b10ca938cf3df919d391604528a30bb. I'm not sure why this also fixes `PreferenceUnitTest::testSetShowsPopulatedUntil` but it does.

hairmare:
I didn't figure out what is up with ShowServiceDbTest yet, but was able to stomp the other two bugs:

* https://github.com/radiorabe/libretime/commit/98101fe44da0cd10af3d141f2a0a381bef3b0ca5
* https://github.com/radiorabe/libretime/commit/ba125ca6824f2b41d9551491b2b0a1329141ad45

zklosko:
It mentions this file: `/home/runner/work/libretime/libretime/airtime_mvc/tests/application/services/database/ShowServiceDbTest.php:485`

```
$this->assertDataSetsEqual(
    new PHPUnit_Extensions_Database_DataSet_YamlDataSet(__DIR__ . "/datasets/test_createWeeklyRepeatRRShow.yml"),
   $ds  // Specifically this line
);
```

hairmare:
I was able to reproduce the `ShowServiceDbTest::testCreateWeeklyRepeatRRShow` failure but didn't figure out why it's not working so far :(

zklosko:
Could it have anything to do with the same database being created over and over again without being destroyed between creations?

hairmare:
> Could it have anything to do with the same database being created over and over again without being destroyed between creations?

I tried resetting the DROPing and CREATEing the database on each test run but all I got was slower test runs and the same issue.

hairmare:
The still breaking test is due to the fact that DBUnit isn't supported anymore. I haven't been able to track down how exactly in DBUnit to fix it (and thus now fully understand why it is't maintained anymore).

I'll look into how we can skip the affected tests without loosing too much coverage. In the long run we will most likely want to cover the kind of tests that are affected by dbunit with some kind of api2/django centric snapshot testing.

zklosko:
Well, ultimately we'll be breaking from Propel and co. to use @paddatrapper's API, so I'd imagine this will be a temporary issue.

paddatrapper (changes requested):
So it seems there are three issues that need solving before this can be merged:

- [ ] Failing python test (replay gain)
- [ ] Failing PHP test - DBUnit issues, potentially skip the test
- [ ] Return a non-zero exit code when a test fails - not happening for Python at the moment

hairmare:
👍 for skipping the DBUnit test. I do think it covers something important and we'll have to put more focus on manual testing of this but the ad-hoc testing I usually do during releasing already contains some libretime handling that does relate to this. Specifically I click into the far future in the ui after programming a ton of recurring shows and hope that calender doesnt end up with overlapping shows in a couple of leap years time.

zklosko:
@hairmare @paddatrapper I'm going to be contributing less over the next few months as I work through my final semester at university, so it would be a huge help if one (or both) of you could work out the last few issues with this PR. In my own trying to remove dbunit from Composer, Composer raises multiple errors that a few Google searches haven't been able to fix. I'm not fully stepping away, I just know this is a priority and I won't be able to complete it as early as others. I can add the two of you to those who can write to my repo or whatever is easiest. Thanks for understanding.

paddatrapper:
@zklosko not a problem. Good luck with your studies!

.github/workflows/package-release.yml
paddatrapper: I've made the release creation create a draft release that we can then modify and release ourselves

.github/RELEASE.md
paddatrapper: For the moment, I suggest we leave it as is and can look at https://github.com/marketplace/actions/release-notes-generator later

.github/scripts/python-pkg-install.sh
paddatrapper: Moved to for list

paddatrapper:
I have got this PR working. It now creates draft releases on tags and fails when Python tests fail. I had to drop the Python tests from Xenial as the tests depend on mock, which is Python 3.6. The way we do testing means that we can't use GitHub's python version stuff to support testing different python versions at the moment. I also dropped the failing PHP test and fixed the Python test that was failing.

All in all I think this is now ready for merging

zklosko (commented):
Looks great to me @paddatrapper! Great work!

Robbt:
I can merge this but I'm not sure how to test and see if it is working. I'm ok with merging it as is but wanted to see if there was a way to see how it works w/o diving into github actions.

zklosko:
@Robbt Go to the repo on my profile and switch to the Actions tab: https://github.com/zklosko/libretime/actions

Robbt:
Looks good. I'll merge it.

PR #1131, opened 2020-12-16
[WIP] Editor for Libretime.

codenift:
I have a different version I started experimenting using NodeJS working side by side with this environment and setup NodeJS as a service, but that will still require more testing and a bit more complicated since we would have to change a few things in LT, would be better adding to another PR anyways. This version is using Waveform js scripts directly, not sure about performance, CPU was hitting normal at least on my computer, since I have 64GB memory so I couldn't tell of any issue there... still needs a few things changed. I also tried it with SoX for fetching waveform and passing it on, but it seems like SoX has an issue with files/songs that include latin words on the title and affects the process with those tracks so I took it off for now, plus I read someone requesting it to be fixed and the maintainer didn't have any interest in updating saying it wasn't priority.... so we can probably use something else, maybe store that data to ccfiles or maybe a new table.

Note: You will notice that turning gain knob while backend on Wavesurfer is set to MediaElementWebAudio, it won't work. If you comment out backend (Line 515 on libretime.js) then it will work. Also note, there's a lot of adjustments needed, but just leaving this here so you guys can play around with. 

I was trying to add a few more things but ran out of time. It's a start. I won't be around for 2 weeks but when I come back I'll put more work into this.


zklosko:
😮 🍾 🎸 

Hot dog! That looks amazing!

In quick testing I'm finding the gain knob isn't working the way I anticipated it to. I see it starts with the value assigned from the import plugin but adjusting the value makes the volume jump. Also, the red and green carets (`<` and `>`) don't appear to do anything; the track's cue points don't change (maybe I'm using it wrong).

Using "School" by Calvin Harris for testing:

- Let's say Libretime Analyzer detects it needs to adjust gain on a track by -11.04
- Adjusting the gain knob a little bit makes the volume significantly decrease (adjust to -9.04)
- Returning the knob to -11.04 makes the volume even softer
- Adjusting the gain to 0.00 makes the track sound more like what we started with (and the readout on the meter agrees)

But that new editor looks kick-ass. What would make it even better is if it could edit multiple files simultaneously instead of opening new files is new tabs. Being able to adjust the album art is definitely a plus; feels more like iTunes back in its heyday.

codenift:
Thanks. Yes, at the moment the gain knob is not synced to work how it should or even talking with the meter itself. Not yet. Still incomplete, but Robbt wanted me to make a PR of it so you guys can play around with it. So the red and green button, I left it there but it is not linked to anything right now, but the plan is to set the cues "start" and "end" even while it plays, disabled if trying to set the end before start. 

Yes, we had talks about multi-editing feature.. eventually we'll get to that. This is just the first steps for the editor... I have a few more ideas for it and hopefully inspire other devs to improve more on it. I will put more work on it soon.

PR #1139, opened 2020-12-26
fix: update to CentOS 8

### LINKED ISSUE #1113: CentOS support is out of date
We currently support CentOS 7.3. The current CentOS 7 release is 7.9 and CentOS 8 has been released. I think we need to re-evaluate our CentOS support. It is not first class in the installer (users have to manually install the packages required before starting the install) and there has been no work on getting CentOS 8 support. CentOS is currently the only reason we still support PHP 5 and I think it would be a good idea to drop that requirement if we drop CentOS support

hairmare:
Update CentOS Vagrant environment to CentOS 8. The changes to the database routines work on `.deb` based platforms as well, everything else is specific only to the `centos.sh` Vagrant init script.

Fix #1113

### DIFF
--- a/install
+++ b/install
@@ -698,12 +698,12 @@ verbose "Using distribution id \"$dist\", release code \"$code\""
 systemInitDetect
 
 if $is_centos_dist; then
-    python_bin="python3"
+    python_bin="python3.8"
     apache_bin="httpd"
     apache_service="httpd"
     web_user="${web_user:-apache}"
 else
-    python_bin="python"
+    python_bin="python3"
     apache_bin="apache2ctl"
     apache_service="apache2"
     web_user="${web_user:-www-data}"
--- a/install
+++ b/install
@@ -624,6 +624,7 @@ is_ubuntu_bionic=false
 is_ubuntu_xenial=false
 is_centos_dist=false
 is_centos_7=false
+is_centos_8=false
 # Use specified distribution and release or detected otherwise.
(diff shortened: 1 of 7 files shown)


install
```diff
@@ -697,10 +698,12 @@ verbose "Using distribution id \"$dist\", release code \"$code\""
 systemInitDetect
 
 if $is_centos_dist; then
+    python_bin="python3"
     apache_bin="httpd"
     apache_service="httpd"
     web_user="${web_user:-apache}"
 else
+    python_bin="python"
```
paddatrapper, line 706: This should probably be python3 everywhere. I think I manually replaced things with python3 in the script elsewhere

paddatrapper: Seeing as we don't support python 2

hairmare: Look back at this I have no clue how I introduced it... I'll do some testing based on the Debian vagrant machines before I fix it.

hairmare: I guess [PEP 394](https://www.python.org/dev/peps/pep-0394/) got me confused. CentOS now uses 3.8 since I'll need that to install psycopg2 for the api later on. Debian is untouched and uses `python3` again.

paddatrapper: Debian systems are unlikely to ever use `python` for `python3`. It may be offered through alternatives, but the Python team's plan (as I recall) is to have `python2` and `python3` and not install a `python` binary at all.

Issue #1133, opened 2020-12-16
Default stream endpoint changes when stream settings saved


paddatrapper:
In both the **Listen** popup and the **Streams** settings menu, the URLs for the streams are incorrectly displayed.

- **Streams**: `http://192.168.xx.xx:8000/_a` instead of `http://192.168.xx.xx:8000/airtime_128` (default setup)
- **Listen**: `127.0.0.1 Libretime Stream #1` instead of `serverIP ...` (default setup)

This is likely messing up the stream URL that's getting sent to the widgets. I've been digging for the part of the PHP code that either sets/retrieves the stream URL data but I haven't been able to find it.

_Originally posted by @zklosko in https://github.com/LibreTime/libretime/issues/461#issuecomment-745399022_

I have noticed this too. The initial stream endpoint is /airtime_128 in icecast (and so liquidsoap), but the default shown in the interface (so I'm assuming somewhere in the DB) is /_a. This is all synchronized to /_a when you press Save on the stream settings for the first time, but before then there are inconsistencies, leading to the above

zklosko:
I just hit a major breakthrough in figuring out how to pass PHP variables to Vue! I can use this to come up with a band-aid fix while we figure out what's setting the URL incorrectly.

zklosko:
Update: turns out I can't... the variable isn't placed in the browser and is calculated from the database. We would need to change the database value. 😞 

I've started by starting a reverse proxy in Apache, routing `IP:8000/` to `IP/streams/` (meaning the default stream would show up on `IP/streams/airtime_128`). This would allow us to call URLs on the same port as the web app. At the moment, this still causes display issues (and the player on the `IP:8000/` page on the redirect doesn't work) but the stream `IP/streams/streamURL` does work. Might be an upstream issue with Icecast.

zklosko:
@paddatrapper @Robbt @hairmare How do you feel about switching players? I found a Javascript library for creating players in HTML5 (no Flash option) called [Howler.js](https://github.com/goldfire/howler.js). Instead of needing to use a CDN or package manager, it only needs one file to run. An example player:

```
var sound = new Howl({
  src: ['stream1', 'stream2', 'stream3'],
  autoplay: true,
  volume: 0.5,
});
```

paddatrapper:
I do think having the dependency in a dependency manager is a good thing. That looks interesting, but I think the discussion should happen in #249

zklosko:
[...]

Issue #1121, opened 2020-11-30
Proposal: Initialize Vue Env


hairmare:
**Is your feature request related to a problem? Please describe.**
This is part of getting #2 fixed. There seems to be consensus that our next frtend will be based on [Vue](https://vuejs.org/). This issue helps track the efforts needed to get our Vue dev infra up and running.

**Describe the solution you'd like**
I would like to initialize Vue development by putting scaffolding in place and configuring CI/CD.

This shall first be used to create individual Vue components that we can embed in the existing frontend-stack. In the long run it should be the foundation of replacing the frontend routing parts of the legacy ZF1 stack while we migrate to the new Django/Django-Rest-Framework based API from #1056.

While we are at it we sould probably also introduce [storybook.js](https://storybook.js.org/docs/vue/get-started/introduction). I still need to decide if this is in the scope of this issue (primarily scaffolding) or if should be split into it's own.

**Describe alternatives you've considered**
I did consider surprising everyone with a working prototype but going down the proposal route is a much more serious way to have at this.

**Additional context**
* #2 (is what this helps solving)
* #1056 (describes the new API and is very much related)
* #1093 (as i plan on using this as a proving ground for the first vue module)
* **TBD** add much more issues that this can/will touch this or should be affected by it

hairmare:
I just added a task re us choosing a design framework/component lib in the Vue space.

Personally we've been focusing on [Vuetify](https://vuetifyjs.com/) @radiorabe lately so for me that is a prime candidate. Are there any other preferences out there? Is Material fine or do we need something with less structure/more control?

paddatrapper:
I like Vuetify, I've used it before. I do suggest we try match the existing style as closely as we can without having to create custom elements. For example use the dark varients of Vuetify elements, etc

zklosko:
If there's still time, can I put in a vote for [Bulma](https://bulma.io/)? It's a gorgeous Bootstrap alternative that is based around CSS (no JavaScript included).

Now that I'm looking at Vuetify, those table add-ins look amazing (albeit very "Google"). I was thinking we were going to try to use DataTables again and try to do the rest from scratch but those color and date pickers are winning me over.

zklosko:
![Screen Shot 2020-12-05 at 4 39 01 PM](https://user-images.githubusercontent.com/31484611/101266032-92b55700-3719-11eb-9971-72c04174e4b9.png)

paddatrapper:
That looks great!

[...]

Issue #518, opened 2018-09-16
Playout history flash buttons are not executed


dosmanak:
Hello. On **playout history** there are three buttons that should execute flash script to copy/csv/pdf export.
I can not run any button both from google chrome and firefox .
I am not sure how to debug it. When I allow flash on the page I can see puzzle button, but nothing happens when I click it on network view nor console. There is not any request made to server so I can not check logs.

Perhaps I am doing something wrong or my installation is wrong. But perhaps someone could test it?
I am at commit 4508d0d8c0f660f6b5449d4d25cb00ea5efc97d4

BTW Is there a reason for flash?

Thank you.

Robbt:
Honestly I am not sure if this code is complete. This was written before we forked the project to LibreTime so it will require some detective work to determine the status of these features.

hairmare:
It did work at some point. From what I remember it isn't working anymore because it tries to use outdated Flash features that any sane, modern browser will block.

I think the reason for Flash here was to allow copying to the clipboard (which couldn't be done natively at the time). Personally I feel a csv download should be enough for any use cases. 

Pull-Requests to refactor this are very much welcome ;)

kmahelona:
Aloha. I’ve had a quick look at this and it appears that the DataTables js package is used to generate the CSV files. I think it’s as simple and pulling the data from the html. This fix will require updating DataTables as the version used is now deprecated. I will have a look at this and do a pull request. We are running Airtime 2.5.1 and we need playout history as we’re a public broadcaster and have reporting requirements. I’ll have to get this bug fixed before I can move us to LibreTime.

Robbt:
Yeah you are write, DataTables is pretty much the UI for LibreTime and so updating it will be a pain, partially because there are a number of hacked plugins and extensions that modify the functionality. It is one of these points in the codebase where we might be better off rebuilding it with an alternative to datatables. It is hard to say for sure, it is one of our biggest instances of technical debt. 

So the issue I think isn't that playout history doesn't exist but that it isn't generating an easy to export CSV correct ? We can probably fix that without overhauling datatables completely.

hairmare:
I've been down the rabbit-hole that is updating datatables (without success).

One of my motivating factors at the time was that modern versions of datatables have support for using HTML5 APIs instead of just Flash in their export bits.

zklosko:
Bumping this because this is still an issue that needs to be fixed, especially by the time Flash Player is no longer supported (this December).

[...]

PR #1140, opened 2020-12-27
Removing TableTools, fixing default stream mount

zklosko:
In this PR, I've removed the TableTools Flash export buttons and functionality, fixed the default logo display, and fixed the default stream mountpoint. At the moment, there is no easy drop-in fix for the Playout History export buttons, but the possibility to create a custom exporter using Vanilla JavaScript and plugin libraries remains. Updating DataTables to 1.10 from 1.9.x completely breaks the interface.

~This PR also removes the podcast player from the microsite (https://github.com/LibreTime/libretime/issues/1121#issuecomment-751420795).~

~Some PHP rewriting was done as well, removing unnecessary echo functions.~

Issues: #1133 #518

### Screenshots

Screen Shot 2020-12-26 at 5 33 57 PM
Screen Shot 2020-12-26 at 5 37 05 PM
Screen Shot 2020-12-26 at 5 36 42 PM


airtime_mvc/application/views/scripts/index/index.phtml
```diff
@@ -306,21 +270,12 @@
...
             $("#player_iframe").contents().find('.bottom_bar').append("<div class='about_us_btn button' data-tab='tab-2'></div>");
-
-            <?php
-            // Is there a better way to do this?
-            if ($this->displayRssTab) {?>
-                $("#player_iframe").contents().find('.bottom_bar').append("<div class='station_rss_btn button' data-tab='tab-4'></div>");
-            <?php
-            }
-            ?>
+            $("#player_iframe").contents().find('.bottom_bar').append("<div class='station_rss_btn button' data-tab='tab-4'></div>");
```
hairmare, line 278: `displayRssTab` is still getting set in `IndexController`:
https://github.com/LibreTime/libretime/blob/6c4b1b5e71f8de260a1b04ee3f04c7489b1832bd/airtime_mvc/application/controllers/IndexController.php#L83
It is managed in `PreferenceController`:
https://github.com/LibreTime/libretime/blob/6c4b1b5e71f8de260a1b04ee3f04c7489b1832bd/airtime_mvc/application/controllers/PreferenceController.php#L111
It looks like this change will remove the ability to hide the rss "tab" or am I missing something any you found "a better way to do this"?

airtime_mvc/application/views/scripts/index/index.phtml
```diff
@@ -43,261 +23,245 @@
...
-        <div id="tab-4" class="podcast tab_content">
-            <div class="rss-title-wrapper">
-                <div id="rss-subscribe">
-                    <a target="_blank" href="<?php echo $this->stationPodcastRssUrl ?>" class="rss_icon"><span><?php echo _("Subscribe") ?></span><img src="css/radio-page/img/rss.png"></a>
-                </div>
-                <div id="rss-title">
-                    <?php echo $this->podcastTitle ?>
-                </div>
+    <div id="tab-4" class="podcast tab_content">
+        <div class="rss-title-wrapper">
```
hairmare, line 27: Not checking `displayRssTab ` seems to break when stations do not have a public podcast (see above for details).

airtime_mvc/application/views/scripts/login/index.phtml
```diff
@@ -1,12 +1,7 @@
 <div class="login_box">
-    <div class="logobox">&nbsp;</div>
-        <h2><?php echo _("Login") ?></h2>
+    <div class="logobox"></div>
+        <h2>Login</h2>
     <div id="login" class="login-content clearfix">
-        <?php if(isset($this->demo) && $this->demo == 1){?>
```
hairmare, line 5: It feels right to remove the demo code. We should remove the correponding code in `LoginController` as well.
https://github.com/LibreTime/libretime/blob/6c4b1b5e71f8de260a1b04ee3f04c7489b1832bd/airtime_mvc/application/controllers/LoginController.php#L100-L102

hairmare (changes requested):
We can re-add export possibilities once we switch to the vue frontend (what I'm preparing for playouthistory in #1122).

Other than some small i18n issues and a few bits of dead code this is looking good so far.

airtime_mvc/application/views/scripts/index/index.phtml
```diff
@@ -1,40 +1,20 @@
...
-                <span><?php echo _('Login'); ?></span>
-                <span class='login-img'></span>
-            </a>
-        </div>
-    <?php endif; ?>
+    ?>
+
+    <div class='login-btn'>
+        <a href='/login' target='_parent'>
+            <span>Login</span>
```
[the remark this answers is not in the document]
zklosko: I completely agree, but how is this being translated? I didn't see a i8 something call attached to this one.

airtime_mvc/application/views/scripts/index/index.phtml
zklosko: Thanks for mentioning it was in the controllers. There doesn't appear to be a setting for it anywhere in the interface, so I removed the logic for it.

airtime_mvc/application/views/scripts/index/index.phtml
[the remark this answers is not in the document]
hairmare: I think it gets translated to the os default language which is english in most cases (the lang switcher on the ui happens after this gets render afaict).

airtime_mvc/application/views/scripts/index/index.phtml
hairmare: it's this thing on `/podcast/station`:

image

airtime_mvc/application/views/scripts/index/index.phtml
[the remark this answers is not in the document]
zklosko: So the php echo call is still needed in all cases?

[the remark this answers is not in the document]
hairmare: Yeah it calls the `_()` function which does the actual translating. Using the underscore as a function name for this is a de-facto standard when working with gettext.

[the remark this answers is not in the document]
zklosko: Gotcha. Thanks for filling me in! I'll work on putting it back now.

airtime_mvc/application/views/scripts/index/index.phtml
zklosko: I meant to only remove the extra login form 🤦

airtime_mvc/application/views/scripts/index/index.phtml
zklosko: Gotcha

airtime_mvc/application/views/scripts/embed/player.phtml
```diff
@@ -109,7 +109,7 @@
 
         Html5Player.prototype.play = function() {
             console.log('play');
-            playerhtml5_audio.src = this.settings.url+'?'+Math.floor(Math.random() * Math.floor(100000));
+            playerhtml5_audio.src = this.settings.url; //+'?'+Math.floor(Math.random() * Math.floor(100000));
```
hairmare, line 112: I think this was intended as a cache buster but it looks broken anyway.

hairmare (dismissed):

paddatrapper:
I can't test this thanks to #1142. @hairmare did you have a chance to test this?

hairmare:
> I can't test this thanks to #1142. @hairmare did you have a chance to test this?

I roughly tested the micro-site player and that the default stream still works on a fresh install. I wasn't able to test the "Playout History" changes since the branch I'm currently on has way to much unrelated changes to that controller/view.

hairmare:
I figured out what the resizing code is supposed to do. It takes care of the popup from the listen button underneath the on air sign.

With this change the window looks a bit to large:

image

With the code working it should get resized to something like this:

image

I'm pretty sure that there are better ways to acheive this nowadays, but I haven't touched anything related to popups in modern HTML/JS recently so I'm not sure what the optimal solution would be here.

zklosko:
Interesting... I'm also using Firefox and I haven't had a problem with the resizing.

zklosko:
I really screwed my branch up. I'm going to start from scratch and make another PR when I'm ready. 😢

PR #1144, opened 2020-12-30
refactor(airtime_analyzer): rename to libretime-analyzer and make entrypoint pythonic

hairmare:
* [x] rename `airtime_analyzer` binary to `libretime-analyzer`
* [x] rewrite `.rst` docs to `.md` in `python_apps/libretime_analyzer`
* [x] move dev dependencies to `requirements-dev.txt` to get them out of `setup.py`
* [x] remove `--no-init-script` (and get rid of the init script installing feature all together, neither the .deb, .rpm nor the `./install` method used the feature)

I tried to touch only the minimaly required files to get the file renamed and to also cleanup setup.py a bit. IMHO with this change it's a bit more pythonic.


paddatrapper:
How well does this support upgrading from an existing install with `airtime_analyzer` installed?

install
```diff
@@ -782,7 +782,7 @@ if [ -f /etc/airtime/airtime.conf ]; then
     
         set +e
         verbose "Stopping airtime services..."
-        systemInitCommand stop airtime_analyzer airtime-celery airtime-playout airtime-liquidsoap airtime-media-monitor
+        systemInitCommand stop libretime-analyzer airtime-celery airtime-playout airtime-liquidsoap airtime-media-monitor
```
paddatrapper, line 785: This will break upgrades where the new `libretime-analyzer` systemd unit has not been installed yet. Having all of these use the legacy aliases ensures that the first upgrade will work. After 1 release we can migrate them and require users to install alpha.9 before installing the next version

hairmare: Good catch! I rebased a fix.

paddatrapper (approved):
LGTM. Though holding off merge until after alpha.9

PR #1145, opened 2021-01-01
Removed HTTPS condition from source of player

samkitk:
ForceHTTPS = false ; It was causing issues on fresh install where the player would not play from the mountpoint because of https conditon.


paddatrapper:
Long term this should probably be read from the config file via an API call. This is a good solution in the mean time in my opinion

Issue #249, opened 2017-07-18
Replace Adobe Flash - It will get obsoleted and firefox complains about blocked object


xabispacebiker:
I do not know if it's the right place to submit a suggestion but I think this issue deserves it.

There is a big discussion on the internet about why we should avoid the use of Adobe Flash in our developments. I personally hate it and block any site with Adobe Flash, and this is not just my personal preference, in the near future, Flash will be obsoleted. Actually there is no need for it to get a music player. Nowadays every browser supports HTML5 audio tags. (Still we could support some back compatibility in case of users with older browser) But I would suggest to REPLACE FLASH ASAP !

I thought we were using JPlayer http://www.jplayer.org

## tracks

* [x] #518

hairmare:
What I saw so far does indicate that we use jPlayer. The radio page also loads [muses](https://www.muses.org/) so it's probably a case of using multiple things that do the same job. 🤷‍♂️ 

I can only seem to reproduce the error in Firefox. Webkit browser like Chrome and Safari don't complain and don't seem to load any swf as far as I can tell (of those two Chrome would usually complain about such things). 

The popup player from the backend works in Firefox without flash.

There is more flash lurking in the datatables part of the js where it's used to enable a print button on history lists and the likes. Updating/Rewriting datatables has proven tougher than I expected so far.

For the audio flash parts we should absolutely switch to a solution that doesn't show any popups in for end users and uses HTML5 audio as the default without triggering Firefox's flash warning. 

Personally I'd be happy to kick all flash to the curb. I'm not sure if anyone has a large user base that still relies on legacy browsers that need flash for audio.

OliCat:
Hello,

I've imported the new mrp.js hosted here [https://hosted.muses.org/mrp.js](url) and added the parameter `'forceHTML5':true` in `/usr/share/airtime_mvc/application/views/scripts/embed/player.phtml`

`            this.settings = {
                'volume': 100,
                'forceHTML5': true,
                'jsevents': true,
                'autoplay': false,
                'buffering': 0,
                'title': 'test',
                'bgcolor': '#FFFFFF',
                'skin': 'mcclean',
                'reconnectTime' : 2000, //Doesn't seem to do anything
                'width': 180,
                'height': 60
            };`

With Firefox, it's ok, no more complains about flash module, the player work.
With Safari and chrome, the player is broken :-(

`TypeError: undefined is not an object (evaluating 'musesPlayer.play')`

[...]

PR #1150, opened 2021-01-02
Refocus: Cleaning up libraries

zklosko:
This PR hopes to start 2021 with some (slightly early) Spring Cleaning by removing unused libraries and files.

Removals include

- TableTools (incl. Flash parts, exporting in Playout History)
- Widgets hidden from main interface
- References to Airtime Pro in the interface and code for Airtime Pro monitoring features
- Extra CSS
- Extra uncalled JS scripts

Doesn't fully address #249 yet, but gets us halfway there.

Update: I've decided to change goals. Instead of focusing on converting objects to Vue, I plan to focus on simplifying the JavaScript objects and libraries being used in order to make maintenance and rewriting to Vue easier. I'd like this to be considered before we release 3.0.0 alpha 9.


paddatrapper:
How does this affect #1122?

zklosko:
It doesn't affect #1122. If anything, it gives @hairmare something to copy/paste logic from instead of rewriting the entire thing from scratch.

Also assuming the rewrite will be a lengthy process, this will help us fix some of the issues we've had and clean up some clutter without having to continue to use outdated frameworks and libraries.

hairmare:
#1122 also contains very much config and code related to making it possible to use new Vue views in the legacy apps as drop in replacements for old interfaces w/o touching routing & logic. It's main difference from this is that it is aiming at having these new views consume data from the django-rest-framework API rather than the legacy backend.

All in all #1122 is pretty much at the state when it's gonna depend on pull requests like API 2.0 #958 and GitHub Actions #1125 soon.

zklosko:
@hairmare okay.

I noticed that my pull request was getting a little big, so I'm deciding to trim it down a little bit. The new home request is going to be focused more around removing libraries that we don't use. There are a lot of them.

I would also like to start removing muses in favor of howler.js so the players work a little bit better without a flash fallback. It would be nice if this could go in the 3.0.9 alpha release.

install
```diff
@@ -1210,6 +1210,7 @@ else
     # not on centos
     ip_device="eth0"
     $is_ubuntu_xenial && ip_device="enp0s8"
+    $is_ubuntu_bionic && ip_device="enp0s2"
```
hairmare, line 1213: I don't really Ubuntu so I can't say if this makes sense.

hairmare (commented):
I like the refocus of this 👍 It LGTM, i'll give it a final clean-room install test and then approve & merge it!

install
zklosko: It doesn't even work, TBH 😝

hairmare (approved):

PR #1153, opened 2021-01-03
feat(mvc): add preview mode setting

hairmare:
Adds a new general setting `feature_preview_mode` which I intend to use in #1122 to make it possible to opt into using vue for already converted views (#1122 has a rough version of an embeddable /playouthistory view).

There was also some discussion on [discourse#new-experimental-features/728](https://discourse.libretime.org/t/new-experimental-features/728) about introducing such a flag.


zklosko:
Does this affect what I'm working on in #1150?

hairmare:
With this you can use `Application_Model_Preference::GetFeaturePreviewMode()` in controllers to decide if you want to enable a feature. The flag can be toggled by an admin in Settings >> General.

paddatrapper:
Looks good! It needs documenting on the website though, probably under https://libretime.org/docs/settings

hairmare:
I updated the docs and rebased.

docs/_docs/settings.md
```diff
@@ -52,6 +52,8 @@ default.
 
 ![](/img/api-settings.png)
 
+You can opt-in to test new features with the **Feature Previews** option.
```
zklosko, line 55: Can you add a little more to this? Like what features we're testing and how this would affect the user. "You can switch back at any time" would be a good phrase to add too.

hairmare: I added more words :smile:

paddatrapper:
I'm happy with that. Do we want to merge this before or after alpha.9?

docs/_docs/settings.md
zklosko: Awesome

hairmare:
> I'm happy with that. Do we want to merge this before or after alpha.9?

IMO it can wait until after, it's not like there are any feature already attached to it.

PR #1158, opened 2021-01-06
fix(soundcloud): remove broken integration

### LINKED ISSUE #1043: Soundcloud Publish untested and not working
We have a button that claims to publish to Soundcloud but we have not tested this integration and I don't know what level of functionality exists here. It has been reported that it does not work by a user on the forum - https://discourse.libretime.org/t/my-podcasts-wont-pubish/559

I suggest that we validate the feature and if it isn't working we remove it before we publish a beta with a broken functionality.

hairmare:
I noticed how very broken the Soundcloud integration is when I tried to do some clean up on `python_apps/airtime-celery`. I didn't find any Discourse posts or GitHub issues of someone attempting to use the feature so I decided to remove it completely.

Fix #1043 

* [ ] remove soundcloud from architecture overview images


paddatrapper (approved):

paddatrapper:
Looks good so far. Will need to update the API in #958 when this is merged

paddatrapper (commented):
Looks good so far. Will need to update the API in #958 when this is merged

paddatrapper:
Didn't mean to approve this yet - that was supposed to just be a comment

airtime_mvc/public/js/airtime/library/library.js
```diff
@@ -236,16 +236,15 @@ var AIRTIME = (function(AIRTIME) {
...
-                "<button class='btn btn-small' id='publish-btn'>" +
-                "<i class='icon-soundcloud-white'></i>" +
-                "<span>" + $.i18n._('Publish') + "</span>" +
-                "</button>" +
-                "</div>"
-            );
-        }
+            if (onDashboard) {
+                $menu.append(
+                    "<div class='btn-group' title=" + $.i18n._('Publish') + ">" +
```
zklosko, line 241: I think this button can go entirely since the publish feature doesn't work. I can open a separate pull request to remove the My Podcast feature.

hairmare: I intentionally only remove the soudcloud bits with this PR.

I'll do another one to get rid of podcast publishing if we decide to get rid of it.

With SoundCloud I was able to just search for "soundcloud" in the code and get rid of most of what I found. Getting rid of podcast publishing will be a bit more work since the best way to find everything is with a search for "publish" and that returns a lot of false postives.

zklosko: I thought we were looking into removing the My Podcast feature because it didn't work.

Yeah, it definitely is harder. Initial attempts to remove it keep breaking the interface. 😞

PR #1160, opened 2021-01-09
Replacing Muses with Howler.js + Microsite rework

### LINKED ISSUE #249: Replace Adobe Flash - It will get obsoleted and firefox complains about blocked object

zklosko:
This PR is a rewrite of our embedded player widget and **Listen** player on the dashboard. Instead of Muses ~and jPlayer~ (which have Flash options), [Howler.js](https://howlerjs.com/) will be the new library of choice.

Users of Firefox may notice issues with caching skipping back and forth. It may be just a Firefox issue, as I don't have the same issue in Safari.

Closes #249, but still doesn't offer us exporting tools from the Analytics page.

Bonus: I've included a rework of the schedule widget that uses Handlebars.js (templating *library*) instead of AngularJS (JavaScript *framework*). The design and functionality are exactly the same as before.


zklosko:
@paddatrapper @hairmare @Robbt Were we still looking to kill the microsite?

Robbt:
I'm not inclined to kill the microsite per se unless it makes things easier. I know most stations won't use the microsite and we shouldn't focus a lot of attention on it but I don't see it as inherently broken like Soundcloud integration was or in need of removal, but that is just my take.

dmnktoe:
Maybe a small switch in the settings to turn the microsite on or off would do for now.

zklosko:
@dmnktoe That would require work on the routing. I'm not familiar with PHP routing, but if you'd like to look into it... 😉

zklosko:
Screen Shot 2021-01-14 at 11 41 50 AM

This is what I'm working on right now. It's a minor redesign of the microsite using Bulma as a CSS framework, which is easier to work with but off-brand compared to the rest of Libretime (which uses Bootstrap). The functionality is good, I just need to work with the margins and other styling and it should be done.

The "Now" and "Next" placeholders will ultimately be filled with Handlebars.JS, although they can be filled by Vue later. I've been having some issues utilizing other JavaScript libraries with PHP variables while in Vue, so I'm working in Vanilla JavaScript because it's easiest. I'm also working on removing jQuery where it's not extra helpful.

For better or for worse, it does remove the background image but it does resemble the color scheme of our interface a little more. I'll just have to work on styling the login button.

paddatrapper:
Re-writing the microsite seems like a very large feature-creep for this PR... Can it be a separate PR?

zklosko:
It could, but Howler.js never really stops loading stream data, which means the buttons don't load on the bottom bar without a redesign. It's needed for the navigation. I also don't think it's that big of a codebase change.

zklosko:
@paddatrapper I know it seems like a lot but without the changes to the microsite, the different sections won't display. What do you say?

paddatrapper:
If it is required for Howler.js to work, then I guess it is fine. The problem with a large PR that touches many things is that it is very difficult to review properly. PRs should be focused on one thing and do that one thing well

zklosko:
I understand. We just need all of the buttons to work.

zklosko:
Quick update: replacing jPlayer is not as straightforward as I anticipated. Like Datatables, jPlayer is integrated in multiple places within the interface, not just the players. As my final semester of undergrad gets underway, I'll likely slow down in contributing and won't have time to finish this anytime soon.

There are some things I was able to do successfully that I'll likely move to new PRs:

- Cleaning up jQuery libraries so we're only using jQuery 1.8
- Conversion of the weekly schedule widget to Handlebars.js + Vanilla JS instead of AngularJS

PR #1164, opened 2021-01-18
Return of Playout History Exporting!

### LINKED ISSUE #518: Playout history flash buttons are not executed

zklosko:
Behold, working HTML5 CSV and PDF export options! Uses [CSV-Export](https://github.com/mikebald/CSV-Export) and [pdfmake](http://pdfmake.org/), respectively. Closes #518 and continues work on #249.

There are additional formatting options for the PDF then are being made use of in this PR, just FYI.

Samples:
[file-1.pdf](https://github.com/LibreTime/libretime/files/5831636/file-1.pdf)
Screen Shot 2021-01-18 at 12 16 58 PM


paddatrapper:
Works well! The only thing I noticed is that by default, the list shown on the website isn't filtered by the time selector at the top (or at least I think it isn't), but the export is. This leads to a potential situation where there is a track listed on the website, but you can't export because there are no files played out in the time range selected. To reproduce:

1. Create a LT instance with no playout history.
2. Upload a track and create a show starting now
3. Schedule 1 track for the show
4. Immediately open analytics

You'll see the track listed on the website, but if you try export you get an error that there are no files in range.

Not sure if this bug is introduced in this PR or was an existing issue

zklosko:
I don't understand. Isn't Playout History past-tense, i.e. *previously* played tracks and not currently scheduled or future scheduled tracks?

Replicating the issue, it appears the pickers round to the top of each minute, so if you play A-ha's "Take On Me" at 9:26:50 a.m., and the picker is set to 9:00 a.m. - 9:26 a.m., the track won't show up because it didn't start playing between 9:00:00 - 9:26:00 a.m. I don't know if I'd call that a bug... it seems more of an intended design. If you wait another minute and refresh the page, the track shows up, even if it's still playing.

zklosko:
It's currently missing the ability to export only checked items. Does that need to be added before this gets merged?

paddatrapper:
Essentially my point is that a track will show up in the playout history Web page list, but not in the exported file, which is an inconsistent experience for users

About selective export - that can be implemented aftert this is merged

zklosko:
Hmmm... I haven't been having that issue at all. If there aren't any items in the display, the exporter tools return an error alert instead. The date/time picker in Libretime affects both the playout history view and the exported files.

paddatrapper:
When I tested using the steps I listed, I got a single entry in the Web history list, but clicking on export showed me the nothing to export popup

zklosko:
Okay, I see what you're saying now. I believe I've fixed it.

paddatrapper:
Yup, that fixed it

Issue #1191, opened 2021-04-17
Debian Buster pycairo install failed


stanlog:
**Describe the bug**
Error on Pycairo install during LibreTime install:
`Writing /tmp/easy_install-ya5mh8e_/pycairo-1.19.1/setup.cfg
Running pycairo-1.19.1/setup.py -q bdist_egg --dist-dir /tmp/easy_install-ya5mh8e_/pycairo-1.19.1/egg-dist-tmp-4yid3dt9
no previously-included directories found matching 'docs/_build'
warning: no files found matching 'README' under directory 'tests'
warning: no files found matching 'README' under directory 'examples'
cairo/device.c:30:10: fatal error: Python.h: Aucun fichier ou dossier de ce type
 #include <Python.h>
          ^~~~~~~~~~
compilation terminated.
error: Setup script exited with error: command 'x86_64-linux-gnu-gcc' failed with exit status 1`

**To Reproduce**
Launch install
`sudo bash install -fiap`

**Expected behavior**
We need install python3-dev before launching installation. After installing python3-dev, no more error. Maybe add it in the pkg list in the script ?

**LibreTime version:**
3.0.0-alpha.9

**Installation method:**
 - OS: Debian
 - OS Version: 10
 - Method: install script from git
 - Details: sudo bash install -fiap

cr3pt:
Install **python3-cairo-dev**

PR #1193, opened 2021-04-19
Install python3-dev

paddatrapper:
Fixes: #1191


Robbt:
Just looking at the diff it looks like in two places we added python3-dev to apt lists that already included it, so having python3-dev listed twice. I haven't tested it yet but this seems like an error (not sure if it will impact the code) but it probably makes sense to fix this before merging.

paddatrapper:
Right, fixed

PR #1218, opened 2021-05-29
Enhance .gitignore

jooola:
Removed some duplicate entries, and split the root gitignore in multiple ones.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -137,6 +137,7 @@ venv/
 ENV/
 env.bak/
 venv.bak/
+pyenv
 
 # Spyder project settings
 .spyderproject
--- a/.gitignore
+++ b/.gitignore
@@ -1,23 +1,163 @@
-*.*~
-*.pyc
-vendor/*
-composer.phar
-*~$
-*\.log.*
-**/*.egg-info/*
-**/build/*
-**/dist/*
-*~
-/airtime_mvc/tests/test_results.xml
-/tests/results.html
-/tests/*.jar
-VERSION
-airtime_mvc/tests/log/*.log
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -1,23 +1,163 @@
...
+
+# Flask stuff:
+instance/
+.webassets-cache
+
+# Scrapy stuff:
+.scrapy
+
+# Sphinx documentation
+docs/_build/
```
paddatrapper, line 97: Should this be moved to `docs/.gitignore`?

.gitignore
```diff
@@ -1,23 +1,163 @@
...
+
+# Environments
+.env
+.venv
+env/
+venv/
+ENV/
+env.bak/
+venv.bak/
+
```
paddatrapper, line 140: While we're listing python virtualenvs, I'd add pyenv to this list

.gitignore
jooola: I copy pasted this whole section from the Github file, so I didn't want to touch it. I would leave it as is, but I am ok with moving it in the docs/.gitignore

.gitignore
jooola: Done

.gitignore
jooola: I commented it out.

paddatrapper: I think you can remove it, just needs to be in the docs one

jooola: But we don't use sphinx documentation right ?

paddatrapper: Good point, not anymore. So actually can be removed completely

PR #1244, opened 2021-06-07
Migrate api_clients tests to pytest

jooola:
Again:
- A lot of refactor in the tests
- Small makefile for managing the api_clients subproject


.github/scripts/python-pkg-install.sh
```diff
@@ -4,6 +4,8 @@ echo "::group::Install Python apps"
 pip3 install nose mock
 
 for app in $(ls python_apps); do
+  [[ -f "python_apps/$app/requirements.txt" ]] && pip3 install -r "python_apps/$app/requirements.txt"
```
paddatrapper, line 7: Please keep this as requirements-dev.txt.

The requirements file is purely for test/development dependencies that are not required for running the application. Having it be requirements.txt makes that less clear. setup.py is the canonical place for runtime dependencies

PR #1246, opened 2021-06-07
Create dependencies manager

jooola:
This PR is an attempt to manage our install dependencies properly and to build some kind of dependencies graph.

A big problem I am having while working with this project, is that each services is not isolated from the others. *"Nobody"* knows why a package is installed, and which services requires it, as everything is packaged in single requirements list for the whole project.

I see this PR as a necessary step before many others such as building Docker images for each service (or apt packages), as I don't want to install the whole setup in each container.

The long term goal is to move towards a more decoupled infrastructure, maybe we can call it micro services. I also think some dependencies are leftovers from the past software (e.g. newer liquidsoap version doesn't rely on some the *gain tools anymore and uses ffmpeg) and could use some cleanup.
In the end each services should manager his own dependencies.

I don't know if someone already did this dependencies splitting work, would be awesome to know ?

For now I made some assumption on why some packages are installed. I want to add a few fields to the dependencies helper script, such as the **reason** why a package is install (e.g. git => in place installation), and what service **requires** it (e.g. liquidsoap => { playout: liquidsoap runtime management, analyzer: track playability analysis}).

But I will need help from people with a longer history with libretime, as I don't understand every parts of it yet.

Some package could be removed if some reason is not worth it.


paddatrapper:
Yeah, a script to do this is a great idea. We haven't looked at splitting up the dependencies before.

You could remove the `requirements-*.txt` and replace it with a piped call to the script in `install` I would think

jooola:
@paddatrapper I created something totally different from my first idea.
I think this is the best way for us to manage our packages dependencies.

There is a bit of duplication, for the liquidsoap packages for instance. But this system is simple and I don't see this getting out of hands.

In addition, this allow me to have isolated environment for each app, for CI or docker images, or apt packages.
With time (and mostly CI testing), we will be able to clean all the unused packages from the project.

jooola:
This is ready to be reviewed, and merged if we think such change can be merged before 3.0 ?

I choose to create a new scripts folder, in which we will put our "scripts" (obviously), and slowly remove the `dev_tools` and `utils` folder. I might have over engineered by adding tests to the packages script, but I think this is a good way to keep our script working.

I am quiet happy with the idea of keeping the dependencies in multiple INI file which can be properly documented.

packages.ini
```diff
@@ -0,0 +1,44 @@
...
+php7.3-mbstring = buster
+php7.3-pgsql = buster
+
+[installer]
+lsb-release = buster, bionic
+xmlstarlet = buster, bionic
+
+[unknown]
+# See https://github.com/LibreTime/libretime/commit/bbcd4fbe2ee05cdb2e3a62ad2b7549157a1cd950#diff-0991ed0aaae32d2b0c9723cd9127f95bc19d176a5fd668d9b331f487136318fe
+systemd-sysv = buster
```
paddatrapper, line 43: Considering we now ship systemd unit files and don't need the sysv init scripts to run on systemd, we should be able to remove these

install
```diff
@@ -739,16 +739,23 @@ if [ "$ignore_dependencies" = "f" ]; then
...
+      "${SCRIPT_DIR}/"
+      "${SCRIPT_DIR}/api"
+      "${SCRIPT_DIR}/python_apps/airtime_analyzer"
+      "${SCRIPT_DIR}/python_apps/pypo"
+    )
 
-    package_list="$(grep -vE '^\s*#' "$package_list_file" | tr '\n' ' ')"
+    set -e
+    package_list=$(
+      "${SCRIPT_DIR}/installer/packages.py" --format=line "${code}" "${packages_files[@]}" ||
```
paddatrapper, line 753: The script is in `${SCRIPT_DIR}/scripts/packages.py`.

packages.ini
jooola: Ok, wasn't sure what the purpose of this package was, didn't really looked it up. I'll remove it.

install
jooola: Woups.

packages.ini
jooola: What about the `sysvinit-utils` package for bionic ? I a not sure I understand it's purpose.

Is it related to sysV support ?

paddatrapper: As far as I know, it is and can be dropped too

jooola:
Waouw merged, this is great ! Thought it would take more time !

paddatrapper:
I don't see the point in waiting for 3.0 and worked in the testing I've done

PR #1248, opened 2021-06-08
Rework CI pipeline

jooola:
- Widen test matrix
- Split test workflow into multiple workflows
- Use service for PostgreSQL setup


jooola:
Sorry for spamming your mailboxes with failed pipelines notifications.

Using a service for testing PHP is a bit tricky, something is not working. I'll try to fix this in my fork when I feel I want to fight this again.

Regarding the CI workflows, I think each project should be tested in its own job, so we don't install each dependencies again and again. But this will come after I am done with #1246

Any thoughts so far ?

jooola:
It looks like the postgresql problem is more a version mismatch that a service problem, the errors are still here even without the postgres service

.github/workflows/test-php.yml
```diff
@@ -0,0 +1,67 @@
+name: Tests PHP
+
+on:
+  push:
+    paths-ignore: ["docs/**"]
```
paddatrapper, line 5: All of these should probably limit to `on.push.branches = [master]` and `on.pull_request.branches = [master]`

.github/workflows/test-python.yml
```diff
@@ -0,0 +1,80 @@
...
+            gstreamer1.0-plugins-good \
+            gstreamer1.0-plugins-bad \
+            gstreamer1.0-plugins-ugly \
+            libgirepository1.0-dev \
+            liquidsoap \
+            liquidsoap-plugin-faad \
+            liquidsoap-plugin-lame \
+            liquidsoap-plugin-mad \
+            liquidsoap-plugin-vorbis \
+            python3-gst-1.0 \
```
paddatrapper, line 47: This will install the Python package version for the system Python and not the Python under test

.github/workflows/test.yml
```diff
@@ -1,21 +1,11 @@
-name: Python and PHP Tests
+name: Tests
```
paddatrapper, line 1: Maybe call this `Lint` and the file `test-lint.yml` because it doesn't actually run any code tests

.github/workflows/test-python.yml
```diff
@@ -0,0 +1,80 @@
...
+      - uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      # TODO: Extract apt dependencies in dependencies manager
+      #       and only install for analyzer
+      - name: Install dependencies
+        run: |
+          if [[ "${{ matrix.runs-on }}" ==  "ubuntu-18.04" ]]; then
+            sudo add-apt-repository -y ppa:libretime/libretime
```
paddatrapper, line 33: This is only needed for upgrading out-dated silan versions. So should also need to be run on `ubuntu-16.04`.

.github/workflows/test.yml
jooola: I don't know, I expect to run pylint somewhere at some point and it would obviously go into lint as the name suggests it, but we only want to run linting on python code changes. So naming is not perfect yet.

.github/workflows/test-python.yml
jooola: Do we want to rely on the distribution python version or the one we ship ?

.github/workflows/test.yml
paddatrapper: The exclude paths can be used for a specific job as far as I know

jooola:
Duplicate of #1315 
Closing.

PR #1250, opened 2021-06-09
Update project structure

jooola:
This simplify the project structure by moving all our service in the root folder.

This is part of a step towards renaming the project to libretime, by removing the Airtime name in place where it is not needed.

Here are some bold moves we need to discuss before going further:
- Rename pypo to playout, clearer name, and `Python Play Out` could be `playout`
- Rename `api_clients` to `api_client`
- Move the `airtime_mvc` related files in the `airtime_mvc` directory, such as the `composer.*` files.

The install script has been changed accordingly, but the python modules inside each didn't change as this should be done in a separate PR.


zklosko:
> Move the airtime_mvc related files in the airtime_mvc directory, such as the composer.* files and vendor folder

If you move the Composer items, you may need to tell Composer where to look for them during the install process. Such files are kept in the repo root by default.

jooola:
> If you move the Composer items, you may need to tell Composer where to look for them during the install process. Such files are kept in the repo root by default.

I don't think I understand what you mean ?

zklosko:
Think of node_modules or package.lock. They're files needed by the package manager/dependency manager for it to work properly.

jooola:
@zklosko Ok, yes, this is already included in the changes, https://github.com/LibreTime/libretime/pull/1250/commits/4fa2f994b63a538458538017c3b03441aa56bc96

jooola:
This PR has been obsoleted by the smaller PR now merged in the main branch. So I am closing this.

PR #1255, opened 2021-06-22
Require python-dateutil version 2.7 or higher

paddatrapper:
isoparse was implemented in version 2.7


samkitk (approved):
Tried on a fresh install - was encountering this issue "from dateutils cannot import isoparse" 

Was calling pypo service to fail.

So I tried this change and it worked.
This commit worked for me.

Issue #1267, opened 2021-07-11
2021-07-11 | Installation Error


z8rsvenbergner:
I use Ubuntu 18.04 like the installer said.

The Installation is not working.
See the Screenshot below:
![CleanShot 2021-07-11 at 16 50 19](https://user-images.githubusercontent.com/75265769/125199785-2c248880-e268-11eb-9001-6c0593f8d640.jpeg)

paddatrapper:
This log is incomplete, but from the looks of it, it is the same issue fixed in #1262

matias-tecnosoul:
same problem here with latest git today.
some more logs:

> Installed /usr/local/lib/python3.6/dist-packages/libretime_analyzer-0.1-py3.6.egg
> Processing dependencies for libretime-analyzer==0.1
> Searching for pycairo==1.19.1
> Reading https://pypi.org/simple/pycairo/
> Downloading https://files.pythonhosted.org/packages/e8/9d/c8be300fc6b1298559d37a071c3833b0b251e0fff334d2e4c408d5789162/pycairo-1.19.1.tar.gz#sha256=2c143183280feb67f5beb4e543fd49990c28e7df427301ede04fc550d3562e84
> Best match: pycairo 1.19.1
> Processing pycairo-1.19.1.tar.gz
> Writing /tmp/easy_install-7qrsi79d/pycairo-1.19.1/setup.cfg
> Running pycairo-1.19.1/setup.py -q bdist_egg --dist-dir /tmp/easy_install-7qrsi79d/pycairo-1.19.1/egg-dist-tmp-x37_xd1a
> no previously-included directories found matching 'docs/_build'
> warning: no files found matching 'README' under directory 'tests'
> warning: no files found matching 'README' under directory 'examples'
> zip_safe flag not set; analyzing archive contents...
> cairo.__pycache__.__init__.cpython-36: module references __file__
> cairo.__pycache__._cairo.cpython-36: module references __file__
> creating /usr/local/lib/python3.6/dist-packages/pycairo-1.19.1-py3.6-linux-x86_64.egg
> Extracting pycairo-1.19.1-py3.6-linux-x86_64.egg to /usr/local/lib/python3.6/dist-packages
> Adding pycairo 1.19.1 to easy-install.pth file
> 
> Installed /usr/local/lib/python3.6/dist-packages/pycairo-1.19.1-py3.6-linux-x86_64.egg
> Searching for PyGObject<=3.36.1,>=3.34.0
[...]

PR #1269, opened 2021-07-13
Restore old python package versions for analyzer deps

paddatrapper:
### Description

Undoes the dependabot version updates as they break under Ubuntu Bionic

### **Links**

Fixes: #1267


paddatrapper:
Found a better way of doing this

PR #1280, opened 2021-08-03
Better error handling in api_client and airtime_mvc

jooola:
Add request method and payload to api_client http error handling, as we have some obscure behavior in this area.

Let the requests library encode the JSON payload if present.

Handle bad payload in some airtime_mvc api calls: Added a basic JSON decode error safe guard for the following actions:
- pushStreamStatsAction
- updateStreamSettingTableAction

### DIFF
--- a/python_apps/api_clients/api_clients/utils.py
+++ b/python_apps/api_clients/api_clients/utils.py
@@ -95,7 +95,7 @@ class ApiRequest:
         final_url = self.url.params(**kwargs).url()
         self.logger.debug(final_url)
         try:
-            if _post_data:
+            if _post_data is not None:
                 res = requests.post(
                     final_url,
                     data=_post_data,
--- a/python_apps/api_clients/api_clients/utils.py
+++ b/python_apps/api_clients/api_clients/utils.py
@@ -121,8 +121,9 @@ class ApiRequest:
             raise
         except requests.exceptions.HTTPError:
             self.logger.error(
-                f"HTTP request to '{res.request.url}' failed"
-                f" with status '{res.status_code}':\n{res.text}"
+                f"{res.request.method} {res.request.url} request failed '{res.status_code}':"
+                f"\nPayload: {res.request.body}"
+                f"\nResponse: {res.text}"
             )
             raise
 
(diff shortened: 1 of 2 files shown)


jooola:
Don't know if this should be merged right now I might push some more changes while debugging the api_client.

Or should we make a separate PR for the fixes ?

paddatrapper:
I'm happy to merge this as is, but if you want to add to it feel free. Just try keep the fixes here related to the PR title - makes creating the changelog easier later on

python_apps/api_clients/api_clients/utils.py
```diff
@@ -91,14 +91,14 @@ def __init__(self, name, url, logger=None, api_key=None):
...
+    def __call__(self, data: dict = None, params: dict = None, **kwargs):
         final_url = self.url.params(**kwargs).url()
         self.logger.debug(final_url)
         try:
-            if _post_data:
+            if data is not None:
                 res = requests.post(
                     final_url,
-                    data=_post_data,
+                    json=data,
```
jooola, line 101: @paddatrapper Ok, half of this PR is rubbish, I clearly didn't understood how the zend framework works. Any correction or confirmation would be really appreciated (cc @Robbt @hairmare), but here is what I understand:
The airtime_mvc api actions don't expect a JSON body, but a JSON encoded string inside a "data" POST form parameter.

In regards to #1172 and similar issues, I think the api_client doesn't get the post `data` anyway.

paddatrapper: Yuck... That sounds like a mess from Zend. We can probably add a optional parameter to the api_client call to map the `data` into a POST form parameter (defaulting to False)

jooola:
Any missing or invalid json string in the `data` parameter will generate the following logs:
```
2021-08-05T15:06:46+02:00 ERR (3): [REDACTED]radio.org [ApiController.php:1453 - updateStreamSettingTableAction()] - An error occured while decoding the 'data' JSON blob: ''
2021-08-05T15:06:46+02:00 ERR (3): [REDACTED]radio.org [ApiController.php:1436 - pushStreamStatsAction()] - An error occured while decoding the 'data' JSON blob: ''
2021-08-05T15:06:51+02:00 ERR (3): [REDACTED]radio.org [ApiController.php:1453 - updateStreamSettingTableAction()] - An error occured while decoding the 'data' JSON blob: ''
```

Many parts of the ApiController/python_apps rely on the power of dynamic typing, so we are missing type checking and schema validation. I think this is a huge problem, at least for communication with the outside of any app, data structures should be typed and validated. Anyway, I don't think this comment is really helpful.

If we are ok with the changes, this can be merged, I don't think I'll do any further improvements.

PR #1282, opened 2021-08-05
Add postgres and rabbitmq password update scripts

jooola:
I use this often to make clean install on my staging environment, though it could be committed back to the main repo.

This is only for advanced user who might stumble on this file.

### DIFF
--- a/utils/update-postgres-password
+++ b/utils/update-postgres-password
@@ -1,0 +1,21 @@
+#!/usr/bin/env bash
+
+set -u
+
+error() {
+  echo >&2 "error: $*"
+  exit 1
+}
+
+# Make sure only root can run our script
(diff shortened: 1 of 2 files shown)


paddatrapper:
This changes the Postgres and RabbitMQ passwords, so the name should probably reflect that.the installer also sets the icecast password

jooola:
@paddatrapper Just exploded the script in 2 separate script.

jooola:
Woups, some typo in the middle

jooola:
Ok good to go !

paddatrapper:
Do we want to reference them from the How to change default passwords documentation? https://libretime.org/docs/default-passwords

jooola:
I'd rather not, this should more be something for a proper `libretime-ctl` cli utility. But feel free to add it.

Issue #1283, opened 2021-08-06
PHP HTTPHelper does not handle HTTPS without force_ssl


paddatrapper:
In `Application_Common_HTTPHelper::getStationUrl()`, the system is supposed to calculate if the schema is http or https based on the port used and the `forceSSL` config flag. It currently is not working correctly for Airtime Analyzer callbacks (see https://github.com/LibreTime/libretime/issues/1274#issuecomment-893711950). The workaround is to set `forceSSL=true`, but this should not be required for installations using port 443 in the config.

paddatrapper:
https://github.com/LibreTime/libretime/blob/master/airtime_mvc/application/common/HTTPHelper.php#L46 seems to be the issue

PR #1284, opened 2021-08-06
correctly determine protocol from port and configuration in PHP

paddatrapper:
The PHP logic now matches the Python. Fixes: #1283

@Robbt would appreciate your review here, PHP is not a strong point for me

### DIFF
--- a/airtime_mvc/application/configs/conf.php
+++ b/airtime_mvc/application/configs/conf.php
@@ -32,6 +32,7 @@ class Config {
         $CC_CONFIG['stationId'] = $values['general']['station_id'];
         $CC_CONFIG['phpDir'] = $values['general']['airtime_dir'];
         $CC_CONFIG['forceSSL'] = isset($values['general']['force_ssl']) ? $values['general']['force_ssl'] : FALSE;
+        $CC_CONFIG['protocol'] = isset($values['general']['protocol']) ? $values['general']['protocol'] : '';
         if (isset($values['general']['dev_env'])) {
             $CC_CONFIG['dev_env'] = $values['general']['dev_env'];
         } else {
--- a/airtime_mvc/application/common/HTTPHelper.php
+++ b/airtime_mvc/application/common/HTTPHelper.php
@@ -32,6 +32,7 @@ class Application_Common_HTTPHelper
         $baseDir = $CC_CONFIG['baseDir'];
         $basePort = $CC_CONFIG['basePort'];
         $forceSSL = $CC_CONFIG['forceSSL'];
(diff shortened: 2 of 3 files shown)


airtime_mvc/application/configs/conf.php
```diff
@@ -32,6 +32,7 @@ public static function loadConfig() {
         $CC_CONFIG['stationId'] = $values['general']['station_id'];
         $CC_CONFIG['phpDir'] = $values['general']['airtime_dir'];
         $CC_CONFIG['forceSSL'] = isset($values['general']['force_ssl']) ? $values['general']['force_ssl'] : FALSE;
```
jooola, line 34: I would reconsider this line, and how it is then used as value in some conditions.
What I understand is that, if anything is set, say `no` (or `yes`, doesn't matter), the `$forceSSL` var will always evaluate as true, and only to false if it isn't set.

This is not the behavior I saw in the python code.

jooola (commented):
Again, this is just patching a small problem, I am really not in favor for keeping all this logic in the runtime code.

URL/URI are super powerful to describe all the different variables/specificity we are trying to keep in the config. So I would rather see a `base_url` config entry, with a full URL and eventually explode it if we need some parts of it, than trying to build a proper URL in different part of the code, over and over.

paddatrapper:
> Again, this is just patching a small problem, I am really not in favor for keeping all this logic in the runtime code.

Sure, but I want to get 3.0 out and look at breaking config changes later (maybe with the deprecation of the PHP code).

> URL/URI are super powerful to describe all the different variables/specificity we are trying to keep in the config. So I would rather see a base_url config entry, with a full URL and eventually explode it if we need some parts of it, than trying to build a proper URL in different part of the code, over and over.

There are two places this happens - this function in PHP and in api_clients/utils.py. Everything else should call one of the two. But yeah, a full `base_url` is probably better as I don't see a case where we need an individual portion anywhere

airtime_mvc/application/configs/conf.php
paddatrapper: I don't know PHP well enough to fix this...

jooola: I'll try to push something

jooola: See https://github.com/paddatrapper/libretime/pull/14

paddatrapper: Merged into this PR

jooola:
@paddatrapper Hey, did you find a better solution or was this change unnecessary ?

paddatrapper:
That was a mistake, re-opened as #1288

Issue #1272, opened 2021-07-19
chaining smart blocks in the show - they play together at the start-end point


chicco64:
I installed libretime using "git clone" with debian10.
I create several consecutive "shows" in the calendar (eg 09.55 - 10.05 // 10.05-10.15 ) and, within the show, I insert a smart-block (longer than the show duration, as obviously it can happen...), the next show starts at the appointed time, but does not stop the previous one, therefore the songs contained in the outgoing show and the incoming show overlap.
In the example, "a rose in the wind" and "slave to love" don't stop when "complicated" starts playing.
Thanks for the reply.
<img width="672" alt="libretime-smartblocks" src="https://user-images.githubusercontent.com/59960414/126132874-e5a38ea3-961f-440b-bade-378c884b6053.PNG">

paddatrapper:
What is the content of `/var/log/airtime/pypo/pypo.log`?

chicco64:
[pypo.log](https://github.com/LibreTime/libretime/files/6839967/pypo.log)

paddatrapper:
And the liquidsoap logs from the same directory?

chicco64:
this is the log in /var/log/airtime/pypo-liquidsoap

[ls_script.log](https://github.com/LibreTime/libretime/files/6847591/ls_script.log)

chicco64:
I proceeded to do a new installation but the problem remains the same, can you give me feedback? thank you

undoablemixlr:
Hello,
I have the same problem, I would also like to have an idea to correct this !

michielgardner:
I have the same problem. It happens on any show that is overbooked, not only when using smart blocks. 

I found that cue_out in the API response of the schedule is not set correctly since API v2. 

You can reproduce this by creating 2 shows of 1 minute. If you put a track longer than 1 minute in these shows, the cue_out for this track in the API response should be set to 60 seconds (which it does on pre-API v2 installs), instead of the full length of the track which causes the overlap.

To install the latest pre-API v2 version which does not have this problem, you can run this after cloning the git repository (before installing):

```git checkout 79ea5dfc739c5a517c14452a3afa32186209ad6e```

Robbt:
Thanks for finding this. I suspected it might be an issue with the new API. Now the challenge of analyzing the code and discovering the bug.

paddatrapper:
[...]

PR #1287, opened 2021-08-10
Fix v2 schedule API endpoint

paddatrapper:
### Description

Fixes: #1272 

### Testing Notes

**What I did:**

Unit tests included

### DIFF
--- a/api/libretimeapi/models/schedule.py
+++ b/api/libretimeapi/models/schedule.py
@@ -22,6 +22,28 @@ class Schedule(models.Model):
     def get_owner(self):
         return self.instance.get_owner()
 
+    def get_cueout(self):
+        """
+        Returns a cueout that is based on the current show. Cueout of a specific
+        item can potentially overrun the show that it is scheduled in. In that
+        case, the cueout should be the end of the show. This prevents the next
+        show having overlapping items playing.
+        """
+        if self.instance.ends < self.ends:
+            return self.instance.ends - self.starts
+        return self.cue_out
+
+    def get_ends(self):
+        """
+        Returns an item end that is based on the current show. Ends of a
+        specific item can potentially overrun the show that it is scheduled in.
+        In that case, the end should be the end of the show. This prevents the
+        next show having overlapping items playing.
+        """
+        if self.instance.ends < self.ends:
+            return self.instance.ends
+        return self.ends
+
     class Meta:
         managed = False
         db_table = "cc_schedule"
--- a/api/libretimeapi/tests/test_views.py
+++ b/api/libretimeapi/tests/test_views.py
@@ -1,7 +1,9 @@
 import os
+from datetime import datetime, timedelta
 
 from django.conf import settings
 from django.contrib.auth.models import AnonymousUser
+from django.utils import dateparse
 from libretimeapi.views import FileViewSet
 from model_bakery import baker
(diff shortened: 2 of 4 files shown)


api/libretimeapi/tests/test_views.py
```diff
@@ -1,12 +1,24 @@
...
 
 
+def fromisoformat(datetime_string):
+    try:
+        datetime_obj = datetime.datetime.strptime(
+            datetime_string, "%Y-%m-%dT%H:%M:%S.%fZ"
+        )
+    except ValueError:
+        datetime_obj = datetime.datetime.strptime(datetime_string, "%Y-%m-%dT%H:%M:%SZ")
+    return datetime_obj.time()
```
jooola, line 19: What moving this in a utils module and add more typing:
```py
from datetime import datetime

def fromisoformat(value: str) -> datetime:
    """
    Parse a iso formatted datetime string.
    """
    try:
        result = datetime.strptime(datetime_string, "%Y-%m-%dT%H:%M:%S.%fZ")
    except ValueError:
        result = datetime.strptime(datetime_string, "%Y-%m-%dT%H:%M:%SZ")
    return result.time()
```

paddatrapper: This is actually not needed and shouldn't be there. Will remove

api/libretimeapi/tests/test_views.py
```diff
@@ -1,7 +1,9 @@
+import datetime
```
jooola, line 1: I might be annoying with that, but I recommend using:
```py
from datetime import datetime, timedelta
```

api/libretimeapi/models/schedule.py
```diff
@@ -22,6 +22,20 @@ class Schedule(models.Model):
     def get_owner(self):
         return self.instance.get_owner()
 
+    def get_cueout(self):
+        if self.instance.ends < self.ends:
+            # Handle cases where the schedule item would overrun the show it is
+            # scheduled in
```
jooola, line 28: What about using doc strings to explain the function globally ?

I am all for comment in the middle of the code, but this one could be rewritten as function doc string, and would allow code writers to access the doc string from their IDE.

Robbt:
This looks good and should solve a critical bug. I can merge it but I haven't had a chance to test it yet.

paddatrapper:
It seems to work, at least for the unit tests I've added. I'd like to fix the things highlighted by @jooola then should be ready to be merged

paddatrapper:
@Robbt this is ready to merge

jooola:
For path manipulation, I recommend to take a look at the [`pathlib` library](https://docs.python.org/3/library/pathlib.html). Using the `Path` object simplify the code and is really powerful.

This PR is good to go, I am just saying for future improvement in our code, we should consider the `pathlib` library over `os.path`, when we don't have to compute a lot of paths.

Example:
```py
import or
from pathlib import Path

resources = os.path.join(os.path.dirname(__file__), "resources")
# Would turn into
resources = Path(__file__).parent / "resources"
```
I always have this table with me https://docs.python.org/3/library/pathlib.html#correspondence-to-tools-in-the-os-module

paddatrapper:
Yeah, pathlib is really nice. I think we do use it in some places, but I could be confusing myself with other projects I develop

PR #1295, opened 2021-08-16
Remove install apt-get dangerous options

jooola:
Apt should fail fast (and report) instead of
forcing a broken installation.

The version check is also not required anymore as
most distribution now provide an apt version > 1.1.


paddatrapper:
This is a good idea. It should also be done in the vagrant scripts where apt is called too

jooola:
Done

PR #1296, opened 2021-08-17
Fix installer bad logic

jooola:
Before the formatting, the check was useless.
The fixes from the formatting/linting PR messed the logic.

### DIFF
--- a/install
+++ b/install
@@ -583,10 +583,12 @@ while :; do
   shift
 done
 
-if [[ -z $web_root || ! -d $web_root ]]; then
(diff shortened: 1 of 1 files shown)


jooola:
Just ran the installer (using ./install -fiap) on a fresh debian 10, got a clean installation.
Maybe I'll habe another fix, but for now this can be merged as a single fix

paddatrapper:
I'm also going to be doing a couple installations in preparation for the meeting tonight, so will be testing this too

Issue #1300, opened 2021-08-17
Track type list is empty on default install


paddatrapper:
**Describe the bug**
When a new LT instance is created, there are no default track types and the install/set up documentation does not instruct the user to create any track types. This is confusing as the only item in the list is 'None'. You have to go to Settings -> Track Types to create a list and Settings -> General to set the default

**To Reproduce**

1. Create vagrant machine using `vagrant up debian-buster`
2. Step through install and restart components
3. Log in and go to Upload Track
4. See Track Type: None

**Expected behavior**
Either we need to improve the documentation and tell users to add track types, or we need to have a set of defaults that help in most cases.

**LibreTime version:** git ff96d55

**Installation method:**
- OS: Debian
- OS Version: Buster
- Method:  Vagrant `./install`

jooola:
This is actually a database migration problem.

The following migration file clearly insert track types on migration, but seems broken.
https://github.com/LibreTime/libretime/blob/master/airtime_mvc/application/controllers/upgrade_sql/airtime_3.0.0-alpha.9.2/upgrade.sql

I don't know if inserting such data is a good default behavior.

paddatrapper:
Looks like it needs adding to https://github.com/LibreTime/libretime/blob/master/airtime_mvc/build/sql/defaultdata.sql too

PR #1304, opened 2021-08-18
Set default track types for new install

paddatrapper:
Fixes: #1300


paddatrapper:
Ok, this doesn't work and I don't know enough about propel to fix it

Issue #1308, opened 2021-08-20
Make V2 API client Object Orientated


paddatrapper:
The API v2 client is the future way all python applications in LT (and potentially 3rd party applications) will interact with the web server. We're building it out as things are migrated. However, the current set up is very flat - 1 piece of logic interacting with the API is a single method in the APIClient class. This should probably take a more object-orientated approach, with Pythonic models reflecting the objects being dealt with. I'm not quite sure how to approach this problem without a lot of code duplication - the DB models are already defined in the API server. Also we eventually will need to do the same thing in JavaScript for the Vue UI, so do we really need to update the model in 4 places (PHP, JS, API and API client) when something changes? There must be a better way of approaching that.

jooola:
I already played around to build a better abstract client we could use for v1 and v2. And I agree the current client is really a difficult piece of software.

Another we want to consider, is to generate the api_client using the openapi schema. At least the models and routes, we can patch everything together if needed. This allow use to 1. generate all call automatically and with typing (yaaayy!), 2. we ensure our openapi spec is up to date, 3. we can test using openapi specs and have a large test suite just with the openapi specs.

I don't know if mixing the database models (maybe some other models only for the api IO) from the backend with the api_client is a good idea, we might want to process  data before sending it. As we briefly discussed if a part of the code should go in the api or in playout.

In addition, on could generate client for other languages, which can be useful for external services (really useful in my opinion).

jooola:
Work in progress on libretime/api-client

Issue #1302, opened 2021-08-18
Stream does not resume playing after offline


rocknowradio:
**Describe the bug**
After scheduling a near 24-hour length show (playing without any issues), if this is followed by 1-5 hours of pause, then creating a new show does not start automatically and needs one or more of the liquidsoap, playout and/or celery restart. After this it works normally.

Might be related to issue [129](https://github.com/LibreTime/libretime/issues/129).

**To Reproduce**

1. Go to LT web interface and schedule a 24-hour show S1 (not sure if the length or something else impacts). Say 2021-08-18 03:00 to 2021-08-19 03:00. Fill it with existing playlists (mp3 format) to have all 24 hours with content.  
2. Show starts playing (or needs services tweaking if it suffers from the same problem described here...)
3. Let the show finish and then 1-5 hours offline.
4. Create a new show S2 the next day around 07:45 (so is being offline for 4:45 hours). Set the show attributes and fill it with playlists as in step 1. 
5. The show S2 is created. It has no content yet as the start hour arrives, and is filled with content in the first minutes after the S2 start hour arrives.
6. Timeline:
2021-08-18 03:00 show S1 created
2021-08-19 03:00 show S1 ends, stream enters offline (between 2021-08-19 03:00 - 2021-08-19 07:45)
2021-08-19 07:45 show S2 is created for the period 2021-08-19 07:45 - 2021-08-20 03:00
2021-08-19 07:45 show S2 has no content yet
2021-08-19 07:48 show S2 is filled with content (actually I am adding 4 playlists, these gets saved, then others etc. so S2 content is modified 2-3 times until)
7. There is no repeat on either shows S1 or S2.
8. In LT web interface: Streams OK; Status reports "Check if that the libretime-celery service is installed" etc.
9. The show S2 does not start. VLC client reports metadata "Offline". After restarting celery, liquidsoap and playout (not being sure which one did not worked) the show S2 started normally.
Unfortunately I did not checked the services status first and went straight into the restart. I will let the current show expire and do the same thing today, and get back with the services' status as well.

**Expected behavior**
The new show S2 should start normally. Or, rather said, one or more of libretime services should not 

**LibreTime version:**
cf75b14e2

**Installation method:**
- OS: Ubuntu 64 bit
- OS Version: 18.04.5 LTS (Bionic Beaver)
- ./install -fiap
- regular install using the documented method on LT website

**Screenshots**
N/A

**Client (please complete the following information if applicable):**
[...]

PR #1309, opened 2021-08-20
Use api_clients method to update file size of items on the server

paddatrapper:
This is not pretty and would really improve from an OO API Client (#1308). But it should fix some of the errors seen in #1302

### DIFF
--- a/python_apps/api_clients/api_clients/utils.py
+++ b/python_apps/api_clients/api_clients/utils.py
@@ -91,7 +91,7 @@ class ApiRequest:
             self.logger = logger
         self.auth = KeyAuth(api_key)
 
-    def __call__(self, _post_data=None, _put_data=None, params=None, **kwargs):
+    def __call__(self, *, _post_data=None, _put_data=None, params=None, **kwargs):
         final_url = self.url.params(**kwargs).url()
         self.logger.debug(final_url)
         try:
--- a/python_apps/api_clients/api_clients/utils.py
+++ b/python_apps/api_clients/api_clients/utils.py
@@ -91,7 +91,7 @@ class ApiRequest:
             self.logger = logger
         self.auth = KeyAuth(api_key)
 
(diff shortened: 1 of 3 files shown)


python_apps/api_clients/api_clients/utils.py
```diff
@@ -91,7 +91,7 @@ def __init__(self, name, url, logger=None, api_key=None):
             self.logger = logger
         self.auth = KeyAuth(api_key)
 
-    def __call__(self, _post_data=None, params=None, **kwargs):
+    def __call__(self, _post_data=None, _put_data=None, params=None, **kwargs):
```
jooola, line 94: Let's protect us from future breaking changes and enforce using a keyword based argument passing to this function.

```patch
-    def __call__(self, _post_data=None, _put_data=None, params=None, **kwargs):
+    def __call__(self, *, _post_data=None, _put_data=None, params=None, **kwargs):
```

paddatrapper: Need to check in the version1.py and version2.py files. Those and their tests should be the only things calling it

python_apps/pypo/pypo/pypofile.py
```diff
@@ -88,8 +88,8 @@ def copy_file(self, media_item):
                 os.chmod(dst, stat.S_IRUSR | stat.S_IWUSR | stat.S_IRGRP | stat.S_IROTH)
 
                 if media_item["filesize"] == 0:
-                    file_size = self.report_file_size_and_md5_to_airtime(
-                        dst, media_item["id"], host, username
+                    file_size = self.report_file_size_and_md5_to_libretime(
```
jooola, line 91: While you are renaming the function, could we replace `libretime` with `api`, it is explicit and I think we shouldn't use the project name anyway.

jooola (commented):
Haven't tested this, but looks good. Only a small comment on naming.

PR #1314, opened 2021-08-31
Generate analyzer fixtures

jooola:
Generate the testing fixtures for the analyzer on the fly.

I added different kind of audio files that will be useful for testing future enhancement in the analyzer app.
Some features were poorly tested, such as the replay gain and the cuein cueout calculation.

I need this to ease testing while replacing silan and rgain3.


paddatrapper:
LGTM

PR #1315, opened 2021-08-31
Rework CI pipeline

jooola:
This PR need #1246 to be merged first, as the new workflow will build a docker image using the package manager to install the dependencies.

This PR does the following:

- Explode tests in multiple jobs.
- Cache apt dependencies in a shared docker image (also usable for local development).
- Run tests on different distribution (debian/ubuntu, and maybe centos).

Since the test rely on a docker image that is build only on the master branch, this PR will always fail.

Supersedes #1248


jooola:
I am unsure, whether we should merge the dev docker image building inside the main test workflow. This would allow to depend on this image building for the next jobs. But it adds a lot of complexity as we don't want to build the image on every run and the image should be updated once in a while or when the packages list has changed.

To have an overview of the pipeline, see https://github.com/jooola/libretime/tree/feat/rework_ci_testing

jooola:
We will be building a Docker image for each system we support, possibly 4/6 systems, and that requires a lot of time. This PR was also meant to improve the total running time, by extracting the packages install in a docker image we had gained a lot.

I could try to optimize the docker building time with better caching. I guess this needs more research.

paddatrapper:
Then I think we can leave it as-is and then iterate on it after this is merged (or before, depending on how long that takes). I don't think it should be a blocker for merging

jooola:
It is working as it is, I can push improvements later, if you think we should merge this.

paddatrapper:
Are there any setting changes required after merging this?

jooola:
> Are there any setting changes required after merging this?

What settings are your referring too ?

paddatrapper:
> What settings are your referring too ?

Any repo settings. Seems not, so I'm happy to merge this now

PR #1317, opened 2021-09-01
 Fix CI docker test image

jooola:
- Lowercase registry organisation name
- Always use LibreTime's org dev image for tests
- Never build the dev image on forks


jooola:
We might go a bit fast here, how will forked project handle this ?

jooola:
@paddatrapper I might have a better solution. I'll provide a fix.

jooola:
@paddatrapper Should be ready to be merged

PR #1318, opened 2021-09-01
Drop SysV and Upstart support

### LINKED ISSUE #1247: Drop SysV init system support
On the community channels, I proposed to drop support for the old SysV init system.
Our oldest supported systems Ubuntu 16.04 and Centos 7 defaults to systemd, and all other recent systems as well.

Keeping Sysv would be to please people who switched back to SysV intentionally. For those who still want to use SysV, I suggest they dig into the project to retrieve the old SysV config files.

I would even ask myself if we should not also drop Upstart support ? Who is still using it ?

jooola:
Closes #1247

I didn't change much in the installer script as I don't know yet whether we should clean it up or rewrite it.

Mostly remove sysv files and cleanup the setup.py files.

This can be reviewed, but I feel I am missing something, hence the WIP state.


python_apps/pypo/setup.py
```diff
@@ -41,9 +13,18 @@
     url="http://github.com/sourcefabric/Airtime",
     author="sourcefabric",
```
paddatrapper, line 14: And the URL/author here?

paddatrapper (commented):
I don't see anywhere else these are used

jooola:
Removed sysv, upstart, updated all the setup.py files with recent project details.

Important change, I relicenced the airtime-celery service to AGPLv3, are we OK with that ? I don't know if this is legally ok ?

Otherwise, should be good to go, we will still be able to cleanup files later on.

jooola:
I'll open another PR to update the setup.py details.

PR #1319, opened 2021-09-01
Fix tests failing on buster

jooola:
The CI pipeline discovered that we have some inconsistency across distribution with some tests.

For now, we could use some workaround until we properly fix the code.

### DIFF
--- a/python_apps/airtime_analyzer/tests/cuepoint_analyzer_test.py
+++ b/python_apps/airtime_analyzer/tests/cuepoint_analyzer_test.py
@@ -1,3 +1,4 @@
+import distro
 import pytest
 from airtime_analyzer.cuepoint_analyzer import CuePointAnalyzer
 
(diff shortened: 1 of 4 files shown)


python_apps/airtime_analyzer/tests/cuepoint_analyzer_test.py
```diff
@@ -1,3 +1,4 @@
+import lsb_release
```
paddatrapper, line 1: Module not found error. Think uou need python3- packages and added to requirements-dev.txt

paddatrapper: Or use platform https://docs.python.org/3/library/platform.html#platform.uname

jooola: I'll use `distro`, I really didn't think this though before using it. `lsb_release` is a bad idea.

jooola:
Should be good to go !

Issue #1108, opened 2020-11-15
Proposal: Remove unworking features from Libretime


zklosko:
Hi everyone. This is never a fun thing to bring up regarding something you're passionate about, but I think it's time to "trim the fat", so to speak. Libretime currently has a handful of features that either don't work, aren't integrated well, or are better served by 3rd party software/SaaS solutions that could add complexity if we were to embed them into a Libretime install. I think they should be removed to help us close out the 3.0 Release Blocker, but everything is certainly open to discussion, which I was hoping for from this post.

Here's a list to get us started:

- Publishing to My Podcast or Soundcloud: currently doesn't work (#1061 #1043 #1042)
- Hosting a podcast from Libretime (My Podcast): doesn't work by extension of broken publishing command, but a number of better options exist, including Bluberry for Wordpress and Anchor.fm. Add a download button to the interface for easy file retrieval?
- Microsite: kinda disappointing for what it is, and not helpful for stations not exposing Libretime to the outside internet. I'd like to suggest we replace the microsite with a customizable login form and focus on widget development for HTML5 and Wordpress. (#1060 #1094 https://github.com/LibreTime/libretime/issues/249#issuecomment-620260712)
- Reset Password function: We don't have an integrated sendmail server or docs on how to set one up; would still require use of an SMTP mail provider. I suggest we remove said form and add to login form: "Forgot your password? Contact your system administrator." (#724)
- ~Stream statistics: Cool function, but requires Piwik (now Matomo) which needs a custom setup and a MySQL database. If we aren't going to integrate Matomo into the install, why not add a place for code from an Analytics SaaS to be pasted to track stats externally? (See https://github.com/LibreTime/libretime/issues/196#issuecomment-450668733)~
- Red logout button in menu: we already have a logout link in the upper right corner of the UI, no need for a second. Fixing in #1101.

Robbt:
I agree with most of this. At least removing the Podcast stuff.

Stream Statistics does work without additional software. You can just do additional stream analysis with Matomo by exporting your icecast logs.

zklosko:
@Robbt Ah, I stand corrected. I'll have to add that to the docs.

paddatrapper:
I think this is a good idea, though adding better docs on email and the required config flags may be trivial enough to do that instead of removing the reset password logic. Could be wrong though, I don't know PHP very well

zklosko:
I'm not too familiar with it either. Would it be hard to embed a Captcha password reset mechanism?

[...]

PR #1321, opened 2021-09-02
Update project details in setup tools.

jooola:
This requires #1318 because I am lazy resolving conflicts.

Updated all the setup.py files with recent project details.

Important change, I relicenced the airtime-celery service to AGPLv3, are we OK with that ? I don't know if this is legally ok ?

### DIFF
--- a/api/setup.py
+++ b/api/setup.py
@@ -1,20 +1,24 @@
 import os
-import shutil
 
 from setuptools import find_packages, setup
 
-script_path = os.path.dirname(os.path.realpath(__file__))
-print(script_path)
-os.chdir(script_path)
+# Change directory since setuptools uses relative paths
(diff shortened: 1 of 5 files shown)


paddatrapper:
The MIT Licence seems to have been a deliberate choice and is compatible with the AGPL, so we need to leave it as MIT. Unless we can get agreement from all the contributors who have code still present to change it (too much work IMO)

jooola:
Ok, removed the license change commit.

The celery service could use a rewrite IMO, so we will see at this moment.

paddatrapper:
If we remove the My Podcast stuff, we may be able to remove it completely - https://github.com/LibreTime/libretime/issues/1108#issuecomment-755283394

PR #1322, opened 2021-09-02
Housekeeping

jooola:
The PR cleanup a lot of old files.

I tried to keep as much information while merging everything related to Campcatser/Airtime in a single file.

Do we agree on the `LEGACY.md` filename ?


jooola:
If we agree on every changes here, this can be reviewed/merged.

.github/invite-contributors.yml
```diff
@@ -1,4 +0,0 @@
-# invite-contributors config
```
paddatrapper, line 1: This needs fixing rather than removing. It should automatically add those who contribute regularly to the contributor team, but would need @Robbt or @hairmare to fix

python_apps/airtime-celery/README.md
```diff
@@ -0,0 +1,63 @@
...
+
+Each instance of airtime-celery has its own worker, and multiple instances can be run in parallel.
+[Celery is thread-safe](http://celery.readthedocs.org/en/latest/userguide/application.html), so this parallelization won't cause conflicts.
+
+# Usage
+
+This program must be run with sudo:
+
+```sh
+sudo service airtime-celery {start | stop | restart | graceful | kill | dryrun | create-paths}
```
paddatrapper, line 20: Can be replaced with systemctl calls, I'm not sure how we handle the non-standard commands though

python_apps/airtime-celery/README.md
```diff
@@ -0,0 +1,63 @@
...
+
+```
+/var/log/airtime/airtime-celery[-DEV_ENV].log
+```
+
+# Troubleshooting
+
+If you run into issues getting Celery to accept tasks from Airtime:
+
+1. Make sure Celery is running ($ sudo service airtime-celery status).
```
paddatrapper, line 59: systemctl

paddatrapper (commented):
I'm happy with the LEGACY.md name

.github/invite-contributors.yml
jooola: I don't really see why we should automatically add contributors as members.

It should be done by hand, and you should have permissions to add new members.

jooola: I'll open a ticket and remove the commit for the PR, so we can discuss this properly.

python_apps/airtime-celery/README.md
jooola: Will do in another PR, this was mainly converting a .rst file to markdown.

PR #1327, opened 2021-09-06
Hide My Podcasts feature

### LINKED ISSUE #1320: it is not possible to publish in mypodcast
I noticed that it is not possible to publish any audio tracks in the "mypodcast" section.
The "Publish" button does not give any results.
I have little experience with libretime yet, can you tell me if there is a solution? Thank you
Capture1

jooola:
Closes  #1320 

This simply hide the my podcast feature, as it isn't working yet. Might be removed in the future.

### DIFF
--- a/airtime_mvc/application/configs/navigation.php
+++ b/airtime_mvc/application/configs/navigation.php
@@ -13,14 +13,16 @@
  * in the menu.phtml script when they are output.
  */
 $pages = array();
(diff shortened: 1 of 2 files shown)


jooola:
Am I missing any link that should be commented out ?

paddatrapper:
I think that's all of them

PR #1339, opened 2021-09-10
Start linting :tada: !

jooola:
- Enhance the way we manage development dependencies (slowly moving towards pyproject.toml file structure)
- Require python >=3.6
- Use a shared makefile for python apps
- Start linting every python app ! (With github annotations !)

### DIFF
--- a/.github/annotations/pylint.json
+++ b/.github/annotations/pylint.json
@@ -1,0 +1,32 @@
+{
+  "problemMatcher": [
+    {
+      "severity": "error",
+      "pattern": [
+        {
+          "regexp": "^([^:]+):(\\d+):(\\d+): (E\\d+): \\033\\[[\\d;]+m([^\\033]+).*$",
+          "file": 1,
+          "line": 2,
+          "column": 3,
+          "code": 4,
+          "message": 5
+        }
+      ],
+      "owner": "pylint-error"
+    },
+    {
+      "severity": "warning",
+      "pattern": [
+        {
+          "regexp": "^([^:]+):(\\d+):(\\d+): ([A-DF-Z]\\d+): \\033\\[[\\d;]+m([^\\033]+).*$",
+          "file": 1,
+          "line": 2,
+          "column": 3,
+          "code": 4,
+          "message": 5
+        }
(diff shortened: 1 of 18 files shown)


jooola:
Ready to be reviewed !

The test will only pass once we fix the linting errors, and some off them are dangerous ! But it won't fail the entire workflow.

paddatrapper:
The lint task fails, but the pipeline reports success, is this intended behaviour? Also there are lots of errors and warnings in the linter because it doesn't take Django styling and practises into account - for example missing members and too few method errors.

jooola:
> The lint task fails, but the pipeline reports success, is this intended behaviour? 

Yes, there is way too much linting errors to fix in one go, and I don't have the energy to fix them as the code might be rewritten for some parts.

I was hoping we could slowly improve the code by reporting lint errors in CI and add annotations in future PRs touching python code.

Once an app received the amount of care it needs, we can move it in the test job. And let it fail if an error occur.

The next stage is the make mypy fail, which will enforce proper typing and report types mismatch, but that requires also a lot of work.

> Also there are lots of errors and warnings in the linter because it doesn't take Django styling and practises into account - for example missing members and too few method errors.

I don't want to disable errors/warning globally, this should be done in a configuration file per project, per file, or even per line, so we don't hide useful warnings. In the case of too-few-methods, in agree that those can be disabled on a per project basis, as they are mainly small warnings if not recommendations. 

In addition, I am not sure about the idea of moving the dev dependencies in the extras_require, as it isn't intended for that use, and I am kind of doing some monkey patching.

One of the main reasons behind this was that I wanted to be able to install the api with the psycogs2 binary package for development and the compiled one (requiring lib-pg) for production. Maybe the dev requirements should stay in there own file, and use extra_requires to select the right version of psycogs2.

jooola:
Also, in order to keep concerns isolated from another, we should requires some dev dependencies in the shared make file (black => format, pylint => lint), so one don't have to copy past the requirements-dev.txt file everywhere.

jooola:
Woups, I intended to changes a few things., I'll push another PR

PR #1342, opened 2021-09-13
Follow up #1339

jooola:
Some extra changes I wanted to ship in the last PR.

And some install fixes.

I think we shouldn't install the scripts entry points in `/usr/bin`, as this is the path for system packages, using the default `/usr/local/bin` seems appropriate. @paddatrapper what do you think ?

### DIFF
--- a/api/Makefile
+++ b/api/Makefile
@@ -6,5 +6,6 @@ PIP_INSTALL := --editable .[dev]
 PYLINT_ARG := libretimeapi
 MYPY_ARG := libretimeapi
 
+format: .format
 lint: .pylint .mypy
 clean: .clean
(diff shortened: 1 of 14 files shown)


paddatrapper:
> I think we shouldn't install the scripts entry points in /usr/bin, as this is the path for system packages, using the default /usr/local/bin seems appropriate. @paddatrapper what do you think ?

I agree, though we do need to be aware that packages may ship things to /usr/bin, so supporting that use case needs to be possible. Also in many instances /usr/local/bin is not in PATH

jooola:
Moved discussion on install-path to #1343

jooola:
I think this one can be reviewed and merged if ok.

PR #1344, opened 2021-09-13
Rewrite new lt.api_client

jooola:
### Description

This is a bold change. It mainly rework the api_client v1, since the v2 api_client is not yet broadly used in the project.

Since we will have to keep the api_client v1 for quiet some time before it can be removed, I created a new api client package which is less complex, and easier to work with/test.

The will provide us a future proof api client, while we can enhance to apps using it.

I first started by trying to remove the hard dependency on a configuration file, when I thought it would be easier to rewrite something cleanly.

I still consider generating the ApiClient using the openapi specs, but until then, this still is valuable.

A list of changes:

- Allow api_key auth via request headers in api v1.
- Use query params instead of weird zend path params.
- Move python_apps/api_clients to api_client
- Start using `lt.*` namespace for our packages.
- Renamed `api_clients.version*.AirtimeApiClient` to `lt.api_client.v*.ApiClient`
- Split pure `ApiClient` and all additional logic into `ApiClientCompat` (exported `ApiClient` is actually `ApiClientCompat`). The additional logics should either go in the API or in the service using it.
- Use loguru for logging
- Use tenacity for retrying failed requests (Does urllib3 provide this mechanism already ?)

Yet todo:

- Create `AbstractApiClient` to share with v2.
- Rewrite `v2.ApiClient`
- Testing ! A lot of testing ! ( A demo/testing instance would be awesome for integration testing, who could host this ?)
- Slowly replace the old lib with the new one in the project using it.

I am publishing the draft here, so I any of you can review/comment this changes.


paddatrapper:
I haven't looked at the code in detail, but a couple of thoughts on the concept:

* `lt.*` namespace -> This is too generic, we should rather use `libretime.*`. If people think that is too much typing, they can import as `lt` anyway.
* Logging -> what is the advantage of loguru? All the other applications, including the API use the standard logging library. This breaks consistency for configuration between them. We already have log rotation with standard system utils (logrotate), formatting is specified per project globally for that project, as is log file location. f-strings remove the requirement for string formatting improvements.
* Retrying failed requests -> urllib3 has this support (https://urllib3.readthedocs.io/en/1.26.x/user-guide.html#retrying-requests).

A v2 only client based on the OpenAPI Spec should probably be its own project, available from pypi/distro packages and be what we migrate our clients to.

Demo instance -> there were some offers on discourse, but I have not heard of any progress there.

jooola:
> lt.* namespace -> This is too generic, we should rather use libretime.*. If people think that is too much typing, they can import as lt anyway.

That fine for me, no strong opinion, as long as we use a namespace.

> Logging -> what is the advantage of loguru? All the other applications, including the API use the standard logging library. This breaks consistency for configuration between them. We already have log rotation with standard system utils (logrotate), formatting is specified per project globally for that project, as is log file location. f-strings remove the requirement for string formatting improvements.

Sorry, I made a lot of decision in my work that haven't been discussed yet, I guess it is the time for me to bring them up. Most of the decision came from my work on a shared library, used to initiate the different apps we have (playout/liquidsoap/analyzer).

See some of my branches with some exploration work:
- https://github.com/jooola/libretime/tree/wip/rewrite_python_apps
- https://github.com/jooola/libretime/tree/feat/liquidsoap/shared

Regarding the logging framework, I was thinking all the apps will all rely on the same `setup_logger` function embed in a `AbstractApp`. Just like sharing the logic to handle the configuration file (we wanted to change how we handle the configuration).
In few words, loguru is easier to use, and way more powerful out of the box. We won't have to bother with the complexity of the old logging, for multi threading & asyncio logging, everything is ready to be used. Exceptions handling, sweet colors :tada:, and more.

> Retrying failed requests -> urllib3 has this support (https://urllib3.readthedocs.io/en/1.26.x/user-guide.html#retrying-requests).

Yup, I add few comment in the code to consider using urllib3 directly, I think this is probably the cleanest option.

> Demo instance -> there were some offers on discourse, but I have not heard of any progress there.

A demo instance is not really usable for testing, but we could install a testing install next to the demo instance. I have a few CPUs/GBs of RAM  I can dedicate to a testing/demo instance, but I won't guaranty high availability (best effort), and the duration of the service, as I might need the resources at some point.

jooola:
Closing as this PR is way outdated. I might open another PR If I think we can gain rebuilding the api-client

PR #1346, opened 2021-09-15
Auth to API v1 using Authorization: Api-Key header

jooola:
This allow calls to API v1 to authenticate using the Api-Key http header instead of query/path params.


paddatrapper:
Should this not follow the same convention as v2? So use the `Authorization: API-Key <token>` header

jooola:
Yes, I am not sure what the standard is for Api-Key. I didn't want to mess with the Authorization header standard, as I didn't see any reference to a Api-Key Authorization scheme. But you are right, I'll change this.

jooola:
Changed the logic, it is really strict, don't add an extra space or it won't let you in ! 

Ready to be reviewed.

PR #1347, opened 2021-09-15
Format using php-cs-fixer

jooola:
So, I finally choose to format the the airtime_mvc. I didn't enabled the risky formatting rules.

This can be closed if we think it not worth it.


jooola:
@hairmare Friendly pinging you !

Robbt:
Not sure the best way to fully review this other than testing it, it seems to be applying the formatting fix, removing the " and replacing them with ' and adding spaces but 559 files is a lot to inspect. I don't have a lot of experience with php-cs-fixer but as long as it doesn't accidentally introduce any logic changes it should be fine to improve the code structure etc.

jooola:
Php-cs-fixer does not introduce "risky" changes unless you ask him to. I did not run any risky changes on the codebase.

The only situation we might have is if php-cs-fixer has a bug and won't handle some possible exotic case.

Php-cs-fixer fails hard where it doesn't understand a PHP files, for example https://github.com/LibreTime/libretime/pull/1347/commits/13a21b7bd528e07088cd55d8423640bcd3dea23c was required in order to format the file.

In my opinion this PR is safe to be merged, but a review is always welcome.

jooola:
Ideally I would also reenable pre-commit to run on the airtime_mvc folder. But that should go in another PR.

jooola:
Actually to reduce the number of files, we can remove the propel generated files from the format config.
They are generated anyway and there is a lot of them ! I will do this now.

jooola:
I found some changes that I wasn't sure about and handled them in separate commits.

As I would like to get this merged soon, I am manually reviewing each file. Once I am done reviewing, I'll do a small test and merge this.

Issue #420, opened 2018-01-23
Config File Name


paddatrapper:
We should rename the configuration file to `libretime.ini` and expect it in `/etc/libretime/`. Backwards compatibility could be provided by symlinking `airtime.ini`in the install script

hairmare:
I'm +1 for this but would like to do it after having tagged 3.0.0-alpha.4 since it's a change that needs to get some git master testing before we can release it. I'll tag it soon as i believe we got the uploading problems from this release window (#352) somewhat pinned down to have info on them in the release notes to soften the blow issue wise.

IMO backwards compat isn't strictly needed as all the code bits that use the file already use env variable to get the actual files location so changing the file location is basically config at this point.

paddatrapper:
Ah Ok, yeah env variables does make it easier. Installer issues being where its installed and setting the env variable?

hairmare:
I was under the impression that its cleaner but turned out to be wrong :(

Some parts are okayish

https://github.com/LibreTime/libretime/blob/6e2cb2b2a80442f5a5aa9e181649393bdabd6ea7/installer/systemd/airtime-celery.service#L8-L10

PHP has the naming in one central location

https://github.com/LibreTime/libretime/blob/4557395a865ed8d7592604b73c47ed3f5353c9dc/airtime_mvc/public/index.php#L34-L36

And then there is also heaps of legacy thingies

https://github.com/LibreTime/libretime/blob/276a69aa1bef527260efc170d7caab7d04b4ca13/python_apps/pypo/liquidsoap/liquidsoap_prepare_terminate.py#L5-L7

Since we need to touch the code anyway, lets somewhat standardize this. Overhauling config reading has been on my wishlist for a while anyway.

I'm not sure how far we want to go, using yaml instead of ini would make lots of the php python compat problems we have today go away if implemented properly.

If we keep the ini file we could read the libretime.ini location first and issue a warning if it doesn't exit and then try reading airtime.ini.

For the python parts it might make sense to implement config loading in a common module for the individual apps to use. The php parts already have the config file reading in one location so it's fairly easy to integrate there.

paddatrapper:
I'm definitely with you with overhauling the config stuff. Yaml would be nice, but developer availability is an issue I guess... Unfortunately I'm reluctant to put my hand up for it because life will get crazy soon and I've still got a fair amount of packaging to do.

[...]

PR #1349, opened 2021-09-16
feat: create libretime_shared package

jooola:
This adds a shared package usable by our python apps, to handle common tasks such as loading the logger, load a configuration file.

It the beginning of a series of PR that will start using this library in order to enhance the different apps.

This library assumes the following:
- We will use `Click` to create a proper CLI for each of our apps, allowing user and devs to tweak low level configurations and run advanced commands from the CLI.
- We will use `Loguru` instead of `logging`, for simplicity and loguru's power out of the box.
- We will wrap each of our log running functions (playout/liquidsoap/...) in a `App` class that will inherit from `libretime.shared.app.AbstractApp`. This Abstract app will setup common functionalities, and will have a consistent behavior across our tools.
- We will use `Pydantic`  and `libretime.shared.config.load_config` to load and validate our configuration. Some schemas will be shared across apps, such as `libretime.shared.config.RabbitMQ` and the `libretime.shared.config.Database`, but any app can define its own schema reusing the previous parts in order to load its own configuration.
- We will use a toml formatted configuration file or/and the environment to store our configuration.

Related-To #420

### DIFF
--- a/shared/README.md
+++ b/shared/README.md
@@ -14,18 +14,18 @@ This library assumes that:
 
 First define a schema for your configuration in order to validate it. A schema is a class that inherit from `pydantic.BaseModel`. Some existing schemas can be reused such as `libretime_shared.config.RabbitMQ` or `libretime_shared.config.Database`.
 
-Load your configuration using `libretime_shared.config.load_config`.
+Load your configuration using a subclass of `libretime_shared.config.BaseConfig`.
 
 ```py
 from pydantic import BaseModel
 
-from libretime_shared.config import RabbitMQ, load_config
+from libretime_shared.config import RabbitMQ, BaseConfig
 
 class Analyzer(BaseModel):
     bpm_enabled: bool = False
     bpm_track_max_length: int
 
-class Config(BaseModel):
+class Config(BaseConfig):
     rabbitmq: RabbitMQ
     analyzer: Analyzer
 
--- a/shared/README.md
+++ b/shared/README.md
@@ -1,0 +1,77 @@
+# Shared
+
+The `libretime_shared` package contains reusable functions and classes for the Libretime project.
+
+## Usage
+
+This library assumes that:
+
+- You will use [`Click`](https://github.com/pallets/click) to build a CLI for your app.
+- You will use [`Loguru`](https://github.com/delgan/loguru) to log messages from your app.
+- You will use [`Pydantic`](https://github.com/samuelcolvin/pydantic/) to validate objects in your app.
+
+### Configuration
+
+First define a schema for your configuration in order to validate it. A schema is a class that inherit from `pydantic.BaseModel`. Some existing schemas can be reused such as `libretime_shared.config.RabbitMQ` or `libretime_shared.config.Database`.
+
+Load your configuration using `libretime_shared.config.load_config`.
+
+```py
+from pydantic import BaseModel
+
+from libretime_shared.config import RabbitMQ, load_config
+
+class Analyzer(BaseModel):
+    bpm_enabled: bool = False
+    bpm_track_max_length: int
+
+class Config(BaseModel):
+    rabbitmq: RabbitMQ
+    analyzer: Analyzer
+
+config = Config(filepath="/opt/libretime/analyzer/config.yml")
+```
+
+### App
+
+Create an app class that inherit from `libretime_shared.app.AbstractApp`.
+
+```py
+from libretime_shared.app import AbstractApp
+
+class LiquidsoapApp(AbstractApp):
+    name = "liquidsoap"
+
+    def __init__(self, some_arg, **kwargs):
+        super().__init__(**kwargs)
+        self.some_arg = some_arg
+
+    def run(self):
+        ...
+
+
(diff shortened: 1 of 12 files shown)


paddatrapper:
All three dependencies are already packaged in Debian. I'm not sure about the static typing though... Having validation is nice, but Python is dynamic by nature and that dynamic typing is something I like a lot. If the type is not apparent from the context, your code needs changing.

jooola:
> I'm not sure about the static typing though... Having validation is nice, but Python is dynamic by nature and that dynamic typing is something I like a lot. If the type is not apparent from the context, your code needs changing.

I am unsure how to respond to that, don't know what you are referring to exactly.

Still this is not static typing but type hinting. We don't remove the possibility to run dynamic code. 
It will allow us to check for inconstant types while linting, not at runtime.

As much types hints we have the easier it will be for contributors to work with the code, as the code is self documented, and linting will prevent dumb type mismatch errors. While can still can run dynamic code where it needs to be dynamic.

If you are referring to the mypy configuration, we don't have to be that strict in the entire project. But when creating a library with a specific API, you want to type hint/control what comes in and out.

Just like our REST API we need to validate the input or we will have to handle way too many cases that might lead to runtime errors.

jooola:
Should we consider `libretime.common` instead of `libretime.shared` ?

I don't have a strong opinion, some prefer `common`, I think following standards should be the way to go.

paddatrapper:
> I am unsure how to respond, don't know what you are referring to exactly.

Ah for some reason I thought it was a static typing system for Python, not just a type hinter. 

> Should we consider libretime.common instead of libretime.shared ?

I also don't have a strong preference, though common does seem to be most, well, common

jooola:
I would like to merge this pretty soon but I need some input first.

## Config file format and schema

Yaml is a more broadly used file format, so I am thinking that using TOML is not necessarily the best idea. I like to TOML format though, and it is closed to INI. I would be open for both, so I wonder what you folks prefer ?

Using Yaml let use use the broadly used Symfony Yaml component for PHP, in addition to the Symfony Validator component.

We should not store the Config schema object in the shared library, this should be handled in each individual package.
Some schema can be stored in the shared library, such as the Postgresql config or the RabbitMq config. Other package will be able to add the to there own schema if needed.

Mayeb having everything in the same place does bring some pros, but I only see a possible config generator or migration tool for a libretime console cli.

paddatrapper:
I like the idea of using yaml for configuration, just because it is so widely used. I prefer it to TOML because I do a lot of kubernetes work, so I work with yaml a lot, while I haven't done much with TOML or INI. I don't have a strong preference either way though

paddatrapper:
Looks good, though the tests are failing

jooola:
The test are failing because I am using a pydantic feature that isn't working as expected. I will try to fix this with some custom code, until a better solution is provided by pydantic.

shared/README.md
```diff
@@ -0,0 +1,77 @@
...
+
+class Analyzer(BaseModel):
+    bpm_enabled: bool = False
+    bpm_track_max_length: int
+
+class Config(BaseConfig):
+    rabbitmq: RabbitMQ
+    analyzer: Analyzer
+
+config = Config(filepath="/opt/libretime/analyzer/config.yml")
```
paddatrapper, line 32: How does this handle common configuration such as database? Does it need to be duplicated into all applications that access it?

jooola: Similar to the RabbitMQ field, you declare the config requirement in your app, and load it:
```py
from libretime_shared.config import BaseConfig, BaseModel, Database

class Analyzer(BaseModel):
    bpm_enabled: bool = False
    bpm_track_max_length: int

class Config(BaseConfig):
    database: Database
    analyzer: Analyzer

config = Config(filepath="/opt/libretime/analyzer/config.yml")
```

I assumed that only the MQ and Database Models are shared across the different apps, otherwise each app will have to define there own Configuration fields.

paddatrapper: But that example uses `/opt/libretime/analyzer/config.yml`, so my reading would be that each app has their own configuration file. Is that correct? Otherwise do they all use the same file and ignore the sections that are not defined for their particular schema?

jooola: Ah ok, yes, the example file path is not realistic. All the configuration will be stored in a single file and each app will have to load from the same file.

This makes me think that a helper function / default field to get the config from the default path can be a good idea, so we have a single place where the default configuration path is.

jooola: I haven't tested it though !

paddatrapper: Maybe also use `/etc/libretime/config.yml` in the example documentation so that it is clearer that it should be the same file throughout

jooola: Done

paddatrapper: A helper function would be very useful, as having the option to specify the config file as an environment variable is useful for the Docker side of things

paddatrapper: I'm happy to merge as is and then helper functions/extensions can be handled as improvements

jooola: Ok, I'll do this in another PR

PR #1351, opened 2021-09-16
Allow ci image user to run sudo without password

jooola:
This will help allow modifications in CI or locally.


paddatrapper:
Can you explain more about why this is needed?

jooola:
When we create the libretime-dev image, we create a normal user with a dedicated home.

When running the image in CI with enter the image as the same user, but this user does not have any permissions to install packages or run as root. With this we can `sudo apt install` anything or run a script as root. Which can be required for local development or in CI.

This sudoers file trick is ugly from a security perspective, but we are running inside a development image, that is not intended for regular usage.

PR #1352, opened 2021-09-17
feat: create a control cli

jooola:
This PR will replace the current bash installer, with a python based installer.

- No dependencies for the install script.
- Use templates to deploy configuration files.
- Removed old configuration files.
- Moved and improved  some configuration files (apache/php). 
- Rename remaining paths to libretime.
- Fine grained install configuration using env variables.
- Support debian based systems (~~centos support on the way~~) and Systemd
- Group all config file in a single directory and use symlinks to deploy them on the system.
- Install/uninstall/clean commands
- (Still WIP...)

To consider:
- Use systemd WorkingDir to dedicate a working directopry for each app.
- More install config variables.
- Document installation options.
- CI Testing.

Most of the old names and files are renamed to Libretime as this is a good occasion to do so.
But I am wondering how we want to handle the upgrade path between airtime v2=> libretime v3=> v*.
If we keep old airtime paths in the Libretime v3 release we will have to maintain a complex upgrade path from v3 to v*.
Since Airtime user have to start fresh and we can still make breaking changes, I think we should update all the paths before v3.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -120,6 +120,7 @@ jobs:
           - analyzer
           - api
           - api_client
+          - console
           - playout
           - shared
           - worker
--- a/console/.pylintrc
+++ b/console/.pylintrc
@@ -1,0 +1,4 @@
+[MESSAGES CONTROL]
+disable=missing-module-docstring,
+        missing-class-docstring,
+        missing-function-docstring,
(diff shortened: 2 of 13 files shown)


jooola:
Gonna split this into 2 PR, one for a control facility, which might be useful to offload some of the installation tasks from the legacy `AirtimeInstall` tool.

And another one for the install script which might only be a rework of the current install script, helped by the control facility.

This would mean the installation process will move from the UI to the console.

PR #1292, opened 2021-08-15
feat: debian 11 support

jooola:
This is heavily inspired by #1168.

Closes #1271 

Needs more testing, specially on the following topics:
 - [x] playout has troubles ~~with the api version~~ starting. (See #1293)
- [x] ~~I don't use vagrant, so someone should test the vagrant stuff.~~ Vagrant is working

I think #1168 could reuse some of this PR to install the proper php packages.

PR #1168, opened 2021-01-24
Ubuntu 20.04 Support

paddatrapper:
Replaces #1120. Includes better error reporting and correct dependencies. Fixes #1109.

To do:

- [x] Add https://launchpad.net/%7Eondrej/+archive/ubuntu/php/+index?batch=75&memo=150&start=150 PPA for PHP 7.3
- [x] Test

PR #1354, opened 2021-09-17
Support php7.4

jooola:
This is an attempt to support php 7.4. Mostly updating dependencies.

Probably the only problem is propel. Should we fork the project, commit + patch the library in here (but really ugly) ?

Related to #1292 #1168


jooola:
Propel forks to consider:
- https://github.com/yarhon/propel1/ Clean simply php7.4 and php 8 support
- https://github.com/DavidGoodwin/Propel This fork might have added some extra changes I am unsure.


Probably other forks exists. @xabispacebiker You also forked, what do you think if we publish the package so we can use it here ?

I could use some help for the PHP stuff, as I am really not a php expert.

xabispacebiker:
> Propel forks to consider:
> 
> * https://github.com/yarhon/propel1/ Simple php7.4 and php 8 support
> * https://github.com/DavidGoodwin/Propel This fork might have added some extra changes I am unsure.
> * https://github.com/MyOnlineStore/Propel
> 
> Probably other forks exists. @xabispacebiker You also forked, what do you think if we publish the package so we can use it here ?
> 
> I could use some help for the PHP stuff, as I am really not a php expert.

Sure! what do you need ? I made it work by updating all the php requirements, at most I had to correct some mistakes by hand but it wasn't that much

jooola:
> Sure! what do you need ? I made it work by updating all the php requirements, at most I had to correct some mistakes by hand but it wasn't that much

What would be awesome it to be pointed to the proper propel fork. Also a change review could catch some bad things !

airtime_mvc/composer.json
```diff
@@ -16,7 +16,7 @@
     "massivescale/celery-php": "^2.1",
     "php-amqplib/php-amqplib": "^2.12",
     "propel/propel1": "1.7.0-stable",
-    "raven/raven": "0.12.0",
+    "raven/raven": "^0.12",
```
jooola, line 19: @paddatrapper @hairmare This is used to send logs to sentry, I don't know if it still used.
Based on the lack of documentation around it and that it seem to be a Airtime Saas tool, I assume it is not used, so I would suggest we remove it ?

paddatrapper: Not used in practice. May require some code changes to remove. Not sure if it is actually called anywhere

jooola:
Tests are passing, we only need some more testing and this should be good to go !

airtime_mvc/application/configs/airtime-conf.php
```diff
@@ -1,6 +1,6 @@
 <?php
-// This file generated by Propel 1.7.0 convert-conf target
-// from XML runtime conf file /vagrant/airtime_mvc/build/runtime-conf.xml
+// This file generated by Propel 1.7.3-dev convert-conf target
+// from XML runtime conf file /home/jo/git/github.com/jooola/libretime/airtime_mvc/build/runtime-conf.xml
```
paddatrapper, line 3: Maybe manually change this comment back to `/vagrant/...`

jooola:
Before merging I would like to clean the history, as we don't squash our PRs yet.

jooola:
@paddatrapper I would like to leave the question open for answers, but I think we should be proactive about this one and merge this.

We should not wait for the explanations from the few persons I pinged, and we will fix bugs if they occur. And I think it is safe to merge, as I haven't experienced any huge issue on the staging environment I am running with this PR and the debian bullseye PR.

jooola:
The work in this branch is missing some serious compatibility fixes, I'll open a new PR to really support php 7.4.

PR #1355, opened 2021-09-18
Update labels

jooola:
Rename some of the labels for more explicit labels.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,6 +6,9 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
+    labels:
+      - ci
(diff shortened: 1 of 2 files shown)


jooola:
I take the liberty of merging this I can handle the labels update in the issue triage at the same time. Otherwise we will have stale bot spamming a huge amount of mails.

paddatrapper:
While fine here, please do not merge your own PRs without a significant delay where another maintainer has not interacted with the PR, as per the C4 guidelines :)

jooola:
@paddatrapper Yes, agreed, this "over-zealousness" only came because this PR is "dumb".

PR #1358, opened 2021-09-21
Bump PHP dependency versions

paddatrapper:
Bumps our PHP dependencies to new versions of upstream. Required to fix Zend failing on Ubuntu Focal (https://github.com/LibreTime/libretime/pull/1168#issuecomment-917685593). Also pinned the dependencies on git master branches to versions where recent releases have been made.


jooola:
Possibly duplicate of #1354

paddatrapper:
Some, but I'm not addressing PHP 7.4, or thinning out unused dependencies. Just trying to get new upstream versions in. I do not intend on touching the propel dependency here

airtime_mvc/composer.json
```diff
@@ -15,36 +15,36 @@
     "propel/propel1": "1.7.0-stable",
     "aws/aws-sdk-php": "2.7.9",
     "raven/raven": "0.12.0",
-    "massivescale/celery-php": "2.0.*@dev",
-    "simplepie/simplepie": "dev-master",
-    "zf1s/zend-application": "^1.12",
+    "massivescale/celery-php": "^2.1",
+    "simplepie/simplepie": "1.5.6",
+    "zf1s/zend-application": "^1.13",
     "composer/semver": "^1.4",
```
jooola, line 21: I was able to bump semver to the latest version, as none of the breaking changes affect us.

packages.ini
```diff
@@ -23,6 +23,7 @@ php7.2 = bionic
 php7.2-curl = bionic
 php7.2-gd = bionic
 php7.2-pgsql = bionic
+php7.2-xml = bionic
```
jooola, line 26: Where does these come from ?

paddatrapper: simplepie requires it, but has only recently added a check that fails if it doesn't exist - https://github.com/simplepie/simplepie/pull/590

airtime_mvc/composer.json
paddatrapper: Bumped. If there are any others, we can probably include them in the PHP7.4 PR

Issue #1360, opened 2021-09-21
API schedule return broken


paddatrapper:
**Describe the bug**
The schedule response from the API v2 is currently broken after #1356. The is_valid check logic is incorrect and results in errors as is_valid is a property and not a DB field.

jooola:
Yup, sorry I didn't tested your PR, I'll try to do so, but this entire vagrant stuff is way too slow/complexe for quick testing.

paddatrapper:
If you have a box up most of the time, it makes it fairly quick to test. The issue I find (at least on libvirt) is that I can't destroy and recreate a box without restarting my machine as the port-forwards do not come up correctly. Haven't had time to find a fix

jooola:
While this issue has been fixed, the is_valid parameter is missing from the openapi specs.

PR #1361, opened 2021-09-21
Correctly filter schedule items based on is_valid

paddatrapper:
Fixes: #1360 and fixes the API integration tests to pass

### DIFF
--- a/api/libretimeapi/views.py
+++ b/api/libretimeapi/views.py
@@ -149,7 +149,7 @@ class ScheduleViewSet(viewsets.ModelViewSet):
         filter_valid = self.request.query_params.get("is_valid")
         if filter_valid is None:
             return self.queryset.all()
-        filter_valid = filter_valid.strip().lower() in ["true", "yes", "1"]
+        filter_valid = filter_valid.strip().lower() in ("true", "yes", "1")
         if filter_valid:
             return self.queryset.filter(starts__lt=F("instance__ends"))
         else:
--- a/api/libretimeapi/views.py
+++ b/api/libretimeapi/views.py
@@ -1,6 +1,7 @@
 import os
 
 from django.conf import settings
+from django.db.models import F
 from django.http import FileResponse
 from django.shortcuts import get_object_or_404
 from rest_framework import status, viewsets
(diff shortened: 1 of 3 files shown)


api/libretimeapi/views.py
```diff
@@ -141,9 +142,19 @@ class PreferenceViewSet(viewsets.ModelViewSet):
 class ScheduleViewSet(viewsets.ModelViewSet):
     queryset = Schedule.objects.all()
     serializer_class = ScheduleSerializer
-    filter_fields = ("starts", "ends", "playout_status", "broadcasted", "is_valid")
+    filter_fields = ("starts", "ends", "playout_status", "broadcasted")
     model_permission_name = "schedule"
 
+    def get_queryset(self):
```
jooola, line 148: Now that the get_queryset is defined I though you had to remove the queryset field ? And also specify the base name in the urlpatterns list ? 

I am mainly asking for education.

api/libretimeapi/views.py
```diff
@@ -141,9 +142,19 @@ class PreferenceViewSet(viewsets.ModelViewSet):
...
+        filter_valid = self.request.query_params.get("is_valid")
+        if filter_valid is None:
+            return self.queryset.all()
+        filter_valid = filter_valid.strip().lower() in ["true", "yes", "1"]
```
jooola, line 152: Small improvement, using a tuple will improve performances, not really a problem though, but we want to reach perfection here !

jooola (commented):
Haven't tested it yet, but seem ok too me.

api/libretimeapi/views.py
paddatrapper: If you leave the queryset field, then you don't need to specify the base name. As I am extending the queryset instead of replacing it in get_queryset, I figured I should leave it and reference it here instead

paddatrapper: If you leave the queryset field, then you don't need to specify the base name. As I am extending the queryset instead of replacing it in get_queryset, I figured I should leave it and reference it here instead

api/libretimeapi/views.py
paddatrapper: Good point, will do that

jooola:
Tested this, it is working, but the open api spec does not know about the is_valid parameter.

paddatrapper:
Ok, will need to dig into how to declare that then

jooola:
@paddatrapper We are in the middle of a testing week with our DAB broadcaster, and I would be happy to have this merged into master so I can fix he few bugs we are experiencing. Could we merge this and keep the issue open for the openapi spec ?

paddatrapper:
Yeah, I'm happy with that

PR #1362, opened 2021-09-21
fix php include for zend file transfer

paddatrapper:
Was failing to find the class


jooola:
Was this introduced by the dependencies upgrade ?

paddatrapper:
I suspect so and not mentioned in the zend changelog...

PR #1363, opened 2021-09-21
Update templates to use new labels for issue filing

paddatrapper:


paddatrapper:
That's pretty cool and would definitely help ensure we get the logs, etc that we need

Issue #1350, opened 2021-09-16
Pip does not support  --install-option="--install-scripts=/usr/bin"


jooola:
#1342 introduced a bug in the installer.

When installing the project in a vagrant box I get the following while install the python apps:
```
ERROR: Location-changing options found in --install-option: ['--install-scripts'] from command line. This is unsupported, use pip-level options like --user, --prefix, --root, and --target instead.
```

So we can either remove the option or revert to using setup.py instead of pip. I think using pip is a better idea, setup tools does not always resolve dependencies properly, and upgrading/uninstalling is cleaner using pip.

I propose to remove the install option and let pip install to /usr/local/bin.

Since the current state of the installer is broken on master, we should either resolve this fast or revert the commit from #1342 while we find a good solution for this.

Related to #1343

paddatrapper:
For the moment, I think we should revert back to setup.py for the affected packages. Long term we should move, but that needs a migration path for existing installs

paddatrapper:
Though if you have a chance to add that as part of the fix moving to /usr/local/bin, I'm happy with that too

jooola:
I'll push a fix after some more testing.

PR #1371, opened 2021-10-01
Install python apps using pip

### LINKED ISSUE #1343: Install entrypoints to default /usr/local/bin
This issue is from this discussion https://github.com/LibreTime/libretime/pull/1342#issuecomment-918181773

> >I think we shouldn't install the scripts entry points in /usr/bin, as this is the path for system packages, using the default /usr/local/bin seems appropriate. @paddatrapper what do you think ?

> I agree, though we do need to be aware that packages may ship things to /usr/bin, so supporting that use case needs to be possible. Also in many instances /usr/local/bin is not in PATH

I don't have any idea of which packages installs to /usr/bin, maybe celery ? I will check

Regarding the presence of /usr/local/bin in PATH, I quickly check in docker images :

```sh
~/git/github.com/jooola/libretime $ docker run -it --rm debian:buster bash
root@a80f2df1f18d:/# echo $PATH
/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
root@a80f2df1f18d:/# exit

~/git/github.com/jooola/libretime $ docker run -it --rm ubuntu:focal bash
root@2a1573e3fabd:/# echo $PATH
/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
root@2a1573e3fabd:/# exit

~/git/github.com/jooola/libretime $ docker run -it --rm ubuntu:bionic bash
root@8dd8c213a44d:/# echo $PATH
/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
root@8dd8c213a44d:/# exit

~/git/github.com/jooola/libretime $ docker run -it --rm centos:7 bash
[root@b63179901d4f /]# echo $PATH
/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
[root@b63179901d4f /]# exit

~/git/github.com/jooola/libretime $ docker run -it --rm centos:8 bash
[root@445c5175db27 /]# echo $PATH
/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
[root@445c5175db27 /]# exit
```

jooola:
### Description

- Install the python apps with pip install instead of bare setuptools.
- Change script install prefix to default `/usr/local/bin`.

This will simplify the way we install/uninstall the python apps.

Closes #1343 
Related-To #1350

I tested this on Buster (vagrant) and didn't face any problems.

BREAKING: Previous entrypoints scripts should be removed from `/usr/bin`
```sh
sudo rm -f \
  sqlformat \
  django-admin \
  django-admin.py \
  markdown_py \
  libretime-api \
  mutagen-inspect \
  mutagen-pony \
  mid3cp \
  mid3iconv \
  mid3v2 \
  moggsplit \
  collectiongain \
  replaygain \
  libretime-analyzer \
  airtime-playout \
  pyponotify \
  airtime-liquidsoap
```

### DIFF
--- a/api/setup.py
+++ b/api/setup.py
@@ -28,10 +28,11 @@ setup(
         "django-url-filter",
         "markdown",
         "model_bakery",
-        "psycopg2",  # TODO: Move to prod requires
     ],
     extras_require={
-        "prod": [],
+        "prod": [
+            "psycopg2",
+        ],
         "dev": [
             "psycopg2-binary",
         ],
--- a/install
+++ b/install
@@ -1028,7 +1028,7 @@ loudCmd "$pip_cmd install ${AIRTIMEROOT}/python_apps/api_clients"
 verbose "...Done"
 
 verbose "\n * Installing pypo and liquidsoap..."
-loudCmd "$python_bin ${AIRTIMEROOT}/python_apps/pypo/setup.py install --install-scripts=/usr/bin"
+loudCmd "$pip_cmd install ${AIRTIMEROOT}/python_apps/pypo"
 loudCmd "mkdir -p /var/log/airtime/{pypo,pypo-liquidsoap} /var/tmp/airtime/pypo/{cache,files,tmp} /var/tmp/airtime/show-recorder/"
 loudCmd "chown -R ${web_user}:${web_user} /var/log/airtime/{pypo,pypo-liquidsoap} /var/tmp/airtime/pypo/{cache,files,tmp} /var/tmp/airtime/show-recorder/"
 systemInitInstall libretime-liquidsoap "$web_user"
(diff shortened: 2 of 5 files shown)


paddatrapper:
Do we need to remove the old binaries from /usr/bin/?

jooola:
> Do we need to remove the old binaries from /usr/bin/?

Yes, but I am unsure whether this should be written in the changelog as part of the upgrade process or in the installer.
I would go for a proper changelog entry, with possibly a cleaning script for the braves.

PR #1373, opened 2021-10-03
Rename python packages/entrypoints

jooola:
Rename the python packages and entry points.

BREAKING CHANGES:
```sh
# Remove old packages using
sudo pip uninstall \
  airtime-playout \
  airtime-celery \
  api_clients
```


jooola:
@paddatrapper Can be reviewed / merged. I just tested this on buster, no problems found.

PR #1376, opened 2021-10-05
Versioned php packages and add missing packages

### LINKED ISSUE #1256: PHP version mess on debian buster with packages.sury.org/php
**Describe the bug**

The installer install php 8 alongside php 7.3 and 7.4 on debian buster with the sury php repository.

I think it's a bad idea to use virtual packages to install php extensions. We could target the versioned one directly.

**To Reproduce**

1. Spin up a debian buster
2. Setup the packages.sury.org/php repository
3. Launch the libretime installation

**Expected behavior**
Handle php version properly.

**LibreTime version:**
master

**Installation method:**

- OS: Debian buster
- Method: `./install -fiap`

jooola:
### Description

Closes #1256 if we keep being careful about using versioned packages.

### DIFF
--- a/airtime_mvc/packages.ini
+++ b/airtime_mvc/packages.ini
@@ -1,22 +1,37 @@
 # This file contains a list of package dependencies.
 [common]
 apache2 = buster, bionic
-php-apcu = buster, bionic
+# The php-pear package depends on php-(cli|common|xml), be sure to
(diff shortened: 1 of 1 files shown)


jooola:
I checked on each distro if the apt install accepted the generated package list, and it is working.

Issue #1378, opened 2021-10-06
Finalize Airtime > Libretime forking


jooola:
I haven't found any issue with regards to rebranding from Airtime to Libretime.

This is a global ticket which will attempt to gather all the steps required to rebrand the project to Libretime and remove most of the airtime occurrences in the code.

This will be considered as a breaking change and should be resolved before we hit v3 beta. Leaving some airtime migration steps for the future will probably only cause trouble and headache, we should do it once and for all.

While we rename the files and dirs:
- We should remove useless Libretime/Airtime names in the code (eg `AirtimeApiClient` > `ApiClient`)
- We might alter some of the paths and names for homogeneous and simpler paths (eg `pypo` > `playout`).
- We should consider using symlinks to deploy configurations across the system (eg `/etc/libretime/apache.conf => /etc/apache2/site-enabled/libretime.conf`).
- We will try to follow the modern https://refspecs.linuxfoundation.org/FHS_3.0/fhs/index.html as a non packaged software.

Some files should remain untouched:
- LEGACY.md

What need to be done:
- [x] Rename `airtime_mvc` to `legacy` (having airtime_mvc in most of the paths will probably make any search and replace difficult) #1379 
- [x] Rename the python packages (=> `libretime-*`) #1373
- [x] Rename the entrypoints/scripts/binaries (=> `libretime-*`) #1373
- [x] Rename the services  (=> `libretime-*.service`)
- [ ] Rename the config paths (=> `/etc/libretime/config.yml`) #420 
- [x] Rename the runtime paths #1380 
- [x] Rename the logs paths  (=> `/var/log/libretime/*`)
- [ ] Rename the shared files paths (=> `/usr/share/libretime/legacy`)

paddatrapper:
Eventually it would be nice to have them as environment variables because it makes things like contanerization easier. Whether it is worth it now is probably up for a cost/benefit analysis

jooola:
In the process of renaming a lot of paths, I wanted to propose less fixed paths.

Currently many paths a hardcoded in the code, and this isn't helping us at all. I see at least 4 different setup:
- dev/testing:  I expect the logs to go in the working directory.
- docker: I would simply log to stderr and leave other tools get the logs.
- packaged setup: In a packaged setup, I expect the files to be deployed the well known `/etc` `/var/log` `/var/tmp` paths.
- custom production setup: one might want to install libretime with custom paths.
- 
[...]

PR #1379, opened 2021-10-06
Rename airtime_mvc to legacy

jooola:
### Description

Related to #1378


paddatrapper:
Are they built as part of the install?

jooola:
I don't really know how po/mo files works.

Maybe some update locale generate them, but after a few searches it seems we need to run msgfmt or similar to generate the files.

PR #1386, opened 2021-10-12
Add missing local api_client dependency

jooola:
### Description

Some package require api_clients but isn't marked was dependency. It assume the package is installed on the system.

When working with other kind of installations, this isn't working at all.


paddatrapper:
Should this also be added to the analyzer? It also uses the API client AFAIK

PR #1391, opened 2021-10-13
Better format enforcing

jooola:
### Description

Remove the global pre-commit hooks for black and isort and use a per project setup.

Maybe running linting in parallel in CI is a bit expensive in CI time, how do we feel about that ?


jooola:
Ok, yeah I didn't knew CI time wasn't counted.

jooola:
Ready to be reviewed / merged.

PR #1400, opened 2021-10-15
Enhance release workflow

jooola:
### Description

Nothing crazy:
- some formatting and naming updates
- moved release script to the root Makefile
- don't install build dependencies with apt

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,7 +18,6 @@ shell-check:
 VERSION:
 	tools/version.sh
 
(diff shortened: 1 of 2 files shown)


jooola:
So I am unsure if you are comfortable with Makefiles, I didn't comment much of the logic going on there. Should I explain a bit some of the steps ?

paddatrapper:
I'm pretty familiar with them. Most of my usage was several years ago when I was in the C/C++ world, so I'm a little rusty, but understand what is going on

PR #1403, opened 2021-10-16
Type fix and checks for tools packages

jooola:


jooola:
Ready to be reviewed/merged.

PR #1409, opened 2021-10-17
Use setuptools entrypoints

jooola:
### Description

This replace the previous bin scripts by setuptools console entry points.

This PR is build on top of #1406 as it will probably take some time to rebase these changes and I am lazy.

Some context and explanation: https://stackoverflow.com/questions/18787036/difference-between-entry-points-console-scripts-and-scripts-in-setup-py

PR #1406, opened 2021-10-17
Rename and move python apps dirs

jooola:
### Description

This move the pythons_apps/* to the top folder.

Instead of `airtime-celery` => `celery`, I changed to `airtime-celery` => `worker` as we might have a name conflict with the celery module.
The `python_apps/icecast2` files seemed unused, but moved them to the installer folder.

:warning: **PLEASE DO NOT SQUASH THIS PR, REBASE IT** :warning:

PR #1410, opened 2021-10-17
Use python namespaces

### LINKED ISSUE #1212: Rename and move apps in the libretime namespace
I propose to rename the new api package to `libretime_api` instead of `libretimeapi`, for consistency with the other tools naming.

`api/libretimeapi/` => `api/libretime_api/`

In some far future, I would even think the python apps could be moved in the root folder.

jooola:
This move all our python packages into the libretime namespace.

Closes #1212 

Depends on #1406 and #1409

:warning: **PLEASE DO NOT SQUASH THIS PR, REBASE IT** :warning:


jooola:
Django is not really happy with this namespace thing. I found a few bugs in the django code that makes using namespace difficult.

jooola:
Django bug report https://code.djangoproject.com/ticket/33207

@paddatrapper Could you confirm that this bug is not a configuration problem ?

paddatrapper:
The doc update was useful. I have never tried running a namespaced Django project before with a custom user model

jooola:
I was unsure how to read one of the comments, if it was helpful and I didn't understand, or it was a personal opinion just saying don't use namespaces packages.

Anyway, I hoped to merge this faster, I'll probably remove the django namespacing changes and create another PR. I prefer not leaving this open for too long, because it's always a pain to rebase / merge on entire folder renames.

jooola:
Since I don't want to waist my time in making this work with Django, I'll simply leave this branch on hold for now. This can be reconsidered later.

I will propose another PR that do the proper renaming for each libretime packages though.

jooola:
Closing this, if we want to use python namespace we should start fresh from the main branch. Syncing such changes is a waist of time.

PR #1424, opened 2021-10-20
Make Vagrantfile more configurable

jooola:
This will be required in for bullseye support, because the nfs version need to be tweaking depending on the guest.

Also some cleanup and extra features.


jooola:
I messed something need to fix a small thing.

Issue #1428, opened 2021-11-03
No module named 'pypo.notify'


SpaceTrucker2196:
**Describe the bug**

When trying to play tracks I'm getting an error for pypo.nofiy 

2021/11/02 22:45:18 [alsa_out(default):3] Samplefreq=44100Hz, Bufsize=32768B, Frame=4B, Periods=74
Nov 02 22:45:18 zebFM libretime-liquidsoap[17964]: Traceback (most recent call last):
Nov 02 22:45:18 zebFM libretime-liquidsoap[17964]:   File "/usr/local/bin/libretime-playout-notify", line 6, in <module>
Nov 02 22:45:18 zebFM libretime-liquidsoap[17964]:     from pypo.notify.main import run
Nov 02 22:45:18 zebFM libretime-liquidsoap[17964]: ModuleNotFoundError: No module named 'pypo.notify'
Nov 02 22:45:18 zebFM libretime-liquidsoap[17964]: Traceback (most recent call last):
Nov 02 22:45:18 zebFM libretime-liquidsoap[17964]:   File "/usr/local/bin/libretime-playout-notify", line 6, in <module>
Nov 02 22:45:18 zebFM libretime-liquidsoap[17964]:     from pypo.notify.main import run
Nov 02 22:45:18 zebFM libretime-liquidsoap[17964]: ModuleNotFoundError: No module named 'pypo.notify'
**To Reproduce**

After a clean install and successfully importing a few tracks.

After creating a new show and adding a track. 

**Expected behavior**
expecting the On-Air indicator to stay lit and track being to play.

**LibreTime version:**
3.0.0-alpha.10

**Installation method:**
021-05-07-raspios-buster

installed with $ sudo bash install -fiap

paddatrapper:
I can replicate this issue using the current master branch. It works if libretime-playout is installed as an editable package `pip install -e .`, but not if installed uneditable.

SpaceTrucker2196:
thanks! Replicated issues make the world turn. Have a great day!

PR #1431, opened 2021-11-08
fix: declare pypo.notify module

paddatrapper:
Fixes: #1428


paddatrapper:
The shell-check warnings are fixed in #1430

Issue #1434, opened 2021-11-09
./install command not found


jooola:
**Describe the bug**

The installer is throwing the following error:
```
./install: line 113: python3 -m pip install --upgrade setuptools~=58.0: command not found
```

This error was introduced by #1430, with a quoting issue. Bash didn't found the `"python3 -m pip install --upgrade setuptools~=58.0"` command.

We could either revert to using eval, or properly fix the quoting

This happens on the main branch the 9/11/2021

paddatrapper:
For the moment we should probably revert to eval and then can look at fixing it

PR #1436, opened 2021-11-09
fix: revert removal of eval for shell commands

paddatrapper:
Fixes: #1434

Temporary revert until we can fix the quoting


jooola:
What about adding a shellcheck ignore comment ?

paddatrapper:
Added

Issue #1435, opened 2021-11-09
Unable to install playout - api_client not found


paddatrapper:
```
debian-bullseye: python3 -m pip install /vagrant/playout
debian-bullseye: Processing ./playout
debian-bullseye: Processing //tmp/api_client
debian-bullseye: ERROR: Could not install packages due to an EnvironmentError: [Errno 2] No such file or directory: '/tmp/api_client'
```

jooola:
This might have something to do with the new pip out-of-tree feature https://github.com/pypa/pip/issues/7555
Have you tried upgrading pip ?

paddatrapper:
This is with pip 20 in a bullseye vagrant box. Upgrading results in pycairo failing to install. I think we should probably drop the out of tree install and just explicitly install api_client for the moment

jooola:
pip is failing because we installed the system package and also require a newer version of pycairo from pip. I guess the pip/disutils/apt packages version mix is not able to handle this properly.

Another way to fix this while still going forward is to remove pycairo from the apt packages to install as we require a newer version anyway.

See https://github.com/jooola/libretime/commit/cfb6cba7dd60f2002b66e1c94e1ef48a4fc77960

paddatrapper:
Looking at the error message, `//tmp/api_client`, perhaps we're writing the file location logic incorrectly?

jooola:
> Looking at the error message, //tmp/api_client, perhaps we're writing the file location logic incorrectly?

I don't think so, upgrading pip makes it work perfectly. And I picked the syntax from a popular stack overflow question.

PR #1441, opened 2021-11-17
fix: assume api client is installed

paddatrapper:
Until pip out-of-tree dependencies work with our setup and we can update to newer versions of our dependencies, assume api client is installed without explicitly adding it as a dependency. The install script installs api_client anyway

Fixes: #1435


jooola:
This should break CI and dev environments. But somehow doesn't seem to, weird.

Issue #1180, opened 2021-02-22
Enable Semantic Commits


paddatrapper:
This is for after 3.0 beta release.

Semantic commits ([conventialcommits](https://www.conventionalcommits.org/en/v1.0.0/)) allows automatic creation of changelogs and better indication of breaking changes. It can be used in conjunction with `semantic-release` to automatically release LT on a schedule with versioning based on semantic versioning

jooola:
An alternative or complement to consider : https://github.com/atlassian/changesets

jooola:
Also https://github.com/git-chglog/git-chglog

jooola:
Another great tool https://github.com/twisted/towncrier

jooola:
Another one, different style https://github.com/release-drafter/release-drafter

jooola:
Another tool suggested by hairmaire https://github.com/go-semantic-release/semantic-release.

This has been discussed in the community meeting the 17/09/2021.

We need to decide whether conventional commits is a too high barrier for newcomers. They might face a wall if they are not used to it, so we need to consider with regards to contributors.

Also this can be mitigated if we squash (rebase if possible) our PR and let the maintainers edit the PR title to match the conventional commit format.

But we should set this up, no need to wait for v3.

jooola:
I am actually happy with the conventional commit message for this repository, we should now enforce it using CI and some dev tools. And I think we all agree on this.

The tooling around conventional commit messages to generate changelogs or releases can be done after some extra research and testing.

jooola:
Maybe https://github.com/commitizen-tools/commitizen can enforce a commit message format

paddatrapper:
We're using squash and merge, which takes the PR name as the commit message if there is more than 1 commit. So this should probably be enforced at the PR naming level

jooola:
I am looking for a way to generate a change log from the conventional commit message history.

Here are some concerns I have.

## Spell / prose errors

Commit messages are possibly filled with spelling errors, as we don't have any spell checker on the commit messages. This make the change log full of spelling errors, and regenerating the change log will always introduce the same spelling errors.

We could ignore the change log file for spelling errors, but I feel we should fix the source of the errors rather than allowing them to land in the repo.

## Breaking change notice format

We can extract the breaking changes messages to build some upgrade procedure, but depending on how well the procedure is written, we might need to add some extra context and better formatting. I've been writing such breaking change notice and I think I can start document and design a format for user to follow while writing these changes.

[...]

PR #1442, opened 2021-11-17
chore: add semantic-release prefix to dependabot commits

paddatrapper:
Not strictly necessary yet, but will make it fit in with #1180 when we get there

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,6 +6,8 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
+    commit-message:
+      prefix: "chore: "
     labels:
       - ci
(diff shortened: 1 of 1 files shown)


jooola:
Woups, merged this too quickly, I think chore is not the right prefix, or is it ?

paddatrapper:
It is - not user facing. If there is more complicated things required to add support for a particular version of a dependency, then we have to do it and not dependabot. In which case it is a feature/fix depending on the nature of the change

jooola:
Btw, dependabot should be smart enough to guess this, see an example here https://github.com/jooola/earhorn/pull/3

paddatrapper:
That's cool

PR #1467, opened 2022-01-02
chore(ci): add closed references notificier workflow

jooola:
This workflow will check for issue links in the code
that has been closed upstream and can be dealt with.

### DIFF
--- a/.github/workflows/closed-references.yml
+++ b/.github/workflows/closed-references.yml
@@ -1,0 +1,29 @@
+name: Closed Reference Notifier
+
+on:
+  schedule:
+    - cron: "0 0 * * *"
+  workflow_dispatch:
+    inputs:
+      issueLimit:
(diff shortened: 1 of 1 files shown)


paddatrapper:
Can you elaborate a bit more on what this does?

jooola:
It scans our codebase for issue links, and checks if the issue is closed upstream. If it is closed, the notifier will open a new issue that the link in the code should be removed or dealt with.

paddatrapper:
Looks good. Only comment is that this isn't a chore(ci), it should just be ci: as if semantic release was run, this should not cause a point release (which chore would). For more examples of type and scope see https://github.com/angular/angular/blob/master/CONTRIBUTING.md#-commit-message-format

jooola:
@paddatrapper Aah good point, not totally familiar with conventional commits yet !

Issue #949, opened 2020-01-15
Libretime running with persistency on Docker (based on Ubuntu 18.04)


kessibi:
Hello, I have recently been working on a way of getting Libretime to work within a docker container (a __single__ container for all applications of Libretime): https://github.com/kessibi/libretime-docker 

I am proud of saying that the installation works correctly and the image can be persistent if volumes are passed. The image still misses some improvement but I thought it was a nice enough proof of concept. Would such a container be what you'd have in the container repository (https://github.com/LibreTime/libretime-container-images) ?

It is possible to test the image with `docker run -p 80:80 -p 8000:8000 -p 8001:8001 --name=libretime odclive/libretime-docker:latest`. The whole install process is described in the README file.

I am fairly keen on getting some feedback from you on this.

PS: Sorry if the issue was not the best way to contact you, I thought it would be relevant enough.

Robbt:
This looks like it solves #624 - I will test it out when I get a chance but this is definitely something we would be interested in integrating into the project. I didn't get very far in terms of working on a single docker container instance of LibreTime.

paddatrapper:
I really like the concept! Some preliminary testing produced https://github.com/kessibi/libretime-docker/issues/6

kessibi:
I do not have an extended knowledge of all the perks of Libretime so I depend on your feedback and the issues people provide me to improve the image.

Thanks for your consideration.

paddatrapper:
As for the question about the container repository - I think this depends on the purpose of that repo. If it is for a production-ready solution, then @ned-kelly multi-container is a better fit. But, if we want that repo to be a simple easy to setup and evaluate LibreTime, then a single container would be a good fit

Aphris-Karu:
> As for the question about the container repository - I think this depends on the purpose of that repo. If it is for a production-ready solution, then @ned-kelly multi-container is a better fit. But, if we want that repo to be a simple easy to setup and evaluate LibreTime, then a single container would be a good fit

You can do both. It does not have to be an ether or situation.

Robbt:
Yeah at this point what we need people to test it and we can integrate it into our docs. It'd also be nice to have an official docker image supported by the project itself or to integrate one of the existing ones. This was supposed to happen with the multi-container one but we didn't coordinate the permissions for the transfer while @ned-kelly was still active and so it didn't happen. It could be revisited or we could create our own image based upon that.

One they are tested we can then just add it to the documentation.

kessibi:
[...]

Issue #624, opened 2018-11-30
Proposal: Build a single docker container based on Ubuntu 16.04 as our recommend install


Robbt:
The technical details can be discussed under #551 but this is more for weighing out the pros and cons and making a decision as to whether this theoretical solution would be the best to direct our effort at.

My thoughts in favor of it are the following:

- we could bundle fixes for any broken packages automatically such as #552 & #380
- it would dramatically simplify the install process for most users (see these [installer](https://github.com/LibreTime/libretime/issues?q=is%3Aissue+is%3Aopen+label%3Ainstaller) issues)
- we could avoid support requests caused by weird configuration issues related to unsupported distros #553
- we could make it easier to deploy fixes without breaking instances being ran in production #623 
- it could be treated as the goal for our beta release see #544 
- it would probably be easy with a proxy to support multiple libretime instances on a single server [see this](https://www.digitalocean.com/community/questions/how-to-bind-multiple-domains-ports-80-and-443-to-docker-contained-applications)
- it is the most logical and simplest way to deploy complex software like libretime that basically depends on a whole server to run

Potential reasons we might want to avoid it:

- docker might be intimidating to people unfamiliar with it (but so is the linux command line)
- we would need to do more work testing and write docs for how to deploy it in production
- it could make us dependent upon a 3rd party to deploy if we used docker hub (but we already depend upon lots of 3rd party package managers)
- we could accomplish the same ease of install with debian packages, flatpak or snaps w/o introducing another level of abstraction and complexity #390

Feedback and thoughts are welcome.

frecuencialibre:
i'm in! 

the only thing i'd add is the experience of another excellent project - Azuracast, which has [recommended](https://azuracast.com/install.html#using-docker-recommended) installation with docker since 2017.

ned-kelly:
@Robbt - Probably worth reading my comments here: https://github.com/LibreTime/libretime/issues/552#issuecomment-443392574

This is already in my personal backlog so happy for this one to be assigned to me.

hairmare:
> flatpak or snaps

Those are basically just another way to run the same kind of containers... I think flatpak will (or already does) support running OCI containers.

> I'd also be keen to experiment on using something like Alpine Linux over Ubuntu

[...]

Issue #551, opened 2018-10-12
proposal: docker install repo(s) under LibreTime github user


frecuencialibre:
hi all, 
at our radio station, after much time spent trying different server setups, we've found that installing LibreTime with docker to be the best choice for [various reasons](https://github.com/LibreTime/libretime/issues/515#issuecomment-424862231).

we're currently deploying @ned-kelly 's [multicontainer libretime](https://github.com/ned-kelly/docker-multicontainer-libretime) build. i've also been trying to contribute back simple fixes, bug reports, and documentation, and we've agreed that it would be great for this build to come under the main LibreTime umbrella. What do folks think? Should this multicontainer repo be created?

Other libretime instalations with docker also exist, and we could also consider creating additional versions down the line. See [this thread](https://github.com/LibreTime/libretime/issues/515).

Probably the main reason NOT to create the repo is the concern is that "officially supporting docker will lead to more support overhead since we'll be explaining docker to folks that really just want to get LibreTime up and running."  this is fair. i personally could help with support triage, and would hope that docker-specific issues would be posted either in the forum or in the docker repo(s) rather than the main libretime repo, but I've only been using docker for a few weeks! Regarding this concern, I think we should learn from [AzuraCast's approach and experience](https://www.azuracast.com/install.html#using-docker-recommended). 

thanks all!
-ryan

hairmare:
I'm on board with this but, like explained elsewhere, I'd like to ensure that we can support multiple different variants of the docker images so I can also use it as the final location for my personal LibreTime on Openshift Dockerfiles.

I'll set up the repo and invite @ned-kelly to it as a maintainer over the weekend.

frecuencialibre:
given the [restrictive nature of Dockerfile naming](https://stackoverflow.com/questions/26077543/how-to-name-dockerfiles), i wonder if it wouldn't be best to have multiple descriptively-named repositories.

hairmare:
The git repo containing the container images don't need to map to images in a 1-1 fashion and docker builds can be configured for a git repo containing multiple Dockerfiles.

frecuencialibre:
hi all, i'm re-opening this since we don't have consensus, and even a little confusion caused by me mentioning this [elsewhere](https://github.com/LibreTime/libretime/pull/584#issuecomment-436868417).

@hairmare the repo with the issues that @ned-kelly suggests we transfer is the [multicontainer docker repo](https://github.com/ned-kelly/docker-multicontainer-libretime).

[...]

Issue #1470, opened 2022-01-02
Check playability using ffmpeg instead of liquidsoap


jooola:
With the recent versions of liquidsoap, most format will be playable, as long as ffmpeg is able to play / transcode them.

With that reasoning I would suggest to remove the hard dependency on liquidoap for the analyzer, and simply do some sanity check on the files using ffmpeg or similar.

PR #1471, opened 2022-01-03
feat: add docker-compose stack

paddatrapper:
### Description

Adds Dockerfiles and documentation around them

- [x] Shared
- [x] API client
- [x] Analyzer
- [x] API
- [x] Legacy
- [x] Worker
- [x] Playout
- [x] Liquidsoap
- [x] Documentation
- [x] CI push to Docker image repos

Fixes: #949
Fixes: #551 
Fixes: #624 

**This is a new feature**: yes

_Do the changes in this PR implement a new feature?_

**I have updated the documentation to reflect these changes**: TODO. Needs deployment and dev notes

### Testing Notes

**What I did:**

Built and tested Docker images

Follow the Docker build documentation (TODO)


analyzer/Dockerfile
```diff
@@ -0,0 +1,31 @@
+ARG RELEASE=bullseye
```
jooola, line 1: I'd suggest using bullseye slim, and if possible alpine ? I know that currently with liquidsoap /gstreamer/... using alpine is somewhat difficult, hence #1470 and #1221. But I would suggest we keep this in mind.

analyzer/Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+FROM python:3-${RELEASE}
+ARG RELEASE
+
+RUN apt-get update -qq && apt-get upgrade -y -qq
+ADD https://raw.githubusercontent.com/LibreTime/libretime/master/tools/packages.py /tmp/packages.py
+COPY packages.ini /tmp/
+RUN apt-get install -y -qq $(python3 /tmp/packages.py --format=line ${RELEASE} /tmp/packages.ini)
```
jooola, line 8: What about making this a single RUN, using wget instead of ADD ? Does it break caching ?

analyzer/Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+COPY setup.py ./
+# It will fail because the source code is not installed yet.
+# Only required for the requirements file, which is written
+# before the failure. This is broken up to prevent needing
+# to reinstall all dependencies when the source code changes
+# and the Docker image is rebuilt
+RUN python setup.py egg_info || true
+# seup.py cannot filter out extra_requires, so remove the
+# [dev] section and leave only the install_requires section
+RUN sed -i '/\[dev\]/,$d' libretime_analyzer.egg-info/requires.txt
```
jooola, line 24: Have you tried to build a wheel in a build stage and only import / install the wheel ? Maybe this will be cleaner.

https://github.com/jooola/debouncer/blob/2e87c6b51822915176ff1738b978680104b38ea5/Dockerfile#L6-L20

jooola (commented):
Hey, some early review how you don't mind.

Awesome that you started doing this, it was also my intent you beat me to it.

I left a few comments. And maybe we should pin the python release version as well.

And I would not recommend using bullseye, I mean we need to move forward and fix the codebase, but I reverted my migration to bullseye in production as I  was getting way too much bugs.

analyzer/Dockerfile
jooola: Liquidsoap does actually have a alpine package https://github.com/savonet/liquidsoap/issues/1646

paddatrapper: I really dislike the alpine images, simply because they are very different and often cause issues (like PATH search failures). The Debian images are well known and easy to modify. Also all our tooling around package names is centered around Debian/Ubuntu. I want to re-use that instead of adding an additional package name for all the alpine things that then change and get dropped very rapidly from their archive.

analyzer/Dockerfile
paddatrapper: wget will not check the cache intelligently. According to https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#leverage-build-cache, ADD will compare the checksum of the URL and the dest and invalidate the cache if they are different. However, RUN will never invalidate the cache on subsequent runs, potentially leaving the file out of date

analyzer/Dockerfile
jooola: Yes, I get what you mean. Maybe we should consider this only once we have fully working images and less dependencies. I think having lightweight containers is still a good thing. But should not generate headache.

analyzer/Dockerfile
paddatrapper: Not exactly cleaner, but reduces the image size slightly

paddatrapper:
Building individual docker images is tricky without #1504 because the build context is each app. So inter-dependencies (e.g. on api_client) are not supported. They can only be installed from the web and not from the local repo, making pypi the logical choice. It does mean that changes to shared/api_client require a version bump before the changes can be used in any application. This is because building will always use the Pypi version, so new changes need pushing there before being used

jooola:
I would recommend importing a wheel file from a build stage over fetching packages from pypi.

I used to release some docker images like this, but by the time the pip package was available on pypi, the published docker image was using the previous version. Also this would not allow use to build images that haven't been published yet. For instance having a image from master for the testing/demo feature would be infeasible.

What I propose is to create a api_client and shared dockerfile only with a build stage, run those before we build the apps, and import the wheels from there.

May I push some ideas to this branch ?

jooola:
Pushed a extra commit with some idea, feel free to prune the commit if you don't want it.

analyzer/Dockerfile.jooola
```diff
@@ -0,0 +1,39 @@
...
+# TODO: Rename packages.ini common section to python
+# TODO: Properly handle python3 setup provided by the initial docker image and
+#       our dependencies automagically installing python from apt.
+#       This mainly has to do the python Gstreamer dependencies.
+RUN python3 /tmp/packages.py \
+    --format=line buster \
+    --exclude=common \
+    /tmp/packages.ini > packages.list
+
+COPY . .
```
paddatrapper, line 17: This will result in dependencies being downloaded and installed every time there is a code update, which makes developing using the docker images slow

analyzer/Dockerfile.jooola
```diff
@@ -0,0 +1,39 @@
...
+
+RUN pip wheel --wheel-dir dist --no-deps .
+
+FROM python:3-slim-buster
+
+COPY --from=analyzer-build /build/packages.list /tmp/packages.list
+
+RUN apt-get update && \
+    apt-get upgrade -y && \
+    apt-get install -y $(cat /tmp/packages.list) && \
```
paddatrapper, line 27: This needs to be run in the build step for some cases, for example analyzer dependencies requires system packages to build

analyzer/Dockerfile.jooola
jooola: Gotcha. In that case would a requirement.txt not be welcomed here ?

analyzer/Dockerfile.jooola
jooola: Sorry I don't understand what you mean

paddatrapper: We need to install system package requirements in the build image for some components, as they are required to install/build pip dependencies

paddatrapper:
TODO: update ADD to use main instead of master

analyzer/Dockerfile.jooola
jooola: If you are thinking about Pycairo, I was able to build the wheel without having the system packages. Or maybe I got lost in the dependency stuff. I 'll need to some research.

analyzer/Dockerfile.jooola
paddatrapper: Definitely. Though a way of doing it without maintaining two lists of dependencies would be great

jooola: Maybe https://github.com/bndr/pipreqs cal help with that, generate it on fly inside docker or as make dependency ?

paddatrapper: We have a working way of generating the requirements.txt already in the Dockerfile - https://github.com/LibreTime/libretime/pull/1471/files#diff-cfe7fb4506860afbb02571526ef58c47331b85e2625ae9ee3c403e4817ee9853R12-R22

jooola: I don't like this piece of code. Not flexible enough and we need to duplicate this in few apps.
Any deviation would require extra change. For instance the api has some extra require that aren't 'dev'.

jooola: Solving #1154 can help for this. Copying the pyproject + lock file will make dependency installation easier for the docker build, and in some other areas too.

analyzer/Dockerfile.jooola
paddatrapper: PyGObject fails to install during build:

```
Building wheels for collected packages: pycairo, PyGObject
  Building wheel for pycairo (setup.py): started                                                         
  Building wheel for pycairo (setup.py): finished with status 'done'
  Created wheel for pycairo: filename=pycairo-1.19.1-cp310-cp310-linux_x86_64.whl size=276674 sha256=30e9797f205a6e877f544c2468e8d40604e4bbdb1f71eb28ea36e349d83590aa
  Stored in directory: /tmp/pip-ephem-wheel-cache-_pqz7v43/wheels/fe/05/de/41f9367a9b16909345795817fbe2a78adee3453126a21e60b1
  Building wheel for PyGObject (PEP 517): started
  Building wheel for PyGObject (PEP 517): finished with status 'error'                        
  ERROR: Command errored out with exit status 1:                                                         
   command: /usr/local/bin/python /usr/local/lib/python3.10/site-packages/pip/_vendor/pep517/in_process/_in_process.py build_wheel /tmp/tmp0d29s0ke
       cwd: /tmp/pip-install-23t7nh4o/pygobject_30ac5eccb97542e5a48265165795f8b1                                                                                                                                  
  Complete output (44 lines):                                                                                                                                                                                     
  running bdist_wheel
  running build
  running build_py
  creating build
  creating build/lib.linux-x86_64-3.10
  creating build/lib.linux-x86_64-3.10/pygtkcompat
  copying pygtkcompat/pygtkcompat.py -> build/lib.linux-x86_64-3.10/pygtkcompat
  copying pygtkcompat/generictreemodel.py -> build/lib.linux-x86_64-3.10/pygtkcompat
  copying pygtkcompat/__init__.py -> build/lib.linux-x86_64-3.10/pygtkcompat
  creating build/lib.linux-x86_64-3.10/gi
  copying gi/types.py -> build/lib.linux-x86_64-3.10/gi
  copying gi/_ossighelper.py -> build/lib.linux-x86_64-3.10/gi
  copying gi/pygtkcompat.py -> build/lib.linux-x86_64-3.10/gi
  copying gi/_gtktemplate.py -> build/lib.linux-x86_64-3.10/gi
  copying gi/_error.py -> build/lib.linux-x86_64-3.10/gi
  copying gi/docstring.py -> build/lib.linux-x86_64-3.10/gi
  copying gi/_constants.py -> build/lib.linux-x86_64-3.10/gi
  copying gi/_propertyhelper.py -> build/lib.linux-x86_64-3.10/gi
  copying gi/_option.py -> build/lib.linux-x86_64-3.10/gi
  copying gi/module.py -> build/lib.linux-x86_64-3.10/gi
  copying gi/__init__.py -> build/lib.linux-x86_64-3.10/gi
  copying gi/_signalhelper.py -> build/lib.linux-x86_64-3.10/gi
  copying gi/importer.py -> build/lib.linux-x86_64-3.10/gi
  creating build/lib.linux-x86_64-3.10/gi/repository 
  copying gi/repository/__init__.py -> build/lib.linux-x86_64-3.10/gi/repository
  creating build/lib.linux-x86_64-3.10/gi/overrides
  copying gi/overrides/GdkPixbuf.py -> build/lib.linux-x86_64-3.10/gi/overrides
  copying gi/overrides/Gdk.py -> build/lib.linux-x86_64-3.10/gi/overrides
  copying gi/overrides/Gtk.py -> build/lib.linux-x86_64-3.10/gi/overrides
  copying gi/overrides/GObject.py -> build/lib.linux-x86_64-3.10/gi/overrides
  copying gi/overrides/Gio.py -> build/lib.linux-x86_64-3.10/gi/overrides
  copying gi/overrides/GIMarshallingTests.py -> build/lib.linux-x86_64-3.10/gi/overrides
  copying gi/overrides/GLib.py -> build/lib.linux-x86_64-3.10/gi/overrides
  copying gi/overrides/keysyms.py -> build/lib.linux-x86_64-3.10/gi/overrides
  copying gi/overrides/__init__.py -> build/lib.linux-x86_64-3.10/gi/overrides
  copying gi/overrides/Pango.py -> build/lib.linux-x86_64-3.10/gi/overrides
  running build_ext
  Package gobject-introspection-1.0 was not found in the pkg-config search path.
  Perhaps you should add the directory containing `gobject-introspection-1.0.pc'
  to the PKG_CONFIG_PATH environment variable
  No package 'gobject-introspection-1.0' found
  Command '('pkg-config', '--print-errors', '--exists', 'gobject-introspection-1.0 >= 1.56.0')' returned non-zero exit status 1.
  
  Try installing it with: 'sudo apt install libgirepository1.0-dev'
  ----------------------------------------
  ERROR: Failed building wheel for PyGObject
Successfully built pycairo
Failed to build PyGObject
ERROR: Could not build wheels for PyGObject which use PEP 517 and cannot be installed directly
WARNING: You are using pip version 21.2.4; however, version 21.3.1 is available.
You should consider upgrading via the '/usr/local/bin/python -m pip install --upgrade pip' command.
The command '/bin/sh -c pip3 install --no-cache-dir -r libretime_analyzer.egg-info/requires.txt' returned a non-zero code: 1
```

jooola: Have you tried building the wheel with `--no-deps` ?

jooola:
I rebased this branch.

analyzer/Dockerfile.jooola
paddatrapper: `--no-deps` is not a valid option for `bdist_wheel` command

jooola:
Rebased the branch onto main.

jooola:
Rebased on the main branch.

@paddatrapper Please tell if I should have my own branch if I am messing with your work.

paddatrapper:
No, it's fine. I want to go through and clean this up hopefully this week so it is ready for testing

paddatrapper:
Where do we want to publish docker images? GitHub container repo and Docker Hub?

jooola:
I lean towards ghcr

jooola:
The web installer currently calls `psql` #1642 and `libretime-api` https://github.com/libretime/libretime/pull/1644#issuecomment-1099049028 to run some migrations.

We will need to handle those to make the docker images work.

paddatrapper:
Blocked by #1758

paddatrapper:
Fixes #949 #551 #624

jooola:
I don't necessarily see #1758 as a blocker, but more like the next required step to be able to do advanced deployment.

If the docker images are run behind a reverse proxy, using the public_url and the proxy should be working just fine. We are still limited on more complex context or like you described in the issue, the reverse proxy charges the traffic.

jooola:
#1758 seem complex in my opinion, I already commented on the fact that we don't know when to use the internal/public url as the api endpoints are shared for both internal and public calls.

I haven't tried to run the different services using docker-compose yet. I should do that first maybe I will have a better understanding.

jooola:
For the documentation, I propose to add tabs everywhere we can find systemd/systemctl commands to provide a docker equivalent (where meaningful). I think we can assume the docker documentation will be for basic docker-compose setup, advanced used probably know what they are doing if they play with k8s or similar.

docker-compose.yml
```diff
@@ -28,3 +33,54 @@ services:
...
+      - libretime_data:/srv/libretime
+
+  legacy:
+    build: ./legacy
+    environment:
+      LIBRETIME_LOG_FILEPATH: "php://stderr"
+    volumes:
+      - ./docker/config.yml:/etc/libretime/config.yml:ro
+      - ./legacy:/var/www/html
+      - libretime_data:/srv/libretime/stor
```
jooola, line 73: Default storage path changed to `/srv/libretime`

docker-compose.yml
```diff
@@ -28,3 +33,54 @@ services:
...
+    volumes:
+      - ./docker/config.yml:/etc/libretime/config.yml:ro
+
+  liquidsoap:
+    build: ./playout
+    command: /usr/local/bin/libretime-liquidsoap
+    ports:
+      - "1234:1234"
+    volumes:
+      - ./docker/config.yml:/etc/libretime/config.yml:ro
```
jooola, line 48: Wonder if we should save the liquidsoap generated configuration file to the host.

docker-compose.yml
```diff
@@ -28,3 +33,54 @@ services:
...
+
+  analyzer:
+    build: ./analyzer
+    volumes:
+      - ./docker/config.yml:/etc/libretime/config.yml:ro
```
jooola, line 53: Missing storage mount.

paddatrapper: It downloads via REST, so doesn't need the storage mount

jooola: I don't think it's downloaded through the HTTP API, a path is send through rabbitmq and that's all. I think only playout downloads through the HTTP API endpoint.

jooola:
With a little hack we might be able to merge this before 3.0.0.

Since we can override the public url using env variables, we can override the public url for the playout/liquidsoap containers by adding a `LIBRETIME_GENERAL_PUBLIC_URL` en variable with the proper host + port in the docker-compose file.

The only problem is the generated urls on the legacy side that are pushed to the analyzer for example. But thats not a huge problem.

@paddatrapper Should we try this ?

The docker setup will be marked as experimental as we should solve this properly in the end.

paddatrapper:
Yes, that is a very good idea

jooola:
Pushed a few changes:
- improved the initial docker-compose spec with healthcheck, updated/missing volumes, formatting and cleanup
- added a single multi-stage dockerfile to build all the different conatiners (Note that some improvements in the multi-stage dockerfile are not in the initial splitter dockerfiles, so if we drop the multi-stage idea, we should update the initial dockerfiles)
- added a ci job that build from this multi-stage dockerfile and tags the containers properly. 
- temporary fix for liquidsoap not binding to 0.0.0.0 by default, this should not be merged as we should make this configurable.

Let me know if you think this multi-stage dockerfile is a better idea, so we can prune the previous dockerfiles.

jooola:
We will need both a production and a dev docker-compose stack.

Should we have a the main `docker-compose.yml` file as dev or prod spec ? In other words: should we have an extra `docker-compose.dev.yml` or `docker-compose.prod.yml` ?

I think we will add stuff for development such as the postgres and rabbitmq ports, or mounting local dirs for live code reloading so maybe the dev stack is the extra spec. But not having to specify the multiple files every time we run the docker-compose command would make our life easier.

Though we might be able to use the COMPOSE_FILE env var and a .env file for development, and specify all the compose files in there.

So maybe:
- docker-compose.yml => prod
- docker-compose.dev.yml => dev

@paddatrapper Thoughts about this ?

paddatrapper:
I think that is the way we should go. So have docker-compose.dev.yml for development

jooola:
@paddatrapper What do you think of merging this as WIP ? While the feature should not be used by the public yet I see this very useful for testing and developing ?

paddatrapper:
Yeah, I think we can merge this as a WIP. Others can get in on finalising it then too

jooola:
I was wondering: for production deployment, we only want to be copy the compose file, and few extra files (libretime config + nginx config) ? We don't want to clone the whole repo ?

paddatrapper:
Maybe. I think for prod people will build their own docker-compose scripts or use other orchestration platforms anyway

jooola:
@paddatrapper I think we are almost done here. I would like to have your input on the changes.

Do you think it's ok if we ship a pip editable install in production ? I don't see any real problem doing so and it should greatly help the dev setup by simply mounting the source files.

I think we can start by publishing internal/dev/main images to gchr.io and once we are confident, we can publish tagged releases to docker hub. I would use ghcr.io for testing and branches images.

At some point we might want to install Make in the python images, so we can run tests inside the docker setup, but that's for future improvements.

As conventions I tried to do the following for the python apps (the legacy image is appart as it is php-fpm):
- Storage is in `/srv/libretime`
- Source code is in `/src`
- Working directory (which means runtime files) are in `/app`

The legacy image have everything under `/var/www/html` and the nginx document root is at `/var/www/html/public`.

We have 2 .env files used to configure the docker-compose stack, if none is present, every variables has production defaults.

See the docker-compose.override.yml head for the dev setup documentation.

jooola:
The only thing I don't understand is why playout is hanging on shutdown, I think the way threads are handled is really dirty, I'll try to fix this.

jooola:
So I've been playing and developing with this for the past days, and it is quite pleasing to use ! Everything is working as expected, and a "production" demo is available in docker/example (since no image are published yet, you have to build the images first and override the `LIBRETIME_VERSION=main` to run them).

This is obviously not bullet proof yet but haven't had any bug so far. This can be considered as the work in progress I was referring some days ago, as we will have to fix bugs, improve the documentation and maybe change installation procedures.

With regards to inter container dependencies, I am not yet sure if we want to wait for others services apart from the database and maybe rabbitmq. For now we have the dependencies set, maybe I'll remove them in the future if it improves the startup time, and they are properly handled.

paddatrapper:
Awesome! Thanks for all the hard work. I'll try test and review on Monday

jooola:
Rebased onto the main branch.

jooola:
> I'm getting the following error running from the root of the project:
> 
> ```
> $ docker-compose run --rm api libretime-api migrate
> ERROR: The Compose file './docker-compose.override.yml' is invalid because:
> services.api.environment.LIBRETIME_DEBUG contains true, which is an invalid type, it should be a string, number, or a null
> ```

Which version of docker-compose are you using ?

paddatrapper:
Fixed by quoting the variable. Running docker-compose 1.29.2

jooola:
> Fixed by quoting the variable. Running docker-compose 1.29.2

Ok makes sense, I've only been working with docker-compose 2.x.

docs/admin-manual/setup/install.md
```diff
@@ -207,13 +208,62 @@ sudo systemctl start libretime.target
...
+source .env
+
+wget "https://raw.githubusercontent.com/libretime/libretime/$LIBRETIME_VERSION/docker-compose.yml"
+wget "https://raw.githubusercontent.com/libretime/libretime/$LIBRETIME_VERSION/docker/nginx.conf"
+wget "https://raw.githubusercontent.com/libretime/libretime/$LIBRETIME_VERSION/docker/config.yml"
+```
+
+### Setup
+
+Once the installation is completed, edit the [configuration file](./configuration.md) at `./config.yml` to fill required information and to match your needs.
```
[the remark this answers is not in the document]
paddatrapper: Also should mention things that need to be particular values for Docker to work. For example DB and liquidsoap hosts.

docs/admin-manual/setup/install.md
```diff
@@ -19,19 +19,15 @@ If you are coming from **Airtime**, please follow the [Airtime migration guide](
 You can install LibreTime using the one of the following methods:
 
 - [:rocket: Using the installer](#using-the-installer)
-- :construction: Using Ansible
+- [:construction: Using docker-compose](#using-docker-compose)
```
paddatrapper, line 22: Maybe :rocket:`? My thinking is because the instructions are complete and there, even though they may not be stable

jooola: Done

docs/admin-manual/setup/install.md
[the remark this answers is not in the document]
jooola: The config values are already properly set, I added a note that those shouldn't be changed.

paddatrapper:
LGTM - tested and seems to work for me now. I'm excited to try this within k8s once we have built images and see how things go

jooola:
> LGTM - tested and seems to work for me now. I'm excited to try this within k8s once we have built images and see how things go

Awesome ! As we discussed earlier, I expect this to be somewhat of a work in progress. I hope that the feedback from the community will help use iron out the docker setup.

I'll merge in that case.

jooola:
Oh but now that you mention it, should we publish a image for the main branch ? Or only tagged release ? I only considered the tagged releases for now.

PR #1472, opened 2022-01-03
ci: add semantic pull request linting

jooola:
This enforce semantic pull request messages before merging.

Do we want to require a scope or do we allow un-scoped prefixes ?
Did I miss a scope in the allowed scopes ?

The subjects should not start with a uppercase letter.

### DIFF
--- a/.github/workflows/lint-pr.yml
+++ b/.github/workflows/lint-pr.yml
@@ -1,0 +1,36 @@
+name: Lint PR
+
+on:
+  pull_request_target:
+    types:
+      - opened
+      - edited
+      - synchronize
+
+jobs:
+  main:
+    name: Validate PR title
+    runs-on: ubuntu-latest
+    steps:
+      - uses: amannn/action-semantic-pull-request@v3
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          scopes: |
+            analyzer
+            api
+            api_client
+            installer
+            legacy
+            playout
+            ui
+            worker
+
+            ci
+          subjectPattern: ^(?![A-Z]).+$
+          subjectPatternError: |
+            The subject "{subject}" found in the pull request title "{title}"
+            didn't match the configured pattern. Please ensure that the subject
+            doesn't start with an uppercase character.
+          wip: true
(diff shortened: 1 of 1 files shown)


.github/workflows/lint-pr.yml
```diff
@@ -0,0 +1,36 @@
...
+            analyzer
+            api
+            api_client
+            installer
+            legacy
+            playout
+            ui
+            worker
+
+            ci
```
paddatrapper, line 29: ci should be a type, not a scope

.github/workflows/lint-pr.yml
```diff
@@ -0,0 +1,36 @@
...
+          scopes: |
+            analyzer
+            api
+            api_client
+            installer
+            legacy
+            playout
+            ui
+            worker
+
```
paddatrapper, line 28: Is this intentional to allow blank scope?

.github/workflows/lint-pr.yml
```diff
@@ -0,0 +1,36 @@
...
+            ci
+          subjectPattern: ^(?![A-Z]).+$
+          subjectPatternError: |
+            The subject "{subject}" found in the pull request title "{title}"
+            didn't match the configured pattern. Please ensure that the subject
+            doesn't start with an uppercase character.
```
paddatrapper, line 34: Maybe describe or link to expected format?

.github/workflows/lint-pr.yml
jooola: Nope, I wanted to split our project components  from abstract scope, but I just found out ci is actually a type.

.github/workflows/lint-pr.yml
jooola: Yup, noted

jooola:
Fixed the few comments

jooola:
I feel like we should require a scope for every Pr commit messages. As it is difficult not to forget a scope for some component specific changes. Looking at the history we merged some commits with the scope 'shared' some without but should have.

If we don't enforce such structure we will either have to sort the changes logs in each scope/component, or we don't care about having changes structured in a coped sections.

paddatrapper:
I have no strong feelings either way, so happy to enforce it

PR #1473, opened 2022-01-04
ci: pin action-semantic-pull-request version

jooola:
This action requires a full version.


jooola:
Hot fix from a prior PR, sorry for merging this on my own

PR #1491, opened 2022-01-06
feat: enhance libretime shared

jooola:
- The path to the configuration file can be changed using env variables through the cli options. So maybe it isn't required to add this helper function.
- A default config filepath constant was created in order to prevent any possible duplication.


paddatrapper:
Didn't realise that it was a quick change, LGTM

jooola:
CI is failing because of network issues. But the runs on the shared packages are valid.

PR #1492, opened 2022-01-06
feat: change config file format to yaml

jooola:
This PR will allow the legacy app to load a yaml config file and replace the old config mechanisms.


jooola:
This legacy app is a pain...

jooola:
This is part of #420 and #1378.

The idea is to improve the configuration management so we are able to add/remove configuration keys without worrying too much. This make the legacy app capable of reading our new planned configuration file format.

In other words, it will be a bigger pain if we wait for legacy to be removed before we start improving the python apps.

jooola:
Depends on #1543

jooola:
Closing in favor of #1821

PR #1493, opened 2022-01-07
feat(shared): let user provide the log level

jooola:
Simplify the logger setup, as choice should be
left to the user to import or guess  the log level from verbosity.


jooola:
@paddatrapper Woups, changed the default tracking branch for my main branch, pushed to upstream/master...

jooola:
Didn't even use the right prefix... Can this be undone ? Or is the master branch locked ?

PR #1495, opened 2022-01-07
feat(playout): enhance playout logging

jooola:
feat(playout): enhance playout logging

Some initial work on modernizing the playout app. This replace any custom logger or
logging based logger with the logging tools from libretime_shared.logging and loguru.

Removed all the thread/function assigned logger (self.logger = ...), as this makes it
part of the logic (passing logger though function args) as it should not.

Of a dedicated logger is required for a specific task, it should use
the create_task_logger function.

- refactor: remove dead code
- refactor: remove py2 specific fix
- feat: remove unused test command
- feat: setup shared cli and logging tools
- feat: replace logging with loguru
- feat: setup shared cli and logging tools for notify
- fix: warn method deos not exist
- feat: make cli setup the entrypoint
- fix: install shared modules globally in production
  use extra_requires to load local packages in dev environement
- feat: configure log path in systemd service
- feat: default behavior is to log to console only
- feat: create log dir during install
- chore: add comment
- fix: don't create useless dir in install
- fix: move notify logs to /var/log/libretime dir
- fix: update setup_logger attrs
- style: linting
- fix: replace verbosity flag with log-level flag
- feat: use shared logging tool in liquidsoap
- fix: pass logger down to api client
- feat: allow custom log_filepath in liquidsoap config
- chore: add pylintrc to playout
- refactor: fix pylint errors
- feat: set liquidsoap log filepath in systemd service
- fix: missing setup entrypoint update

BREAKING CHANGE: for playout and liquidsoap the default log file path changed to None
and will only log to the console when developing / testing. Unless you are running the
app as a systemd service (production) the default logs filepaths changed:
from "/var/log/airtime/pypo/pypo.log" to "/var/log/libretime/playout.log" and
from "/var/log/airtime/pypo-liquidsoap/ls_script.log" to "/var/log/libretime/liquidsoap.log"

BREAKING CHANGE: for playout-notify the default log file path changed
from "/var/log/airtime/pypo/notify.log" to "/var/log/libretime/playout-notify.log"

### DIFF
--- a/playout/install/logrotate/libretime-liquidsoap.conf
+++ b/playout/install/logrotate/libretime-liquidsoap.conf
@@ -1,4 +1,4 @@
-/var/log/airtime/pypo-liquidsoap/ls_script.log {
+/var/log/libretime/liquidsoap.log {
   compress
   rotate 10
   size 1000k
--- a/playout/install/systemd/libretime-liquidsoap.service
+++ b/playout/install/systemd/libretime-liquidsoap.service
@@ -2,6 +2,8 @@
 Description=Libretime Liquidsoap Service
 
 [Service]
+Environment=LIBRETIME_LOG_FILEPATH=/var/log/libretime/liquidsoap.log
+
 ExecStart=/usr/local/bin/libretime-liquidsoap
 User=libretime-playout
 Group=libretime-playout
--- a/playout/libretime_liquidsoap/generate_liquidsoap_cfg.py
+++ b/playout/libretime_liquidsoap/generate_liquidsoap_cfg.py
@@ -1,13 +1,15 @@
-import logging
 import os
 import sys
 import time
 import traceback
+from pathlib import Path
+from typing import Optional
 
 from libretime_api_client.version1 import AirtimeApiClient
+from loguru import logger
 
 
-def generate_liquidsoap_config(ss):
+def generate_liquidsoap_config(ss, log_filepath: Optional[Path]):
     data = ss["msg"]
     fh = open("/etc/airtime/liquidsoap.cfg", "w")
     fh.write("################################################\n")
@@ -34,31 +36,31 @@ def generate_liquidsoap_config(ss):
         fh.write("ignore(%s)\n" % key)
 
     auth_path = os.path.dirname(os.path.realpath(__file__))
-    fh.write('log_file = "/var/log/airtime/pypo-liquidsoap/<script>.log"\n')
+    if log_filepath is not None:
+        fh.write(f'log_file = "{log_filepath.resolve()}"\n')
     fh.write('auth_path = "%s/liquidsoap_auth.py"\n' % auth_path)
     fh.close()
 
 
-def run():
-    logging.basicConfig(format="%(message)s")
+def run(log_filepath: Optional[Path]):
     attempts = 0
     max_attempts = 10
     successful = False
 
     while not successful:
         try:
-            ac = AirtimeApiClient(logging.getLogger())
+            ac = AirtimeApiClient(logger)
             ss = ac.get_stream_setting()
-            generate_liquidsoap_config(ss)
+            generate_liquidsoap_config(ss, log_filepath)
             successful = True
         except Exception as e:
             print("Unable to connect to the Airtime server.")
-            logging.error(str(e))
-            logging.error("traceback: %s", traceback.format_exc())
+            logger.error(str(e))
+            logger.error("traceback: %s", traceback.format_exc())
             if attempts == max_attempts:
-                logging.error("giving up and exiting...")
+                logger.error("giving up and exiting...")
                 sys.exit(1)
             else:
-                logging.info("Retrying in 3 seconds...")
+                logger.info("Retrying in 3 seconds...")
                 time.sleep(3)
         attempts += 1
--- a/playout/libretime_liquidsoap/main.py
+++ b/playout/libretime_liquidsoap/main.py
@@ -1,31 +1,33 @@
 """ Runs Airtime liquidsoap
 """
-import argparse
-import logging
 import os
 import subprocess
+from pathlib import Path
+from typing import Optional
 
-from libretime_playout import pure
+import click
+from libretime_shared.cli import cli_logging_options
+from libretime_shared.logging import level_from_name, setup_logger
+from loguru import logger
 
 from . import generate_liquidsoap_cfg
 
 PYPO_HOME = "/var/tmp/airtime/pypo/"
 
 
-def run():
-    """Entry-point for this application"""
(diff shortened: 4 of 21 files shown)


jooola:
Depends on #1496

playout/install/systemd/libretime-liquidsoap.service
```diff
@@ -2,6 +2,8 @@
 Description=Libretime Liquidsoap Service
 
 [Service]
+Environment=LIBRETIME_LOG_FILEPATH=/var/log/libretime/liquidsoap.log
```
jooola, line 5: This is part of system integration config, and I wonder if using a [EnvironmentFile](https://www.freedesktop.org/software/systemd/man/systemd.exec.html#EnvironmentFile=) instead of directly loading Environment inside the service file would make system integration easier. But this would mean a set of files to dispatch on the system (/etc/default/libretime-* or /etc/libretime/env).

For now we have 3 variables, I don't know if this will grow. And most of them should be similar (eg LIBRETIME_CONFIG_FILEPATH).

For now I am happy with this, but we can keep this in mind if it get's out of hand.

playout/libretime_liquidsoap/generate_liquidsoap_cfg.py
```diff
@@ -34,31 +36,31 @@ def generate_liquidsoap_config(ss):
...
-    logging.basicConfig(format="%(message)s")
+def run(log_filepath: Optional[Path]):
     attempts = 0
     max_attempts = 10
     successful = False
 
     while not successful:
         try:
-            ac = AirtimeApiClient(logging.getLogger())
+            ac = AirtimeApiClient(logger)
```
jooola, line 52: I am doing this in some parts of the code, as I don't want to miss any logs, in the future, the api client should remove this attr and directly import a logger from a module.

Note that loguru.logger is compatible with the logging.Logger

playout/libretime_liquidsoap/main.py
```diff
@@ -40,7 +42,8 @@ def run():
         "--verbose",
         script_path,
     ]
-    if args.debug:
-        print(f"Liquidsoap {liquidsoap_version} using script: {script_path}")
+    if log_level.is_debug():
         exec_args.append("--debug")
+
+    logger.debug(f"Liquidsoap {liquidsoap_version} using script: {script_path}")
     os.execl(*exec_args)
```
jooola, line 49: We are using the same log file for the python startup and liquidsoap.

playout/libretime_playout/main.py
```diff
@@ -246,15 +161,11 @@ def run():
 
     liquidsoap_startup_test(telnet_lock, ls_host, ls_port)
 
-    if options.test:
-        g.test_api()
-        sys.exit(0)
-
```
jooola, line 252: This was the only command used from the OptionParser. I removed it as the test_api() function uses non existent function from the api client

playout/setup.py
```diff
@@ -43,5 +43,11 @@
         "pytz",
         "requests",
     ],
+    extras_require={
+        "dev": [
+            f"libretime-shared @ file://localhost/{here.parent / 'shared'}#egg=libretime_shared",
+            f"libretime-api-client @ file://localhost/{here.parent / 'api_client'}#egg=libretime_api_client",
+        ],
+    },
```
jooola, line 51: I would rather have this in the main install_requires, but it seems this leads to problems when installing for production.

I think its a pip version issue, but maybe we could publish these package to pip ?

paddatrapper: We can definitely publish to pip. The only issue there is that it won't help with development versions. pip packages will need to be versioned, so can publish as part of the release CI, but development versions will still need to install from file

playout/install/logrotate/libretime-liquidsoap.conf
```diff
@@ -1,4 +1,4 @@
-/var/log/airtime/pypo-liquidsoap/ls_script.log {
+/var/log/libretime/liquidsoap.log {
```
paddatrapper, line 1: We need to mention the change in logging directory in the release notes for beta.1/alpha.11 whichever comes first

playout/libretime_playout/notify/main.py
```diff
@@ -16,19 +14,25 @@
...
 
 # custom imports
 # from util import *
 from libretime_api_client import version1 as api_client
+from libretime_shared.logging import INFO, setup_logger
+from loguru import logger
 
-LOG_LEVEL = logging.INFO
-LOG_PATH = "/var/log/airtime/pypo/notify.log"
+# TODO: Get log settings from cli/env variables
```
paddatrapper, line 31: Please create an issue to track this once this PR is merged

playout/install/logrotate/libretime-liquidsoap.conf
jooola: This should go in the commit message, I'll add the BREAKING change notice in the PR description, it should be be present in the changelog once we generate it.

playout/libretime_playout/notify/main.py
jooola: Ok

playout/install/logrotate/libretime-liquidsoap.conf
paddatrapper: Even better

playout/setup.py
jooola: This need a proper discussion, as I would even go for publishing each app to pypi. I'll open a ticket.

jooola:
If this has been reviewed and ready to be merged, I'll squash the PR and test the changelog generation for the breaking notice.

Once this is done this can be merged.

paddatrapper:
Changes look good. Just need to test in Vagrant before I'm happy to merge

paddatrapper:
libretime-liquidsoap fails to start with

```
Jan 12 08:00:19 buster libretime-liquidsoap[1787]:     from libretime_liquidsoap.main import run
Jan 12 08:00:19 buster libretime-liquidsoap[1787]: ImportError: cannot import name 'run' from 'libretime_liquidsoap.main' (/usr/local/lib/python3.7/dist-packages/libretime_liquidsoap/main.py)
```

jooola:
Tested the latest changes, everything seems to be working as expected.

I merged the fix in #1506 into this for testing, I'll squash / rebase on master once this is fully reviewed and #1506 is merged in master.

paddatrapper:
Looks good. I'm happy to merge this when you've squashed/rebased

jooola:
Ok, this can be merged, but the small tests I made using [git-chlog](https://github.com/git-chglog/git-chglog) didn't generate exaclty what I expected, but they provide a changelog template we can tweak for our needs.

PR #1496, opened 2022-01-07
feat: replace verbosity flag with log-level flag

jooola:
using a string flag is better when using environment variables.


paddatrapper:
Can you have a look at the failing test?

jooola:
Seems like one of the analyzer test is flaky. It has something to do with splitting the filename containing an extra '_' as it is a random generated string.

PR #1498, opened 2022-01-09
chore: allow no port fowarding in vagrant

jooola:
This allows one to connect to 192.169.10.100 instead of localhost, and reduce port conflict with other services

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -3,18 +3,21 @@
 
 #
 # Tweak guest CPUS and MEMORY using:
+# export VAGRANT_NO_PORT_FORWARDING=true
 # export VAGRANT_CPUS=4
 # export VAGRANT_MEMORY=4096
 # vagrant up debian-buster
 #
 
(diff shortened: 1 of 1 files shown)


paddatrapper:
This should be documented in https://github.com/LibreTime/libretime/wiki/Vagrant-Developer-Setup. Also what is the motivation for this change? You can connect to 192.168.10.100 without needing to disable port forwarding

jooola:
It reduce port conflicts with existing service running on my device. I have to admit that everyone is using 8000 and 8080.

paddatrapper:
Sounds reasonable. By documentation, I mean mentioning the option in the docs. That way people know about it without needing to go through the code

jooola:
Added the bit of documentation, but I would argue that we should move this documentation in the repo so we don't have 2 different place to edit in order to bring a new feature.

paddatrapper:
It was in the repo and moved out so that it was seperate from the user documentation 9n the website

PR #1511, opened 2022-01-13
chore: add auto-apt-proxy to vagrant prepare

### LINKED ISSUE #1396: More caching for packages install in vagrant boxes

We should find a way of caching the packages installation in the vagrant box, so we don't have to run the huge packages installation over and over. Or at least cache the downloaded packages.

- Mount the apt cache folder to prevent downloading over and over the same files
- Build on the fly a new image with only the package installation and use that cached image instead (like docker caching ?)

jooola:
While this should be working out of the box, I had some issues when using `deb.debian.org`, and using the old mirrors urls fixed it. See https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=986356

Fixes #1396

### DIFF
--- a/installer/vagrant/debian.sh
+++ b/installer/vagrant/debian.sh
@@ -1,6 +1,9 @@
 #!/usr/bin/env bash
 
+sed --in-place 's/deb\.debian\.org/cdn-fastly.deb.debian.org/g' /etc/apt/sources.list
+
 DEBIAN_FRONTEND=noninteractive apt-get update --allow-releaseinfo-change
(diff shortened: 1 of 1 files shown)


jooola:
Ok, I am hitting a lot of 502 with this. I'll try to update the source list in a smart way.

jooola:
Waouw, this is now working nicely. No more waiting for package download !

paddatrapper:
Awesome! Please document how to use this in https://github.com/LibreTime/libretime/wiki/Vagrant-Developer-Setup (I assume it is just 'install apt-cacher-ng on your machine before running Vagrant')

jooola:
I added some documentation.

PR #1519, opened 2022-01-17
feat(playout): migrate notify cli to click

### LINKED ISSUE #1502: Add cli / env options to playout notify
See https://github.com/LibreTime/libretime/pull/1495#discussion_r780940127

playout-notify still has hardcoded values for the common options such as log filepath or log level.

We need to set this up with the new shared library. A small rework of the CLI is anyway really welcome for playout notify.

jooola:
Fixes #1502

- replace notify option parser with click
- refactor call from liquidsoap to playout-notify

BREAKING CHANGE: the playout-notify log file in '/var/log/airtime/pypo-liquidsoap/notify.log' was removed and merged into the existing '/var/log/libretime/liquidsoap.log' log file.

### DIFF
--- a/playout/libretime_liquidsoap/1.1/ls_lib.liq
+++ b/playout/libretime_liquidsoap/1.1/ls_lib.liq
@@ -1,7 +1,13 @@
-def notify(m)
-  command = "timeout --signal=KILL 45 libretime-playout-notify --media-id=#{m['schedule_table_id']} &"
+def gateway(args)
+  prefix = "timeout --signal=KILL 45"
+  command = "libretime-playout-notify #{args}"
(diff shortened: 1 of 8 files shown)


jooola:
While this is ready to be reviewed, we need to figure out whether we want the notify logs to be inside the liquidsoap logs or in separate files.

I haven't test this yet, but the changes in the PR will put both liquidsoap and notify logs in the same file (liquidsoap.log).

paddatrapper:
In the liquidsoap logs is probably fine. They are very much inter-connected anyway

jooola:
Ok, I can confirm the behavior, notify is logging in the liquidsoap log file.

PR #1524, opened 2022-01-18
feat(analyzer): load config using shared helpers

jooola:
This loads the config for analyzer using the shared helpers.

- moved the `config_filepath` variable setting to the systemd service.
- some refactoring to simplify the analyzer entrypoint.

:warning: Please do not squash.


jooola:
Tested on vagrant buster.

PR #1527, opened 2022-01-19
feat(shared): allow cli parametrized decorators

jooola:
Allow some customization of the decorator before decorating functions.


paddatrapper:
Can you explain a bit more why this is helpful?

jooola:
The cli_config_option decorator now provides a way to require the option and/or to pass a default value instead of None.
This can be helpful when working with commands that do require the configuration file to be set, and I don't want to copy past the code only for a single parameter.

The cli_logging_option decorator change was only updated to prevent confusion between using `()` or not when applying the decorator, they are now consistent.

Issue #1451, opened 2021-11-27
File based settings


jooola:
**Is your feature request related to a problem? Please describe.**

Many settings are editable from the admin UI and are stored in the database.

I suggest to move most of these settings currently in the DB to a dedicated configuration file, or the one we already have. In my case, this allow easier backup/modifications between deployment, for settings that actually does not change that often. This should also simplify other aspects of the configuration handling.

I am mostly thinking about the following settings :
- Server timezone
- Week start day
- CORS
- Public API toggle
- Streams settings
  - Global stream settings
  - Output streams settings
  - Live streaming settings 

I we want to go all the way down this road, we might also consider moving some other settings to a file based storage, but we should leave it editable from the admin UI.

paddatrapper:
This sounds reasonable to me

zklosko:
The stream settings may change when troubleshooting, so they should still be accessible to admins.

jooola:
The only issue we need to fix to move forward with this, is how to handle handle environment variables when running in a docker environment.

This might become tricky when handling stream settings. We could put a hard requirements on a file based configuration, but I would try to avoid this.

zklosko:
I don't see the harm extra database reads would be doing. I guess they could be Redis variables if we were going to replace RabbitMQ, but I don't see the need.

jooola:
## From cc_pref

| keystr                          | valstr                             |
| ------------------------------- | ---------------------------------- |
| timezone                        | Europe/Paris                       |
| allowed_cors_urls               | https://station.radio.org          |
| live_stream_master_password     | SOME_SECRET                        |
| live_stream_master_username     | source                             |
| live_dj                         | false                              |
| live_dj_source_connection_url   | http://stream.radio.org:8802/show  |
| master_dj                       | false                              |
| master_dj_source_connection_url | http://station.radio.org:8801/main |

## From cc_stream_setting

| keyname                  | value                 | type    |
| ------------------------ | --------------------- | ------- |
| icecast_vorbis_metadata  | true                  | boolean |
| dj_live_stream_mp        | /show                 | string  |
| dj_live_stream_port      | 8802                  | integer |
| master_live_stream_mp    | /main                 | string  |
| master_live_stream_port  | 8801                  | integer |
| output_sound_device      | false                 | boolean |
| output_sound_device_type | ALSA                  | string  |
| s\*\_admin_pass          | SOME_SECRET           | string  |
| s\*\_admin_user          | admin                 | string  |
[...]

PR #1532, opened 2022-01-21
feat(api): load config using shared helpers

jooola:
We are slowly getting there !

After the analyzer and playout, the api sould be able to load from our new config helpers.


legacy/application/configs/conf.php
```diff
@@ -60,11 +60,11 @@ public static function loadConfig()
         $CC_CONFIG['cache_ahead_hours'] = $values['general']['cache_ahead_hours'];
 
         // Database config
-        $CC_CONFIG['dsn']['username'] = $values['database']['dbuser'];
-        $CC_CONFIG['dsn']['password'] = $values['database']['dbpass'];
-        $CC_CONFIG['dsn']['hostspec'] = $values['database']['host'];
         $CC_CONFIG['dsn']['phptype'] = 'pgsql';
-        $CC_CONFIG['dsn']['database'] = $values['database']['dbname'];
+        $CC_CONFIG['dsn']['hostspec'] = $values['database']['host'];
```
paddatrapper, line 64: Will need to add the port option here too

api/libretime_api/settings.py
```diff
@@ -128,7 +141,7 @@
 
 LANGUAGE_CODE = "en-us"
 
-TIME_ZONE = "UTC"
+TIME_ZONE = CONFIG.api.time_zone
```
paddatrapper, line 144: We probably want a way to sync this with the UI config option in the DB at some point

legacy/application/configs/conf.php
jooola: I am unsure if this is easily doable, the legacy codebase totally ignored this field, and it would mean tracking doing every part of the code to fix this. I say that because there is a lot of code duplication and we need to tackle each one of them.

jooola: Ok, this wasn't that difficult, I haven't tested this, but let's hope we won't have trouble with this. We will only notice bug if people change there database port anyway.

api/libretime_api/settings.py
paddatrapper: We already have a timezone-via-ui setting, so we define it in two places. Also the playout uses the system timezone to determine playout. Not something to fix now, but we do need to find consistency somewhere

jooola: Yes, we should have this timezone setting in a single place. For that reason moving this configuration entry to the configuration file would make this a lot easier to share across services.

This topic is actually part of #1451

jooola:
I wonder how we can handle loading the required configuration only when some command requires the full settings.

For instance the `collectstatic` command seem to only requires the `STATIC_*` settings, but we cannot run it because the config validation raise a missing key error.

But we want to load the postgresql settings from the `Config` when running the tests or the server.

@paddatrapper Any ideas ?

paddatrapper:
> EDIT: Maybe we should not call collectstatic before the configuration is deployed ?

We could do it when we run the migrations using the same process

> Or use a different settings module when running the server ?

This is also possible. We could break the `STATIC_*` config into its own settings file, source that with `collectstatic` and include it into the main config file for use there generally.

jooola:
Do we actually need the django static files to be served with the web server ? I mean the only files that are deployed are either for the admin interface which isn't used, or for the rest framework documentation page. Those few assets could be loaded by uswgi directly or disabled in production.

paddatrapper:
> Do we actually need the django static files to be served with the web server ? I mean the only files that are deployed are either for the admin interface which isn't used, or for the rest framework documentation page. Those few assets could be loaded by uswgi directly or disabled in production.

Hmm... Probably not actually. There aren't that may that are required. Once the frontend changes, that will all need serving from the web-server, but that will be unrelated from the Django set up

jooola:
I've tried to run without deploying the static files, but django does not serve the static file when running in production. We end up with a broken UI.

So we should either disable all the admin and the rest framework static files in production, or deploy them as we currently do and tweak the settings to be able to run without the settings validations.

Not serving the files in the first place seem cleaner but we loose the api browser in production (I am not sure if this is really valuable)

Serving the files, means we need to make the settings module dirty / hacky.

I am unsure what is best for us, @paddatrapper ?

api/libretime_api/settings/_internal.py
```diff
@@ -0,0 +1,134 @@
...
+    },
+    {
+        "NAME": "django.contrib.auth.password_validation.NumericPasswordValidator",
+    },
+]
+
+REST_FRAMEWORK = {
+    "DEFAULT_RENDERER_CLASSES": [
+        "rest_framework.renderers.JSONRenderer",
+    ],
```
jooola, line 71: The default rest_framework browsable API has been removed, the swagger api is still present, even in production.

api/libretime_api/settings/_internal.py
```diff
@@ -0,0 +1,134 @@
...
+    "libretime_api.apps.LibreTimeAPIConfig",
+    "django.contrib.auth",
+    "django.contrib.contenttypes",
+    "django.contrib.sessions",
+    "django.contrib.messages",
+    "django.contrib.staticfiles",
+    "rest_framework",
+    "django_filters",
+    "drf_spectacular",
+]
```
jooola, line 18: The admin application has been removed from the installed apps.

api/libretime_api/cli.py
```diff
@@ -7,6 +7,7 @@
 
 def main():
     os.environ.setdefault("DJANGO_SETTINGS_MODULE", "libretime_api.settings")
+    # os.environ.setdefault("LIBRETIME_CONFIG_FILEPATH", "/etc/airtime/airtime.conf")
```
jooola, line 10: I wonder what would be more practical. Having a default value here simplify the api cli usage on production, but this implies the configuration will always be loaded from the files, and never from the environment variables.

What could change it that even if a config_file path is provided, we don't fail if that file is missing.

api/libretime_api/settings/_internal.py
paddatrapper: That sounds reasonable. As long as we have some sort of generated, browsable documentation of the API, I think that is good enough

jooola: We could automatically enable/disable the default browsable API when running in dev, using the debug flags.

paddatrapper: I'm not sure it is needed, but could be helpful when developing the API

api/libretime_api/cli.py
paddatrapper: I don't quite follow what you mean here?

jooola: Having the path to the configuration file set by default, will allow the user to run the libretime-api commands without having to specify the path to the configuration file.  This means, instead of writing:
```bash
LIBRETIME_CONFIG_FILEPATH=/etc/airtime/airtime.conf libretime-api migrate

# We could simply have
libretime-api migrate
```
The current implementation fails if a filepath to a config file is provided but the file doesn't exist.

So I wonder if we should allow for missing config file even if a file path is provided (using a default or a user provided file path) so that we have the convenience of not having to write the config file path in the CLI and allow users to only use environment variables.

jooola: Also, since we want to have the stream settings in the configuration file, I don't know how one could cleanly configure stream outputs using environment variables, as they are quiet complex.

I am unsure about our goal regarding environment variable configuration.

Currently we can configure LitreTime using both methods, but we might need to force having a configuration file, even if it will only be used for the stream settings.

paddatrapper: Yes, I think having a sensible default is a good idea.

Being able to configure things through environment variables is important for containerized applications. We can use [pyaml_env](https://github.com/mkaranasou/pyaml_env) to read environment into the config file. That would just require we have a straightforward way of translating a config key into an env var. For example `a.b.c` could be `A_B_C`. Lists like  `a.b[c]` can become `A_B_1_C`

api/libretime_api/settings/__init__.py
```diff
@@ -0,0 +1,46 @@
+# pylint: disable=unused-import
+from os import getenv
+
+from ._internal import (
```
paddatrapper, line 4: How are you differentiating between what should be in `__init__` and `_internal`?

jooola: Roughly I thought that internal are dev oriented settings that shouldn't change unless you know what you are doing, and __init__ is the entry point, where advanced users might be able to do some tweaking.

I could revert to the single file settings, those are left overs from my previous experiments with multiple settings modules (prod, minimal).

PR #1538, opened 2022-01-25
chore: update pre commit hooks

jooola:
- Update pre-commit hooks
- add pyupgrade to keep the codebase modern 

As always :warning: Please do not squash

### DIFF
--- a/analyzer/libretime_analyzer/message_listener.py
+++ b/analyzer/libretime_analyzer/message_listener.py
@@ -77,7 +77,7 @@ class MessageListener:
                 self.wait_for_messages()
             except (KeyboardInterrupt, SystemExit):
                 break  # Break out of the while loop and exit the application
-            except select.error:
+            except OSError:
                 pass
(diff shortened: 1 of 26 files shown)


analyzer/libretime_analyzer/message_listener.py
```diff
@@ -142,7 +142,9 @@ def msg_received_callback(self, channel, method_frame, header_frame, body):
         metadata back to the Airtime web application (or report an error).
         """
         logger.info(
-            " - Received '%s' on routing_key '%s'" % (body, method_frame.routing_key)
+            " - Received '{}' on routing_key '{}'".format(
```
paddatrapper, line 145: This could use a f-string

jooola: I fixed the one.

The tools didn't catch everything, maybe for good reasons or incapability. That said, I don't want to go over the entire code base to fix these. I hope that with time and repetitive editing most of the code will be modernized.

PR #1547, opened 2022-01-31
feat(playout): change liquidsoap working dir

jooola:
BREAKING CHANGE: When running liquidsoap as a systemd service, the working directory is now /var/lib/libretime/playout.


paddatrapper:
Can you have a look at https://github.com/LibreTime/libretime/runs/5003107544?check_suite_focus=true?

paddatrapper:
Why are you implementing this change?

jooola:
I already opened an issue for this https://github.com/Delgan/loguru/issues/581, the bug was introduced with the latest version of loguru (updated this weekend).

jooola:
> Why are you implementing this change?

What do you mean ?

paddatrapper:
> What do you mean ?

What do we gain from changing the liquidsoap working directory?

jooola:
This will:
- make it consistent with the rest of the paths defined in #1378 
- prepare to move the **generated** liquidsoap config to /var/lib/libretime/playout/liquidsoap.liq, as it should not be in /etc/airtime/liquidsoap.cfg

Issue #1453, opened 2021-12-02
Webstream is stopping after some time


jooola:
**Describe the bug**

We attempt to stream a remote Icecast stream through Libretime.

Everything is working as expected, but the stream stops after a random amount of time, possibly between 5 et 10 minutes.

Updating the show schedule with an extra track starts the stream again, and it will stop again after some time.

**To Reproduce**

1. Setup a web stream
2. Schedule the stream for playing
3. Listen and wait for silence

**Expected behavior**

The stream should not stop and keep stream as long as the stream duration or the show is not over.

**LibreTime version:** main branch the 2021/12/02
**Installation method:** 

- OS: Debian buster
- Method: ./install -fap

**Additional context**

```
2021/12/02 18:00:35 [server:3] New client: 127.0.0.1.
2021/12/02 18:00:35 [lang:3] dynamic_source.get_id
2021/12/02 18:00:35 [server:3] Client 127.0.0.1 disconnected.
2021/12/02 18:00:35 [server:3] New client: 127.0.0.1.
2021/12/02 18:00:35 [lang:3] dynamic_source.id
2021/12/02 18:00:35 [lang:3] ["Done"]
2021/12/02 18:00:35 [lang:3] ["Done"]
2021/12/02 18:00:35 [server:3] Client 127.0.0.1 disconnected.
2021/12/02 18:00:35 [server:3] New client: 127.0.0.1.
2021/12/02 18:00:35 [lang:3] streams.scheduled_play_start
2021/12/02 18:00:35 [lang:3] dynamic_source.output_start
2021/12/02 18:00:35 [lang:3] timeout --signal=KILL 45 libretime-playout-notify --media-id=2 &
2021/12/02 18:00:35 [server:3] Client 127.0.0.1 disconnected.
2021/12/02 18:00:36 [lang:3] ["Done"]
2021/12/02 18:00:37 [decoder:3] Method "MAD" accepted "audio/mpeg".
2021/12/02 18:00:37 [http:3] Decoding...
2021/12/02 18:00:37 [http:3] New metadata chunk: ? -- Emission - Polemixetlavoixoff.
2021/12/02 18:00:39 [schedule_noise_switch:3] Switch to map_metadata_7793 with transition.
2021/12/02 18:00:39 [lang:3] transition called...
2021/12/02 18:00:39 [switch_7791:3] Switch to cross_7785.
2021/12/02 18:00:39 [map_metadata_7793:3] Inserting missing metadata.
[...]

PR #1557, opened 2022-02-02
fix(playout): webstream stopping on manual schedule fetching

jooola:
Fixes #1453 

:warning: Do not squash


jooola:
Both commit fixes the same problem, the scheduled_now_webstream should not be empty or the webstream will be stopped.

Issue #1528, opened 2022-01-19
Create a control facility for installation/setup


jooola:
In order to better control the setup process with config file generation, database initial migration, I propose to write a python based CLI.

**Regarding the installation script:**

I previously proposed to rewrite the entire ./install script to python but by the time I wrote some parts of it I felt I was building something way too complex.
The complexity came because I wanted to make the installer configurable as much as possible, without using an orchestration tool (I proposed to use such orchestration tool but the idea was rejected as newcomers might have trouble with such tools). So I was basically creating an orchestration tool.

I don't want to replace the install script, but it should install a very common setup, with only a few configurable bits such as the listening port or running user.

For more complex setup, I propose to write a good documentation with code snippets, and/or the `ansible-role-libretime` I've been working on.

The installation process should only install the software, and not assume any configuration yet, as this should be handled in the setup part.

**Regarding the setup part of the installation (config file generation, database migration, storage directory creation):**

The current setup is in the legacy PHP code and I would like to extract this feature as cleanly as possible, but this might involve some big changes.

This setup part should handle the following:
- Prompt the user for the configuration details (with good default) and save the configuration to the libretime config file.
- Prompt the user whether the postgresql database/rabbitmq vhost should be created (default is no).
- Migrate the legacy database schema and default data.
- Configure and update the Icecast passwords in the icecast config file and the database (This was previously handled in both the install script and the php setup code, and this feels hacky).
- Create the storage directory and configure the music dir in the database.

Here are a few things that I am unsure:
- should the initial database migration have a dedicated command, and become a setup step ? In the future the api will have to run the django migration on each install/upgrade, this command would be a way to handle legacy and api migration until the legacy migration are rewritten as django migrations.
- should we move the legacy storage configuration from the database to the configuration file ? This makes the setup process less hacky and does not require use to create a 'MusicDir' during the setup process. And some parts of the configuration is already in the config file, so maybe we should merge everything together.
- #1451 seem to make sens, also regarding the icecast password setup above, which is hacky.
- ...

This step is part of #420 because the configuration file generation is a key part of the installation process.

I would like to get your thoughts about this, as while working on it, I noticed that it might actually be a big change.

paddatrapper:
Why do you think this is the best way to go over something like a web-ui configuration wizard?

> Prompt the user whether the postgresql database/rabbitmq vhost should be created (default is no).

Default should probably be yes if we use the existing install script

jooola:
[...]

PR #1558, opened 2022-02-03
feat(legacy): consolidate constants

jooola:
- simplified entry points
- move and rename tests dirs
- group config/constant/vendor loading in a single preload.php file
- remove trailing slashes from path constants and make paths consistent
- move most of the hardcoded file names and path to constants
- change default storage path to /srv/libretime/storage
- remove unused sentry config path constant
- merge airtime.conf.tmp and airtime.conf.temp files as a same installer config file
- use preload.php in libretime setup forms
- rename some AIRTIME_* constants to LIBRETIME_*


paddatrapper:
Do we want to expose the port in the legacy web setup? Not sure if it is an efficient use of time considering #1528

jooola:
It isn't much work to add an entry to the setup form but that would go in #1559. If #1528 is accepted then yes, I would not bother adding the field. It need to be accepted though :wink: !

paddatrapper:
Right, yes that was the PR I was meant to make that comment on :joy:

PR #1568, opened 2022-02-05
chore: import relases notes from release pages 

### LINKED ISSUE #1535: Move releases notes to docs/
While I never wrote any, I think writing and browsing releases notes on the release page is difficult.

I think we should have these files versioned, and available on the website.

In addition, we are now able to generate a detailed changelog file that will also live in the repository, so we will probably want both changelogs and release notes not too far from each other.

This would also make the release process simpler, with a release being cut by merging a PR.

jooola:
Fixes #1535 

:warning: Please do not squash

### DIFF
--- a/docs/admin-manual/setup/install.md
+++ b/docs/admin-manual/setup/install.md
@@ -65,6 +65,17 @@ git clone https://github.com/libretime/libretime
 cd libretime
 ```
 
+:::info
+
+When upgrading, you can pull the latest changes from the repository before checking out the desired version:
+
+```bash
+cd libretime
+git pull
+```
+
+:::
+
 And checkout the latest version:
 
 <CodeBlock language="bash">
@@ -87,6 +98,12 @@ Additional options can be listed with the following command:
 ```bash
 ./install --help
 ```
+
+:::info
+
+When upgrading be sure to run the installer using the same arguments you used during the initial install.
+
+:::
 
 Once the installation is completed, open [http://localhost:80](http://localhost:80) to complete the [setup wizard](#setup-wizard).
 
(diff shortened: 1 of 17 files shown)


docs/release-notes/_category_.json
```diff
@@ -0,0 +1,3 @@
+{
+  "label": "Release notes"
```
zklosko, line 2: @jooola Instead of doing it this way, I'd recommend we put the release notes in a [separate docs instance](https://docusaurus.io/docs/docs-multi-instance) so we can keep the sections and their search indexes separate.

paddatrapper:
Also this is cause the doc build pipeline to fail

docs/release-notes/_category_.json
jooola: Sorry I am unsure about what you want to accomplish. Could you explain the end goal ? From what I understand we are not heading toward a versioned documentation, so having multiple docs instance does not seems necessary based on the docusaurus docs.

jooola:
We have a lot of broken links the in the releases notes, I'll try to update them, but I feel such document should not be altered too much.

paddatrapper:
> Known issues: We could have a document referencing the know issues and when they are fixed with there work around instead of copy pasting the text in each release note. We also have some workaround instruction in the wiki.

I'm tempted to leave this in the release notes, as the issues are specific to a release (unless they cannot be fixed in a release cycle). So when you install a specific release, you bring with it a specific set of known issues. I'm not sure that can be merged into a single Known Issues page, unless you have a matrix of which affect which releases, at which point it may as well be in the release notes.

> Generic upgrade and install documentation, we have a dedicated upgrade and install documentation, so I think the install documentation in the release notes should be removed. We should only keep the specific tasks to run for a particular upgrade.

Agreed

> Duplicate upgrade procedure: The configuration files merging in alpha.3 required to run some tasks, those task are duplicated across all the releases notes upward, but the changes only occurred on alpha.3. So I propose to force admins to read each release notes before jumping multiple version at once, and same goes for the people coming from airtime.

Agreed. Eventually I think we should support skipping minor and patch releases, but require upgrading only 1 major release at a time. But that comes with a stable release process and support, not now.

jooola:
> I'm tempted to leave this in the release notes, as the issues are specific to a release (unless they cannot be fixed in a release cycle). So when you install a specific release, you bring with it a specific set of known issues. I'm not sure that can be merged into a single Known Issues page, unless you have a matrix of which affect which releases, at which point it may as well be in the release notes.

Maybe we could still reduce duplication and keep links in the release notes to those specific issues.

jooola:
Ok let's handle the known issues in another PR if we really have a need to, this is ready to be merged (please do not squash)

PR #1590, opened 2022-02-10
chore: clean and move logo|icon to website

jooola:


jooola:
README logo is not working yet, please do not merge

paddatrapper:
Having the psd file is important because it is the source code for the logo (even if it is a proprietary format). The multiple formats and sizes are also useful for branding on other sites (e.g. OpenCollective)

jooola:
I propose to use the svg files as source code. Svg can generate picture in at any scale, so those should be the source of truth. Svg is open source and editable using Inkscape.

I'll regenerate the png files

jooola:
This is ready. I feel svg is a modern way of maintaining icons/logo for the web. If you think we should keep some other format in the repository I am happy to generate them (colors variants ?).

But I prefer not to rely on photoshop documents that are not editable by those who can't/don't want to run photoshop.

paddatrapper:
> This is ready. I feel svg is a modern way of maintaining icons/logo for the web. If you think we should keep some other format in the repository I am happy to generate them (colors variants ?).

Sure. We should probably keep the colour variants, but can drop the different sizes because SVG.

> Also if I understand this correctly, we might need to update some links in external services ?

As far as I know, none of them link directly, rather have pictures duplicated from this set of files

jooola:
- Added 2 dark and light color variant for the logo
- Generated png files for logo colored variants
- Recomputed svg paths and re scaled paths inside the document viewport (removed sizes)
- Passed the svg files trough svgo for cleanup

PR #1596, opened 2022-02-14
chore: install bandit and flake8

jooola:
More linters and already catched some ugly hacks.

Please do not squash

### DIFF
--- a/analyzer/libretime_analyzer/status_reporter.py
+++ b/analyzer/libretime_analyzer/status_reporter.py
@@ -118,9 +118,7 @@ def send_http_request(picklable_request, retry_queue):
         bare_request = picklable_request.create_request()
         s = requests.Session()
         prepared_request = s.prepare_request(bare_request)
-        r = s.send(
-            prepared_request, timeout=StatusReporter._HTTP_REQUEST_TIMEOUT, verify=False
-        )  # SNI is a pain in the ass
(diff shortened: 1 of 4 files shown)


analyzer/libretime_analyzer/status_reporter.py
```diff
@@ -163,7 +161,7 @@ def is_web_server_broken(url):
     caused by Airtime or the webserver itself being broken temporarily.
     """
     try:
-        test_req = requests.get(url, verify=False)
+        test_req = requests.get(url)
```
paddatrapper, line 164: This could cause issues with LibreTime being served from a self-signed certificate as described in the docs - https://libretime.org/docs/getting-started/ssl/#deploying-a-self-signed-certificate

jooola: Maybe we should not advice deploying self signed certificates ? These days having a free certificate is easy using lets encrypt.

This seem not consistent with the existing api_client that do not ignore the certificate check ?

jooola: And if we really want to support self signed cert we should use something like https://stackoverflow.com/questions/30405867/how-to-get-python-requests-to-trust-a-self-signed-ssl-certificate using env variables or such.

paddatrapper: I agree that let'sencrypt is better. Maybe let's drop it and just note that it is a breaking change?

PR #1592, opened 2022-02-11
feat(analyzer): replace pika with kombu

jooola:
This PR replace pika with kombu while rewriting some parts of the analyzer.

The analyzer now listen on a `analyzer.pipeline` queue, with a dedicated routing key, and process the incoming file.
The result is send back to the message queue in a result queue `analyzer.result` with a dedicated routing key.

The result queue will send a report each time a message is received on the `analyzer.result` queue. The message are persisted on rabbitmq instead of a file storing pickled requests.

When a report request fail because of unreachable api, the request will be retried a few times, and finally the message will be re-queued back to rabbitmq. A sleep period can be introduced before fetching messages from the queue, if the previous call failed.

This workflow basically re implement a celery like task manager.

The entire pipeline now rely on a context object storing errors, metadata, status, and possible configuration. This allow for more complex processing options, while simplifying argument passing.

Fixes #1278

PR #1603, opened 2022-02-15
feat(analyzer): rework organise_file using pathlib

jooola:
The new behavior for duplicate file is now:
```
    If you import multiple copies of the same file, the behavior is:
    - The first filename is preserved.
    - The next filenames receive the current date append to the name.
```

- use datetime instead of uuids
- massively using pathlib to manipulate paths
- append datetime is now in iso format

Related to #1592

### DIFF
--- a/analyzer/libretime_analyzer/pipeline/organise_file.py
+++ b/analyzer/libretime_analyzer/pipeline/organise_file.py
@@ -1,119 +1,54 @@
-import errno
-import os
 import shutil
-import time
-import uuid
+from pathlib import Path
+from uuid import uuid4
 
 from loguru import logger
 
+MAX_DIR_LEN = 48
+MAX_FILE_LEN = 48
 
-def organise_file(audio_file_path, import_directory, original_filename, metadata):
-    """Move the file at audio_file_path over into the import_directory/import,
-    renaming it to original_filename.
 
-    This analyzer copies a file over from a temporary directory (stor/organize)
-    into the Airtime library (stor/imported).
-
-    If you import three copies of the same file, the behaviour is:
-    - The filename is of the first file preserved.
-    - The filename of the second file has the timestamp attached to it.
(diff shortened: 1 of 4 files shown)


jooola:
Please do not squash

analyzer/libretime_analyzer/pipeline/organise_file.py
```diff
@@ -1,119 +1,56 @@
...
-    # the wrong information for the file we just overwrote (eg. the song length would be wrong!)
-    # If the final file path is the same as the file we've been told to import (which
-    # you often do when you're debugging), then don't move the file at all.
+    dest_path /= orig_filename.stem[0:MAX_FILE_LEN] + orig_filename.suffix
 
-    if os.path.exists(final_file_path):
-        if os.path.samefile(audio_file_path, final_file_path):
-            metadata["full_path"] = final_file_path
+    # Handle when a file already exists
+    while dest_path.is_file():
```
jooola, line 39: I don't think we will have to chance to generate a duplicate uuid4, but we never know !

paddatrapper:
What are the chances of importing the same file twice in the same second? Would uuid not ensure uniqueness more reliably?

jooola:
I initially replaced the date time suffix with an uuid, but then I thought we would loose information.

Using an UUID would make the code simpler, and theoretically safer. But the chances that 2 files with the same name (and with same metadata if any), are uploaded at the same second seem really low. The only possible scenario this could happen is if a bulk import or bulk scan a list of files with the same name, but in that case I expected the analyzer not to run more than 2 or 3 concurrent pipeline.

paddatrapper:
The date is already in the database, so I think having a UUID here is better. The filenames themselves are not actually important in the grand scheme - everything displays things based on metadata in the DB anyway

jooola:
Replaced datetime with uuids

PR #1607, opened 2022-02-16
chore: fix broken links

jooola:
I wonder why [ascap](https://www.ascap.com) is not responding to lychee, while I can access it though my browser.


docs/getting-started/reverse-proxy.md
```diff
@@ -71,14 +71,14 @@ server {
     location / {
         proxy_set_header Host $host;
         proxy_set_header X-Real-IP $remote_addr;
-        proxy_pass http://libretime/;
+        proxy_pass http://192.168.1.10/;
```
paddatrapper, line 74: Maybe add something to https://libretime.org/docs/getting-started/reverse-proxy/#setup about the link between `192.168.1.10` and `libretime` also the `ping` command needs updating

paddatrapper (changes requested):

docs/getting-started/reverse-proxy.md
jooola: Done

paddatrapper (approved):

PR #1609, opened 2022-02-16
fix: use ';' as the deliminator in CSV export

triangle-stream:
Change delimiter from "," to ";" as some metadata - like copyright - could contain a comma that breaks the CSV.

### DIFF
--- a/legacy/public/js/libs/CSVexport.js
+++ b/legacy/public/js/libs/CSVexport.js
@@ -18,7 +18,7 @@ Compress with: http://jscompress.com/
             csvOutput = "",
             csvRows = [],
             BREAK = '\r\n',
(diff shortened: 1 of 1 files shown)


jooola:
Shouln't the data be enclosed in quotes if data contain the delimiter char ?

RobertWaterloo:
Changing the delimiter to a semi-colon is likely to create more problems than it solves. CSV is "Comma Separated Values", with a well-defined structure. 

There's an RFC for it: https://datatracker.ietf.org/doc/html/rfc4180

--Bob.

paddatrapper:
I agree with quoting as the solution here instead of changing the deliminator. If this doesn't already happen, then that is a bug in the CSV generator we have

PR #1624, opened 2022-02-22
feat(api_client): load config using shared helpers

jooola:


jooola:
Please de not squash.

This is ready to be merged.

PR #1626, opened 2022-02-22
feat(api): split api into multiple apps

### LINKED ISSUE #1622: How do we organise/split the api code
This is an issue I've been thinking about a long time now, but since I haven't worked much in the API it wasn't really the moment to raise it.

I wonder how we plan to organize the code in the api service. I think we cannot continue with the current organization which simply packs everything in single files and do not have any logical separation by concerns.

A option is to split each concern by app, for instance having a `core` app with auth, users, version and status calls.
Another app could be `shows` with all the show, show instances, playlist and smart blocks.
Another app might be `storage` with files, but I am tempted to have this into core, while this might include some analyzing tasks.
And so on.
Also we might also require a `legacy` django app which hold all the legacy related tasks such as migrations or command line actions. This might be faded out in a far far future.

We could also use a single app but split concerns in different python modules, but I know that migrations conflicts might become difficult to handle in that case.

There are extensive discussion on the internet on how one should organize his Django project, and I think we should think about this before doing any further changes to the API. 

@paddatrapper What is you idea on this ?

I am leaning toward a multi app architecture, separated by logical concerns, concerns that you maybe described in https://github.com/LibreTime/libretime/issues/935#issuecomment-571940765 but I feel those categories are already too close to the real code. I would see wider categories.

jooola:
Fixes #1622

- Split the api into 4 apps: core, history,  schedule, storage
- Exploded the settings into testing/prod


jooola:
Test are passing, the openapi schema is the same as before, linting is good enough.

I think that's it.

We have 4 main apps:
- core
- history
- schedule
- storage

I don't know if we should move the podcasts related api's in there own app.

We probably want to explode the permission files but this PR is large enough.

paddatrapper:
> I don't know if we should move the podcasts related api's in there own app.

Potentially, but I think we can pull that out at a later point if needed.

> We probably want to explode the permission files but this PR is large enough.

Definitely. There is a lot of expansion I would like to do there, but requires removing the PHP first. For example making group permissions customizable so stations can create custom groups with specific permissions and assign users to those groups.

api/libretime_api/storage/views/third_party_track_references.py
```diff
@@ -0,0 +1,10 @@
+from rest_framework import viewsets
+
+from ..models import ThirdPartyTrackReference
+from ..serializers import ThirdPartyTrackReferenceSerializer
+
+
+class ThirdPartyTrackReferenceViewSet(viewsets.ModelViewSet):
+    queryset = ThirdPartyTrackReference.objects.all()
+    serializer_class = ThirdPartyTrackReferenceSerializer
+    model_permission_name = "thirdpartytrackreference"
```
jooola, line 10: This is the one model I could not really understand it's purpose and where it should be put.

From the DB schema, it has a tight relation between the celery tasks, the storage.File, and the podcast but what purpose it has I still wonder.

Am I right having this in storage and not in the Core app ?

paddatrapper: As far as I know, yes. But I don't really understand it either

jooola:
This is ready to be reviewed and merged.

I exploded the settings modules into testing/prod settings and updated the entrypoint with the new settings module path.

jooola:
I am actually redoing the split, I should have it ready in few hours.

jooola:
@paddatrapper Waiting for the CI to finish and this will be ready to be merged.

jooola:
Rebased to update the test matrix.

PR #1640, opened 2022-02-25
feat(worker): add service log filepath

jooola:
Add worker file logging.


paddatrapper:
@jooola is this good to merge for alpha.11? Would make things consistent across the applications

jooola:
@paddatrapper Agreed, I need to do some extra testing but should be pretty easy, give me one hour and I'll get back whether this is working or not.

PR #1643, opened 2022-02-26
fix(tools): allow multiple exclude flag for packages.py

jooola:


paddatrapper:
Can you fix the lint check to include the tools scope?

PR #1644, opened 2022-02-27
feat: rewrite legacy migrations to django migration

jooola:
Depends on #1626 

A rewrite of the migrations previously managed by the legacy upgrade controller.

Fixes #1377

PR #1645, opened 2022-02-27
test(api): use pytest to run api tests

### LINKED ISSUE #1233: Migrate api tests to pytest
Next up for our migration to pytest #1178, I will migrate the test for the Django API.

Still this requires some changes in order to setup a future proof testing framework. I need some inputs on the following ideas:

- Explode settings.py, into a module with multiple settings for all the different environment. And rely on [`settings.configure()` tool](https://docs.djangoproject.com/en/3.2/topics/settings/#using-settings-without-setting-django-settings-module) for loading the settings depending on the current environment (prod/ci/development). For now everything is hard coded for production and it's not easy to work with.

- Use SQLite3 in memory in development, I don't necessarily want to spin a postgres setup every time I want to run some tests. So I would see:
  - dev tests => sqlite
  - dev e2e => postgresql (vagrant/docker/docker-compose)
  - ci => postgresql

- Split unit testing from e2e testing, and do e2e testing.
- Create some seeders or mocks for the database, so we can create a "real" db easily. (I would go for seeders)

This is what I though while thinking on how I could easily run test for the pytest migration.

jooola:
pytest is able to run both unittest testcases and pytest tests. Using both should ease the developers life.

Fixes #1233


jooola:
Depend on #1644 pytest seem to automatically setup the schema using the migrations.

PR #1659, opened 2022-03-03
feat: create bulk-import tool

jooola:
A small tool to build imports an entire directory of files. It can be used as replacement for the "Watched folders" features.

### DIFF
--- a/tools/Makefile
+++ b/tools/Makefile
@@ -2,7 +2,9 @@ all: lint test
 
 include python.mk
 
-PIP_INSTALL =
+PIP_INSTALL = \
+	requests \
+	types-requests
 PYLINT_ARG = tools
 MYPY_ARG = .
 PYTEST_ARG = .
--- a/tools/bulk_import.py
+++ b/tools/bulk_import.py
@@ -1,0 +1,130 @@
+#!/usr/bin/env python3
+# pylint: disable=missing-class-docstring,missing-function-docstring
+
+import json
+import sys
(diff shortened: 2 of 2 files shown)


zklosko:
How is the watched path set?

jooola:
The script does not watch a folder, it simply does a bulk import. Run `./tools/bulk_import.py --help` to get the usage manual.

paddatrapper:
If we intend to point people to this, it should probably be documented in the admin docs. Also is this using v1 or v2 of the API?

jooola:
It upload the files to the legacy api, as the api v2 does not have a upload endpoint, right ? At least it doesn't trigger the entire import pipeline with the analyzer.

zklosko:
Can we add this to #1620? I think once this is in we can finally release our 3.0 beta.

paddatrapper:
> It upload the files to the legacy api, as the api v2 does not have a upload endpoint, right ?

Not currently by the looks of things. Needs implementing in the FileViewSet. Once we drop the PHP API, we can update the model to use a FileField and abstract most of it to Django to handle

paddatrapper:
LGTM. Needs documentation and then I'm happy to merge

jooola:
I would like to let it sit here for a bit, I have some improvements/fixes in mind and I would prefer no shipping this in alpha.11.

jooola:
I'll reimplement this tool into a django management command.

PR #1677, opened 2022-03-16
docs: unsure we restart service after upgrade

jooola:


jooola:
Crap, missed the typo...

Issue #1185, opened 2021-03-11
Icecast SSL - HOW ?


abanobboles:
I know about the topic would already often spoken, but unfortunately nowhere really described a good solution.

It would be great to have a dummy safe tutorial on how to get a stream running with ssl in Icecast. 

This is essential important!

Thanks

abanobboles:
And yes it is a bug, because all the software is of little use to me if I can not set up something important.

Sorry, but I've been looking for a solution for weeks.

Robbt:
Here is our existing SSL documentation - 
https://libretime.org/docs/reverse-proxy - I think this can be used with Icecast - I haven't tested it yet
https://libretime.org/docs/ssl - this is just for apache and I'd definitely suggesting using certbot vs. a self-signed certificate

abanobboles:
Thank you @Robbt  for you answer, but I tried it.
It did not work for Icecast

paddatrapper:
I use this to do SSL reverse proxy for Icecast:

```
server {
    listen 443 ssl;
    server_name fm.bitcast.co.za;
    ssl_certificate /etc/letsencrypt/live/fm.bitcast.co.za/fullchain.pem;
    ssl_certificate_key /etc/letsencrypt/live/fm.bitcast.co.za/privkey.pem;
    add_header Strict-Transport-Security "max-age=15552000;";
    add_header X-Frame-Options "SAMEORIGIN";
[...]

Issue #1032, opened 2020-05-11
WebCaster support requires SSL liquidsoap harbour


paddatrapper:
Browsers are blocking insecure requests on otherwise secure pages. Thus, something like [WebCaster](https://webcast.github.io/webcaster/) will not work without a SSL liquidsoap harbour. (See https://github.com/webcast/webcaster/issues/7#issuecomment-428494803)

We need to work out if we can offer both a secured and un-secured harbour so that people can use either depending on their needs.

risharde:
Definitely needs to support ssl

jooola:
Now that we have an easy way to customize liquidsoap from the configuration file, I think we can start drafting an advanced setup (not by default and for users experienced with the CLI):
- Add a `liduisoap.certs` config section that accept certificates paths.
- Update the liquidsoap entrypoint generation to setup the ssl certs if provided.
- Add a `secure` boolean switch in the inputs stream config section to enable ssl or not.
- Write a guide on how to set this up, using a certbot certificates.

paddatrapper:
Do you think we will expose both a secure and non-secure port simultaneously?

jooola:
I think everything should be secure, I was mainly answering to the original ticket, that mentioned to have both insecure and secure ports.

Reading the liquidsoap (need some testing to comfirm), we can listen on the same port and have multiple mountpoint. So I don't know if splitting main and show into 8001 and 8002 is a feature or from a limitation. With that in mind we could have 8001 for insecure and 8002 for secure connection.

paddatrapper:
Ok. In that case I think `secure_port` would work over a feature flag

PR #1688, opened 2022-03-18
docs: rewrite reverse proxy guide

jooola:


jooola:
I feel like this documentation is really important, but is currently too poor to properly help users .

I have learned the hard way that icecast should not be reverse proxied... I've been stubborn by not listening what was written on the internet. But I don't like have the streams on other ports that 443 though, and even less without TLS termination.

I think we should not advertise to put icecast behind any reverse proxy, and I was starting to propose in this PR. But maybe we should agree on a recommended way of reverse proxying the different services. (Maybe Icecast 2.5 will support being reverse proxied ?)

I want to fix issues such as #1032, but I don't know if it is possible to reverse proxy these endpoints, as they talk "icy", but it is liquidsoap that is handling the stream. Has someone experimented reverse proxying those endpoints ? Or do we HAVE to use the liquidsoap ssl harbor ?

I would prefer to leave any TLS certificates usage to the reverse proxy (and only on the reverse proxy) and not advertise for complex setups where one has to get multiple certificates for different services (secure icecast for example). I would leave this for power user that don't necessarily need documentation to setup such infrastructure.

Here is a schema of what we could recommend:
[](https://mermaid.live/edit#pako:eNqdk01PwzAMhv-KlSPaxoAdUCU4wWESBzR2QR1CoXVpRJNUjrsPbfvvuLRjHdpAwofGSt7Hzeska5X4FFWkssIvklwTw_Ru5kDCOEZyyPG4TV6a-VC9vZMu81agi_jZVwRhFRgteIKSzFwzgiALTx8tVkdh3gjZWIwf6mwqGQSkuUmwJ4tSwIF3e1l0PbwednjRJTrIjprxGNtKhBx2ybLQK19x_NiMx8hWUpMXPflcdvB99u2-JL9cNdYJ50gBm6kOVUdG3vFrzlyK592vzga41LYscJB4Gx1YPGTCSWg0ujpJQX_Qv4XNBFNDmDCwh69im27lQ5jEOVI8-RpObwf6deEpkjWuPuK2bEPvKXTpsd41sqbGOIOKCjABSKfGd71t9hfgLzYwobZd-Nxq4za7w_wHH3K_-MF_29k9Cbi5ue205dfFcGS1vaSqp6y0UptUHuC61s0U5yi2VSRpipmuCp6pmduKtCpTafl9atiTijJdBOwpXbF_WrlERUwV7kR3Rsslta1q-wkKEUii)

Maybe we should not write that it is possible to expose the playout ports (8001/8002) directly to the internet, as this leads to security issues. For example to auth mechanism is in plaintext, which is kinda bad.

jooola:
Browsing through the issues, I saw that @paddatrapper https://github.com/libretime/libretime/issues/1185#issuecomment-797357155 was able to reverse proxy icecast using nginx, I wonder if you had tstas of the nginx server to check if everything is working as expected, I am currenlty reverse proxying using Haproxy, and I have some problems:
image

jooola:
To answer myself, I noticed that my problem is actually not icecast related but only a bad configuration on the haproxy side. I changed it, will see how it goes.

So maybe we can recommend to put icecast behind a reverse proxy, against all the recommendation not to do so ?

jooola:
- Removed outdated documentation
- Renamed the icecast documentation to streams configuration for future addition of HLS.
- Warning about NOT reverse proxying icecast based protocols (Icecast and liquidsoap).
- We will implement HLS and SRT in the future.

jooola:
> By not putting icecast behind the SSL you will get warnings about unencrypted data on an encrypted site. We probably should add something warning about this. For example the mini-site and the listen button popup

We could document how to setup a secure icecast. Open an extra port and specify path to ssl certs. But maybe a link to an existing tutorial is better ?

paddatrapper:
Yeah, probably linking somewhere official is a better solution

jooola:
Icecast documentation does not provide any instruction on how to setup a secure Icecast server.

I might write one as a tutorial with certbot + renewal + icecast on port 80 + 443. But I'll do that once I have set my own Icecast server.

By the way I saw that icecast 2.5 might implement chunked encoding, which should make it reverse proxy-able. https://wiki.xiph.org/Icecast_Server/known_reverse_proxy_restrictions

PR #908, opened 2019-11-04
Bringing back Watching Folder feature

numericalatina:
[This is still a work-in-process]

I'm opening the pull request to let the community know I'm working on this feature, in case there is someone else making the same effort, so we can concurr.

PR #1693, opened 2022-03-24
feat: add watch folders

kmahelona:
### Description

Watch folder support. The original PR is https://github.com/althost/libretime/pull/4 but I'm moving this to Libretime so we can get this over the line. There's a bunch of work ahead which is meging main into here and sorting out conflicts.

Closes #908 

### TODO: Documentation

_Are there documentation changes required as a result of these changes? See
https://github.com/LibreTime/libretime/wiki/Documentation-Requirements_

### Testing Notes
Watch folders have been running 24/7 at our radio station since the first COVID lockdown.

**What I did:**

_What did you do to validate this PR?_

### **Links**

_Issues links or other related resources. Use the line Closes: #bug_number to
link a bug in the issue tracker._


jooola:
@kmahelona What are your plan regarding this PR ?

We should try to talk about this one before you are putting too much work into it.

jooola:
I think this should go as a django command that analyze files missing from the DB to the analyzer.

Once #1595 is done, passing more options will be straightforward and we can trigger a file analysis without moving the file afterwards, and leaving it inplace.

One could then set a cronjob to scan the storage every night or trigger it manually.

kmahelona:
Did we move to using django (yay!)? I have this feature working pre-django but whatever is most sensible I'm happy to support. From where I left off, this feature was ready to be merged it just needed to be updated with the master branch.

jooola:
@kmahelona Hey, I am gonna close this.

The proposal doesn't comply with the direction LibreTime is taking. Please see the documentation to see why an extra service isn't the way we want to go https://libretime.org/docs/next/developer-manual/design/architecture/.

We currently have a simpler bulk_import command using django #1877 that can be used with a cron job to regularly scan a folder. If you feel like this command is lacking some features, feel free to raise an issue or provide PR.

Anyway thanks for the work. I am open to any further discussions.

kainz:
Does this mean that non-schedule oriented activities are to be considered out of scope?  Reducing that idea to its essence would imply that library and media management should exist as a separate service/interface entirely.  What if libretime were adapted to use a separate backend service for media management as a whole then, such as [mpd](https://github.com/MusicPlayerDaemon/MPD) (not a good fit, i know), or maybe [navidrome](https://github.com/navidrome/navidrome/) (genericish streaming server)?

paddatrapper:
@kainz no one said folder import is out of scope? It is being worked on in #1877 in a way that is more aligned with the architecture of libretime

kainz:
> @kainz no one said folder import is out of scope?

What I'm trying to say is that folder import and metadata management are not the same concern, and when the media library grows enough, bulk scans become less reasonable.  If the logic for this PR were implemented as part of #1877 's bulk import would it not be closed by the same logic that was used to close this PR?

jooola:
> If the logic for this PR were implemented as part of https://github.com/libretime/libretime/pull/1877 's bulk import would it not be closed by the same logic that was used to close this PR?

No, the one of the reason is that this PR ships way too much at once, and duplicate features already in the analyzer (only referring to those not duplicating existing features):
- BPM detection
- Extract artwork
- Scan folders to import the files in the DB
- Maybe some other stuff didn't dig deep enough

The second reason is that we want to reduce the number of services to a minimum and merge everything inside the django framework. Using the Django ORM for example would already greatly simplify the code in this PR.

Though, a lot of code in this PR is interesting, but I will not be merged as it is.

To respond to your first comment, I already thought about this, and it crossed my mind to merge the effort with funkwhale.audio to let them handle the library management, and let us schedule and play the files. But that's only a idea for the time being.

jooola:
Some extra answer.

> For dealing with at least those situations, library management deserves to be detailed in libretime's documented architecture if you aren't abstracting that into a service component, doesn't it?

The micro service architecture is not the architecture we are heading to.

I feel that "watched folders" is a bad name for a set of multiple features:
- Bulk import files in the library (handled in #1877)
- Scan existing files to update the metadata in the DB (should be this PR)

So I am wrong when saying that #1877 can replace this PR.

A scanning feature would be greatly appreciated, but should reuse the code of the analyzer, use the Django ORM, and be embed in tasks run by Celery. And this PR doesn't tick any of those requirements.

Issue #1689, opened 2022-03-21
Playout broken on Ubuntu 20.04


paddatrapper:
See https://github.com/libretime/libretime/issues/1606#issuecomment-1041226485 and https://github.com/libretime/libretime/issues/1606#issuecomment-1046302776

paddatrapper:
I've tested this with stock Ubuntu liquidsoap and can replicate the issue even with the commit introducing Ubuntu 20.04 support to LT. Therefore it definitely appears to be a liquidsoap issue and I am working on getting a fixed version into the PPA

PR #1697, opened 2022-03-26
fix: add PPA for newer liquidsoap version on Ubuntu

paddatrapper:
Fixes: #1689


jooola:
Haven't tested this though, I'll test while doing a global test for the new release.

PR #1705, opened 2022-03-28
chore: ignore docusaurus extra packages upgrade

jooola:
those packages need to be upgraded alongside with @docusaurus/core so we can ignore any individual version bump.


jooola:
Needs to be force merged.

PR #1706, opened 2022-03-28
ci: install git in libretime-dev testing image

jooola:
This might fix a lot of false positive while downloading the
code to the testing folder.


paddatrapper:
You can trigger a re-run of CI tests just of those which fail, which seems to work to ensure that they then actually pass

PR #1708, opened 2022-03-28
fix(installer): test and create correct log path

paddatrapper:


install
```diff
@@ -1027,13 +1027,13 @@ verbose "\n * Installing necessary python services..."
...
+if [ ! -d /var/log/libretime ]; then
   loud "\n-----------------------------------------------------"
   loud "              * Installing Log Files *               "
   loud "-----------------------------------------------------"
 
-  verbose "\n * Creating /var/log/airtime"
-  loudCmd "mkdir -p /var/log/airtime"
+  verbose "\n * Creating /var/log/libretime"
+  loudCmd "mkdir -p /var/log/libretime"
   mkdir_and_chown "$web_user:$web_user" "/var/log/libretime"
```
jooola, line 1037: Duplicate actions, I think we can simply remove the initial mkdir

install
```diff
@@ -1051,9 +1051,7 @@ verbose "...Done"
 
 verbose "\n * Installing playout and liquidsoap..."
 loudCmd "$pip_install ${AIRTIMEROOT}/playout"
-mkdir_and_chown "${web_user}:${web_user}" "${LIBRETIME_WORKING_DIR}/playout"
-loudCmd "mkdir -p /var/log/airtime/{pypo,pypo-liquidsoap} /var/tmp/airtime/pypo/{cache,files,tmp} /var/tmp/airtime/show-recorder/"
-loudCmd "chown -R ${web_user}:${web_user} /var/log/airtime/{pypo,pypo-liquidsoap} /var/tmp/airtime/pypo/{cache,files,tmp} /var/tmp/airtime/show-recorder/"
+mkdir_and_chown "${web_user}:${web_user}" "${LIBRETIME_WORKING_DIR}/playout/{cache,files,tmp}"
```
jooola, line 1054: The subdirs are created by playout, we don't need to create them here anymore.

jooola (changes requested):

jooola (approved):

PR #1723, opened 2022-04-01
chore(api): add API metadata

paddatrapper:


api/libretime_api/settings/_internal.py
```diff
@@ -98,6 +98,12 @@
 
 TEST_RUNNER = "libretime_api.tests.runners.ManagedModelTestRunner"
 
+SPECTACULAR_SETTINGS = {
+    "TITLE": "LibreTime API",
+    "DESCRIPTION": "Radio Broadcast & Automation Platform",
+    "VERSION": "2.0.0",
```
jooola, line 104: Duplicate value, should be taken from API_VERSION

jooola (changes requested):

PR #1727, opened 2022-04-01
chore: rename api_client to api-client

### LINKED ISSUE #1722: Rename api_client to api-client
We should rename the api_client to api-client so that it is consistent with the other python apps

jooola:
Fixes #1722


jooola:
The test requirement aren't met because the requirement use the old name of the tests.

PR #1731, opened 2022-04-04
chore: upgrade pre-commit hooks

jooola:

PR #1732, opened 2022-04-04
chore(shared): pin click to 8.0.*

jooola:
8.1.0 dropped support for python 3.6


jooola:
pre-commit is failing, and requires #1731

PR #1736, opened 2022-04-05
docs: create architecture design document

### LINKED ISSUE #1610: Drafting a design and architecture document
LibreTime is currently missing a design document, to lead the global decision made in the LibreTime development.

I remember a document explaining the different parts of the project some time ago, It was on the old mkdocs website, I'll try to resurrect this document.

## Initial thoughts

From the few issues I read with design decision that have been proposed, some want to build a micro service architecture, others are heading to a monolithic API with an external Playout service.

Right now I feel we are without global vision for this application. And this lead to having the person provide a contribution to impose his idea. While I understand we cannot afford turning down contributions, we still need to have some kind of road map and design decision to prevent the project from being monkey patched every time someone want to implement an fix/feature.

For example, in #1592, I am working to replace pika with kombu in the Analyzer, it might might sens if we want to work in a micro service architecture, but has no point if we merge the analyzer inside a celery/worker task. While we still want to rely on a hardly functional multi tenant feature for the analyzer.

Another example, in #1501, I think we should not drop rabbitmq which gives advanced routing capabilities, but it does makes less sens to have such tool in a monolithic architecture. It seem that we will always have the communication between the API and playout, which is why I feel having rabbitmq will reduce some of the work required for playout. But again, this should be part of a design decision.

I also don't know if we try to keep SAAS feature while we want for any newcomers to be able to install the app. As both don't have the same needs and handling both is sometimes a huge amount of work.

## After the community meeting

During the community meeting, we have discussed the lack of such document with @paddatrapper and we agreed that we should start drafting one.

I don't think the microservice architecture is right way to go, it does not fix or improve any problematic situation in LibreTime. If we want to properly split code by concerns they are plenty of ways to do it.

The multi tenant feature does not seem to be used by anyone that we know, and brings complexity to the project. Spinning an additional analyzer instance for an additional client isn't much overhead in my opinion. I think that 1 client for 1 LibreTime stack is the right way to go, while we can put aside icecast or reverse proxies or such infrastructure tool that still can be easily shared in a multi client perspective.

I see 2 main component in LibreTime, 'Create the schedule' and 'Play the schedule'. The schedule creation component is a monolithic application, with the frontend talking to the django API. The Play block has it's own requirements in terms of logic and uptime, and should be left aside.

[Original schema without playout > api communication](https://mermaid.live/edit#pako:eNp1UttqAjEQ_ZUlTxb0B3wolNqCoGBpoQ_ZRcbNqKF7MxeKqP_e2cxG16r7knOZDJOzcxB5rVCMxcZAs02-JmmV0FdCBRs0AzlnkD2x3vhVofOBXISzVVm3fsUNcoPgcGnzLSpfoHwNPHFbTKKW8ZX2W5u6clgp-d6BngeNli-LaU_5rc0PmlZMvgPsPLp3M0dTwP4yxYLYoxnaytq7UENnzyn0zmtla2iknJ1x1qvQOeZgnZRTBtntRCVaSwkudx49Sjlnmny0NJYrcLACS_akQ61z9SeS0ej5HFf0Ig8m5RV1gkGKfa_Vq4nujhnKOG12Gf_r-ehiF2i3MEyCccmTvQsPdhdmbMyrFpzjTFt657EXwKOKcxMxFCWaErSi5T60F1JBG1BiKsYEFa7BFy4VaXWiUt_Qu_BNaVcbMXbG41CAd_Xnvsoj55qJBtqwksXTHyJ3FlM)

[](https://mermaid.live/edit/#pako:eNp1UttqAjEQ_ZWQJwv6Az4USm1BULC00IfsImMy1tC9mQtF1H_vbLLR3VrzkjPnzMxOzs6Ry1ohn_IvA82OfcyyitEpoYIvNCOxjCB_iHzjN4WWI7EKd8tG3vpNbCANgsO1lTtUvkDxHGLmdsgSl8eS9mxNXTmslHjtQE-DRoun1bzH_NTmG01Lss8AO43qbuZoCjhcp1hRdG-GNrP2LuTQ3VMKvfda2RoaIRYXnPcytEQJ1gkxjyC_nahEa8nB9d6jRyGWMWRvbZjSFTjYgCV51qFWGfwJNpk8XuxKWoqDSH4lnmCgUt8hO5jo3zFDWnQ7qhH_6XmvsDO0W5gYBOHq560Wph-aHvjO4fS1uH9BOS20pcefeq7cy7g04WNeoilBK9r4Y1uQcVqLEjM-JahwC75wGc-qM6X6hh6LL0q72vCpMx7HHLyr3w-VTHHMmWmgtSsjef4FpXUcvQ)

With the design briefly described above, I found answers to some of my questions:

- #1592 is useless if we plan to merge the analyzer as a worker task. This merge is a low priority as the current celery stack (php + celery v5) is blocking any easy improvements.
- #1501 we should keep RabbitMQ anyways, it bring a powerful routing capability needed between the 2 main components, and while we don't head towards a microservice architecture, anyone can plug his own tool in it.

## Please give us feedback

Maybe my ticket does not sound open for changes, but I would love to hear from those who disagree or have better ideas, and discuss with them.

Please note that if someone if pushing towards a particular design (that we should agree on anyway), that person should consider contributing to the project. As I don't see the myself working on something I am not comfortable with and without the person that had the idea in the first place.

cc @LibreTime/Maintainers

@paddatrapper You suggested we post this ticket on discourse instead of Github ? I could link to the discussion thread if Github is not the right place and lock this until we have all our inputs.

jooola:
old: https://mermaid.live/edit#pako:eNp1VNtq3DAQ_RWhpxSS_YClFNpuoYENbEkhD_ISxtLsWkS2HF26hDj_Xl2srL0Xv2jmnKPRaGbkd8q1QLqkO6UPvAHjyN9V1ZHwtdDBHs0Ne8jG9kvGe18ryW_YJq0Rzbj19d5A3xBuEBw-W96g8ArZz-QT1yAp2DZvid8Ba-h79oQ1CeuEOMbTBtkP4C_YiQmfckRrQ27P0Ev2kG3yfXN_oopsjH_OHLR5QcOe0jLhwknZScbJ_XoFb8fbbYJ37W5Rqb1LmrBOGCVfvRRWQ8_Y-tPeThSSIwfrGLvPxpRrlGXs9_qxYJMsS0VePXpkn0X5E90iF-CgBhvo1WhFZtZ1cnf3jQy_hHTD2KEiyF7iQ10zFowElMBzdJbSxTzJ16jLzRhPSfZJ0Ks7J2Mw15wndjXG2KtxxrMTmAX5J6HoSdYvLh1ZtqTC4WK_IEIfOqVBkJ1UOJAzZTr2OAiZPPop0jgFF7kwBeU--U0mtHTrBB7W0jrshnnES4IUNpL0lrZoWpAi_B_eI1bRMOgtVnQZTIE78MpVtOo-gtT3ocgYJ0YbutyBsnhLwTv9-NZxunTGYxGtJISX1I6qj_-wmXoB

[](https://mermaid.live/edit#pako:eNp1VG1r2zAQ_itCnzpo8wPCGGzLYIUUMjLoBzmUs3WJRWXL1ctCqfvfpxcrsRPHX3T3PI_udKeTP2ilONIl3Ut1rGrQlvxdFS3xXwMtHFDfsadk7L4kvHOlFNUd28Q1oAk3rjxo6GpSaQSLL6aqkTuJ7Gf0ia2RZGyXtoTviCV0HXvGkvh1RJzjKY3sB1Sv2PIRH8-IxvizvUAn2FOyyffN44UqsCH-NXNU-hU1e47LiPOZkhONi_o6Ce_n6jbeu1VbUCpno8avI0aKNye4UdAxtj7Zu5FCVFiBsYw9JmPM1dIw9nu9zdjolLkjbw4dslNT_gQ3yzlYKMF4ejVYmTFWaS9nbJuMgE-mgTw8fCP9Ly5sP9xcFiQv8r7fCfNGBHLCKTokm4KT888WRb4GXbq5IXW0ZzKNiFOymxFHszTVXFdxM8Zw4cNDSY5nFuSfgKwnSb-YS5m3xC7j4rAgXB1bqYCTvZDYkytlTHuepkSe_RhpGKVZzo9Sric97Ijmq72A-7UwFtt-GnFOEMMGkt7TBnUDgvufzEfACupfS4MFXXqT4x6ctAUt2k8vdZ1vMobxUpou9yAN3lNwVm3f24ourXaYRSsB_jk2g-rzPylZkIc)

 I probably didn't mention every points that were raised in the Github issue, feel free to add more details to this document !

Fix #1610

### DIFF
--- a/docs/developer-manual/design/architecture.md
+++ b/docs/developer-manual/design/architecture.md
@@ -1,0 +1,108 @@
+---
+title: Architecture
+---
+
+This document explains the design details and goals for the architecture of LibreTime. It describes the result of a [discussion that happened on Github](https://github.com/libretime/libretime/issues/1610).
+
+## Previous architecture
+
+The previous architecture of LibreTime (based on AirTime) was missing a proper separation of concerns. It was build around a legacy MVC app written in PHP, and services in Python to accomplish specific tasks.
+
+## New architecture
+
+Below is the new architecture goal of LibreTime, with a proper separation of concerns.
+
+```mermaid
+flowchart TD
+    users([Users])
+    public([Public])
+
+    subgraph create_schedule[Create the schedule]
+        webapp[Web app]
+        subgraph core[Backend]
+            message_api[Message API]
+            api[Web API]
+            worker[Worker]
+        end
+    end
+
+    subgraph play_schedule[Play the schedule]
+        playout[Playout]
+        liquidsoap[[Liquidsoap]]
+        icecast[[Icecast]]
+        hls[[HLS]]
+    end
+
+    message_queue[[Message Queue]]
+    database[[Database]]
+    storage[[Storage]]
+
+    users --> |Edit| webapp
+
+    webapp --> api
+    api --> database
+    api --> storage
+    api --> message_queue
+
+    message_queue <--> worker
+    worker --> database
+    worker --> storage
+
+    message_queue <--> message_api
+    message_api --> database
+
+    message_queue <--> playout
+    playout <-. via message queue .-> message_api
+    playout --> |e.g. download file| api
+    playout <--> liquidsoap
+    liquidsoap --> icecast
+    liquidsoap --> hls
+
+    public --> webapp
+    public --> |Listen| icecast
+    public --> |Listen| hls
+```
+
+The LibreTime architecture is split into 2 main monolithic blocks `Create the schedule` and `Play the schedule`. Both blocks must be able to scale horizontally.
+
+:::note
+
+A microservice architecture was rejected as it will not fix or improve any aspect of LibreTime.
+
+:::
+
+:::warning
+
+This document tries to focus on creating and playing a schedule, it does not consider features such as monitoring, logging or archiving.
+
+:::
+
+### Create the schedule
+
+This block contains the following components:
+
+- a web API,
+- a worker to run background tasks,
+- a message API to communicate with the `Play the schedule` block, and other services,
+- a web app to interface with the users.
+
+The web API, the worker and the message API rely on the [Django framework](https://www.djangoproject.com/) to handle database, message queue and storage access.
+
+### Play the schedule
+
+Since the `Play the schedule` has its own requirements in terms of logic and uptime, it is handled separately from the `Create the schedule` block. This block needs to be able to be duplicated in a high availability context.
+
+This block contains the following components:
+
+- a Playout app that communicates with the `Play the schedule` block to gather the schedule,
+- a Liquisoap app that plays and mixes the scheduled items, and dispatch them to the delivery services,
+- an Icecast server that delivers a legacy audio stream to the public,
+- a HLS stream that delivers a modern audio stream to the public.
+
+### One setup per radio station
+
+LibreTime is not meant to be used in a multi-tenant architecture, and an entire LibreTime installation should be dedicated to a single radio station. Previous SAAS or multi-tenant features from Airtime should be deprecated or removed.
+
+### Separation of concerns
+
+The `Create the schedule` block must only prepare a schedule, and the `Play the schedule` must only play that schedule. A strong separation of concerns is required between the 2 blocks to allow the `Play the schedule` block to meet its uptime requirements while not depending on the `Create the schedule` in case of a failure. Development will be simplified if both blocks share a single and properly defined protocol.
--- a/docs/developer-manual/README.md
+++ b/docs/developer-manual/README.md
@@ -10,4 +10,4 @@ Welcome to the **LibreTime developer manual**, you should find guides to integra
 
 ## Improve and contribute to LibreTime
 
-- :construction: Work in progress
+- Learn about the [architecture of LibreTime](./design/architecture.md)
--- a/docs/developer-manual/design/_category_.yml
+++ b/docs/developer-manual/design/_category_.yml
@@ -1,0 +1,1 @@
+label: Design


jooola:
The schema is missing the file storage.

paddatrapper:
Unless the worker is going to be merged into the API (by that I mean running from the API, not as an API CLI command that is run independently), the worker should call the API to get things from storage. Otherwise running LibreTime through an orchestration platform like Kubernetes becomes very difficult and expensive, as the storage then needs to be ReadWriteMany instead of standard ReadWriteOne. It also makes changing the storage backend difficult, as we can't just lever Django's existing backend abstraction.

paddatrapper:
Can you elaborate what the Message API is?

jooola:
> Unless the worker is going to be merged into the API (by that I mean running from the API, not as an API CLI command that is run independently), the worker should call the API to get things from storage. Otherwise running LibreTime through an orchestration platform like Kubernetes becomes very difficult and expensive, as the storage then needs to be ReadWriteMany instead of standard ReadWriteOne. It also makes changing the storage backend difficult, as we can't just lever Django's existing backend abstraction.

The worker will be a celery app embed into the django API. This means it will rely on the django ORM and django Storages to manipulate data (not doing so is in my opinion looking for complexity where it could be simple). 

From the code perspective, this will be the same python package (represented as the `backend/core` subgraph).

But the worker will have to be run in an independent process as described here https://docs.celeryq.dev/en/stable/django/first-steps-with-django.html#starting-the-worker-process

How this process is run on production depends on how we implement it, either we wrap everything as a single entity and use a process manager such as supervisord between for example the systemd unit and the multiple apps (gunicorn, celery). This solution could also apply to the docker setup, having supervisord run in docker and behave as a process manager running both gunicorn and celery.

I am unsure about this solution though. On a systemd system, I would rather have multiple system unit running both gunicorn and celery independently. In containers I would also try to separate the 2 for the sake of simplicity and not introduce another layer of process manager (gunicorn is actually also some kind of process manager).

I don't think the kubernetes storage grants edge case should make the code more complex.

> Can you elaborate what the Message API is?

I described this Message API in the proposal issue here https://github.com/libretime/libretime/issues/1610#issuecomment-1059235353

It is a server listening for messages from the message queue, and answering or publishing messages for the other services to consume. As desribed in the comment in the proposal, the message API should be able to both answer a request/reply for the schedule, and to push the schedule to playout.

It probably has to run as a separate process, or we need to find a way to start a thread to spin the messages queue server within the gunicorn/django app.

paddatrapper:
> I am unsure about this solution though. On a systemd system, I would rather have multiple system unit running both gunicorn and celery independently. In containers I would also try to separate the 2 for the sake of simplicity and not introduce another layer of process manager (gunicorn is actually also some kind of process manager).

Being able to spin up multiple workers and scale horizontally is definitely a plus. I also think they should be two separate processes.

> I don't think the kubernetes storage grants edge case should make the code more complex.

Kubernetes is going to be the core way I deploy LibreTime once the Docker containers are working. In general, I do not consider it an edge case as I maintain LibreTime to scratch my own itch. I do not intend to force people to do anything, but I do strongly disagree with anything that makes that particular use-case more difficult. In this particular case, the issue is moot, as the Celery app being within Django means I can configure a object storage backend and they both can talk to that.

jooola:
> Kubernetes is going to be the core way I deploy LibreTime once the Docker containers are working. In general, I do not consider it an edge case as I maintain LibreTime to scratch my own itch. I do not intend to force people to do anything, but I do strongly disagree with anything that makes that particular use-case more difficult. In this particular case, the issue is moot, as the Celery app being within Django means I can configure a object storage backend and they both can talk to that.

Unsure if your comment awaits an answer. 

This discussion sounds like system admins and developers throwing there problems over the wall and blaming the other for not taking into count there requirements :laughing: . DevOps seem to be the bridge between both worlds, and I would rely on CI/CD to automate those annoying tasks.

I do not intend to block any use cases, but as a general rule, I think that when any specific use case affects all the others it make more trouble than good, and the effort should be put in the use case deployment strategies instead of the code affecting everyone.

Sorry, I diverted from the issue at hand... 

I haven't used Kubernetes yet, so I probably don't understand much of the requirements you have. Does the fact that celery uses django (and settings) fixes your use case ?

paddatrapper:
> Does the fact that celery uses django (and settings) fixes your use case ?

Yes, when using an object storage/non-local backend

paddatrapper:
> I would rely on CI/CD to automate those annoying tasks.

Automate what tasks in this case? The storage paradigm is the domain of the code in this case. Unless there is proper management of concurrent reads and writes (race conditions occurring from two processes reading and writing to the same file for example), multi-process, single storage is not a trivial thing to manage.

jooola:
> Automate what tasks in this case? The storage paradigm is the domain of the code in this case. Unless there is proper management of concurrent reads and writes (race conditions occurring from two processes reading and writing to the same file for example), multi-process, single storage is not a trivial thing to manage.
 
Ok, I understand the requirements now. Yes you are right, concurrent access to a FS should be handled carefully. We could handle this with a shared locking system (for example https://github.com/jazzband/django-redis#locks). This will become a requirement anyway if we plan to let Django or Celery scale horizontally.

The fact that Celery will be running as a separate process but inside the Django project with locks should answer your concerns ?

paddatrapper:
> The fact that Celery will be running as a separate process but inside the Django project with locks should answer your concerns ?

Yup

jooola:
@paddatrapper Could you run the deploy command please? I think we have a permission issue of who can trigger the deployement. Seem like I didn't give myself enough permission to run this command.

docs/developer-manual/design/architecture.md
```diff
@@ -0,0 +1,108 @@
+---
+title: Architecture
+---
+
+This document explain the design details and goals for the architecture of LibreTime. It describes the result of a [discussion that happened on Github](https://github.com/libretime/libretime/issues/1610).
```
paddatrapper, line 5: explains

docs/developer-manual/design/architecture.md
```diff
@@ -0,0 +1,108 @@
...
+
+The previous architecture of LibreTime (based on AirTime), was missing a proper separation of concerns. It was build around a legacy MVC app written in PHP, and services in Python to accomplish specific tasks.
+
+## New architecture
+
+Below is the new architecture goal of LibreTime, with a proper separation of concerns.
+
+```mermaid
+flowchart TD
+    manager([Manager])
```
paddatrapper, line 17: This should probably be Staff - not only managers have access to the Web app

docs/developer-manual/design/architecture.md
```diff
@@ -0,0 +1,108 @@
...
+
+:::warning
+
+This document try to focus on creating and playing a schedule, it does not consider features such as monitoring, logging or archiving.
+
+:::
+
+### Create the schedule
+
+This block hold the following components:
```
paddatrapper, line 82: holds or contains may be better

docs/developer-manual/design/architecture.md
```diff
@@ -0,0 +1,108 @@
...
+
+- a web API,
+- a worker to run background tasks,
+- a message API to communicate with the `Play the schedule` block, and other services,
+- a web app to interface with the users.
+
+The web API, the worker and the message API rely on the [Django framework](https://www.djangoproject.com/) to properly handle database, message queue and storage access.
```
paddatrapper, line 89: Move properly to the end of the sentence to avoid the split infinitive

docs/developer-manual/design/architecture.md
```diff
@@ -0,0 +1,108 @@
...
+
+### Play the schedule
+
+Since the `Play the schedule` has it's own requirements in terms of logic and uptime, it is handled separately from the `Create the schedule` block. This block need to be able to be duplicated in a context of high availability.
```
paddatrapper, line 93: its (without the apostrophe)

paddatrapper: Also needs

docs/developer-manual/design/architecture.md
```diff
@@ -0,0 +1,108 @@
...
+
+This block hold the following components:
```
paddatrapper, line 95: Holds or contains

docs/developer-manual/design/architecture.md
```diff
@@ -0,0 +1,108 @@
...
+
+- a `Playout` app that communicate with the `Play the schedule` block to gather the schedule,
```
paddatrapper, line 97: Communicates

docs/developer-manual/design/architecture.md
```diff
@@ -0,0 +1,108 @@
...
+- a Liquisoap app that play and mixes the scheduled items, and dispatch them to the delivery services,
+- a Icecast server that delivers a legacy audio stream to the public,
```
paddatrapper, line 99: Plays

paddatrapper: An Icecast

docs/developer-manual/design/architecture.md
```diff
@@ -0,0 +1,108 @@
...
+- a HLS stream that delivers a modern audio stream to the public.
+
+### One setup per radio station
+
+LibreTime is not meant to be used in a multi-tenant architecture, and an entire LibreTime installation should be dedicated to a single radio station. Previous SAAS or multi-tenant features from Airtime should be deprecated or removed.
+
+### Separation of concerns
+
+The `Create the schedule` block must only prepare a schedule, and the `Play the schedule` must only play that schedule. A strong separation of concerns is required between the 2 blocks to allow the `Play the schedule` block to meet it's uptime requirements while not depending on the `Create the schedule` in case of a failure. Development will be simplified if both blocks share a single properly defined protocol.
```
paddatrapper, line 108: its (without the apostrophe)

paddatrapper:
From a technical perspective, this looks good to me. I can't think of any other components we may need or things we should do differently. I like the separation of concerns and I think the structure allows room for growth of either side independently

jooola:
This shouldn't make the design definitive, I think we will be able to add/changes details as we go and encounter problems. Some of our discussion previous/future could even become chapters of this document, the configuration management for example.

On a wider perspective, we could even embrace ideas like the 12 factor app guidelines, but that would be for future issue/PRs.

But this answer most of the concerns I had with regards to big changes I proposed in the past and gives us a good baseline.

paddatrapper:
Do you want to restructure the commit history or shall I squash it?

PR #1737, opened 2022-04-07
ci: add website preview deployement

jooola:


jooola:
Sorry for the spam...

To make this work I need the following:

- A deploy key to `libetime/libretime.github.io`. Basically a ssh key (`ssh-keygen -t ed25519 -f ./id_ed25519`) with the public key stored in the `libetime/libretime.github.io` deploy keys with write access, and the private key in the secrets (as `PREVIEW_DEPLOY_KEY`) of a newly created `website-preview` environment.
- To make the `libetime/libretime.github.io` repository public again an configure Github page to be published to `libretime.github.io`.

jooola:
I have something working on my end https://github.com/jooola/libretime/pull/77

paddatrapper:
You probably don't need a deploy key, but the push over will need a personal access token with public repo access. I wonder if setting up a libretime service account is a good idea for this

paddatrapper:
Added the deploy key, updated the repo and made it public

paddatrapper:
Can you check the CI failure - https://github.com/libretime/libretime/runs/5881357686?check_suite_focus=true?

It seems to be related to access to this repo to post comments rather than the push access to libretime.github.io

jooola:
Hmm, ok there is a lot of security implication in this new workflow, I'll try to put some safeguards so we don't naively make our repo open to attacks.

jooola:
Deployment environment are often used for production or staging environment, not sure if preview environment are the right way to go.

We could have a `website-preview` environment to require a review before deploying the PR. This should prevent malicious action.

jooola:
Or we could simply wait for a manually triggered workflow using `workflow_dispatch`. This sound like a simpler idea.

paddatrapper:
Probably having a manual action makes more sense, but I'm not sure if it is possible to trigger for a branch on a different fork... Something we'd need to test

Environments are tricky because the assumption is that there is a single deployment and that gets rolled over each time. They don't handle multiple concurrent deployments well

jooola:
Closing to reduce spamming. I'll reopen once I have something better to merge.

PR #1738, opened 2022-04-07
docs: create 3.0.0-alpha.12 versionned documentation

jooola:


paddatrapper:
Is there no way of doing this based on git history, i.e. a git commit hash reference? Duplicating all the docs every release doesn't seem sustainable

jooola:
This is how docusaurus handle versioned documentation, I agree it's not really clean, but I guess that they thought this through when building the software. If we really don't want to handle it this way, we will have to rely on a branch and some extra workflows. I'll have a look at a possible different solution.

paddatrapper:
Ok. Then I think cleaning this up is something we can kick down the road

jooola:
What could happen is, we have the PR's review website on libretime.github.io/pr-{pr-number}, we could have the latest versioned docs on libretime.org and the unreleased docs on libretime.github.io/next/ ?

paddatrapper:
A lot of people run main at the moment, so I think let's have the two on the website for now

PR #1740, opened 2022-04-08
ci: generate schema and push to api-client repo

paddatrapper:
Allows automatic generation of api clients based on the Django schema. They still need testing, but the workflow is now up and running to generate them.

### DIFF
--- a/.github/workflows/api-schema.yml
+++ b/.github/workflows/api-schema.yml
@@ -1,0 +1,32 @@
+name: Push API schema
+
+on:
+  push:
+    branches: [main]
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  generate:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v3
+      - name: Clone api-client repo
+        uses: actions/checkout@v3
+        with:
+          repository: "${{ github.repository_owner }}/api-client"
+          path: api-client-repo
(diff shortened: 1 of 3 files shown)


.github/workflows/api-schema.yml
```diff
@@ -0,0 +1,34 @@
...
+        uses: actions/checkout@v3
+        with:
+          repository: "${{ github.repository_owner }}/api-client"
+          path: api-client-repo
+          ssh-key: "${{ secrets.API_CLIENT_DEPLOY_KEY }}"
+      - name: Generate schema
+        run: |
+          make install
+          source .venv/bin/activate
+          DJANGO_SETTINGS_MODULE=libretime_api.settings.testing libretime-api spectacular --file schema.yml
```
jooola, line 27: What about moving this to a make file target?

.github/workflows/api-schema.yml
```diff
@@ -0,0 +1,34 @@
...
+          cp schema.yml ../api-client-repo/schema.yml
+        working-directory: api
+      - name: Push schema changes to API clients
+        uses: stefanzweifel/git-auto-commit-action@v4
+        with:
+          commit_message: Update schema
```
jooola, line 33: Could we include the commit sha `${{ github.sha }}` in the commit message ? It should help for browsing the changes involved in the schema update.

paddatrapper: Can do. It won't be 100% accurate though, as if there is a rebase where commit 1 changes the schema, but is part of 2 or more commits, only the final one will trigger the CI and be recorded.

PR #1741, opened 2022-04-08
ci: update api-schema commit message

jooola:
Fixing the commit message and use conventional commits for the commit message.


jooola:
Commit references will be like libretime/api-client@1dafa7d83270fc7aee22bba91327d2d65d7c4c81

PR #1747, opened 2022-04-09
ci: test on ubuntu jammy

jooola:


analyzer/tests/pipeline/analyze_playability_test.py
```diff
@@ -33,8 +33,8 @@ def test_analyze_playability_invalid_filepath():
 
 
 def test_analyze_playability_invalid_wma():
-    # Liquisoap does not fail with wma files on buster, bullseye, focal
-    if distro.codename() in ("buster", "bullseye", "focal"):
+    # Liquisoap does not fail with wma files on buster, bullseye, focal, jammy
+    if distro.codename() in ("buster", "bullseye", "focal", "jammy"):
```
paddatrapper, line 37: Maybe it would be better to check for the distro releases it does fail on? So just Ubuntu 18.04

jooola: I expect to remove this check anyway, as we cannot rely on it anymore.

jooola: In addition I prefer to have to test fails and add an extra line here, instead of not testing it because one missed to add the extra line/bytes to test on a specific distro.

paddatrapper: We do need to remove it once 18.04 is EOL as the newer liquidsoap versions work with wma files

PR #1752, opened 2022-04-10
ci: add website preview

jooola:
I removed the fancy comments with links to the preview. This makes the workflow simple and secure.

:warning: Please do not squash


jooola:
If we really want to add a comment with a link the preview page, I can add an extra workflow that starts once the preview workflow is complete. If we think it isn't really necessary, this should be ready to be merged.

PR #1743, opened 2022-04-09
ci: build test images for ubuntu jammy

jooola:
Some `packages.ini` files were not updated to support jammy because I could not test the outcome. (mainly referring to the php stack)

PR #1755, opened 2022-04-11
fix(installer): install software-properties-common when required

paddatrapper:
Addresses point 1 of #1743


jooola:
@paddatrapper Added a small modification. Feel free to merge.

PR #1759, opened 2022-04-12
ci: setup command dispatcher

jooola:
The new command dispatch workflow require a service PAT to trigger the underlying workflow.

I can't decide if we want to call this a BOT_PAT or SERVICE_PAT.

### DIFF
--- a/.github/workflows/command.yml
+++ b/.github/workflows/command.yml
@@ -1,0 +1,21 @@
+name: Command
+
+on:
+  issue_comment:
+    types: [created]
+
+jobs:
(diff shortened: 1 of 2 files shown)


paddatrapper:
WEBSITE_PREVIEW_TOKEN would be my suggestion - easier to understand when looking at the list of variables

jooola:
I think I would like to not mix the command dispatch secret with the preview specific secrests

paddatrapper:
Ok. Then DISPATCH_SERVICE_TOKEN

PR #1764, opened 2022-04-13
ci: check if api schema is outdated for all commits

jooola:
The commits are all pushed separately to trigger a workflow on the remote repository.

### DIFF
--- a/.github/workflows/api-schema.yml
+++ b/.github/workflows/api-schema.yml
@@ -3,40 +3,104 @@ name: API schema
 on:
   push:
     branches: [main]
+    paths:
+      - .github/workflows/**
+      - api/**
+
+  pull_request:
+    branches: [main]
+    paths:
+      - .github/workflows/**
+      - api/**
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
-  cancel-in-progress: true
 
 jobs:
-  generate:
-    if: github.repository_owner == 'libretime'
-    name: Generate
-
+  check:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
       - uses: actions/setup-python@v4
         with:
           python-version: "3.x"
 
-      - name: Clone client repo
-        uses: actions/checkout@v3
+      - uses: actions/cache@v3
         with:
-          repository: "${{ github.repository_owner }}/client"
-          path: client-repo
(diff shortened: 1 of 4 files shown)


jooola:
This feels really fragile. If a commit in between fails to generate the schema file, it will probably break the entire workflow run.

I wonder if we can find a better way of handle the schema file update.

I tried with a pre-commit hook and storing the schema file in the repo but this is a heavy task for pre-commit.  But it might be the best solution.

paddatrapper:
You could add logic to skip a commit if the generation fails. But perhaps that is something that can be added later if we find that it is an issue

jooola:
Let's have this workflow also run in the pull request and if a schema generation fails for a single commit the workflow should fails. This should make it safer once it is merged into main.

jooola:
Ok this is ready to be merged. It should be working once it land on the main branch, hope it is at least.

Please do not squash.

The workflow will check if the schema is outdated on any commit off a PR.
And once pushed to the main branch, dispatch any commit that changed the schema file to the client repository.

.github/workflows/api-schema.yml
```diff
@@ -3,40 +3,101 @@ name: API schema
...
     steps:
       - uses: actions/checkout@v3
-      - uses: actions/setup-python@v4
         with:
-          python-version: "3.x"
+          fetch-depth: 0
 
-      - name: Clone client repo
-        uses: actions/checkout@v3
+      - uses: actions/setup-python@v3
```
paddatrapper, line 27: Please update this to v4 and specify the `python-version "3.x"`

jooola:
Weird seems like a dependency update introduced a regression.

paddatrapper:
I'm re-running the tests

paddatrapper:
Seems to be something new? The run 2hrs ago on #1886 succeeded - https://github.com/libretime/libretime/actions/runs/2515857685

jooola:
While working on this, the schema required to be updated from the one I generated yesterday or so. All the filtering args were removed from the schema for the schedule endpoint.

paddatrapper:
Hmm... OK. That is problematic because that is how playout determines the next set of media to play

jooola:
Found this https://github.com/jooola/libretime/runs/6947227245?check_suite_focus=true#step:5:373

As the test aren't failing on bionic, maybe the deps didn't get upgraded because of some python constraint.

I'll check if I can fix this.

jooola:
Yup, that was it. I'll open a PR.

PR #1767, opened 2022-04-14
chore: add bump python version tool

### LINKED ISSUE #864: Can't tell which version is actually installed.
I just did a fresh install two days ago, using GIT I noticed on the version tool tip is says I am using an old version and that 3.0.0-alpha.8/ is available. I went to to my install directory, ran “git pull” and it says it is up to date?

What am i missing here?
Selection_495

jooola:
This will bump the version in the setup.py files before creating a tarball, I think we could do this during release and create a version bump commit.

Fixes #864

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,8 +18,11 @@ shell-check:
 VERSION:
 	tools/version.sh
 
+bump-python-version: VERSION
+	tools/bump-python-version.sh
+
 .PHONY: tarball
-tarball: VERSION
+tarball: VERSION bump-python-version
 	$(MAKE) -C legacy build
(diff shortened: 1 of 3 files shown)


paddatrapper:
This tool is definitely needed, but I think the dev version of the apps should be 'latest' or 0.0.0 if it requires a semver. The release pipeline then sets the version of the archive, but doesn't commit it back. Makes things like managing the git tags, etc easier

paddatrapper:
This tool is definitely needed, but I think the dev version of the apps should be 'latest' or 0.0.0 if it requires a semver. The release pipeline then sets the version of the archive, but doesn't commit it back. Makes things like managing the git tags, etc easier

PR #1770, opened 2022-04-17
feat: replace exploded base_* with public_url

jooola:
BREAKING CHANGE: The `general.base_*`, `general.protocol`
and `general.force_ssl` configuration fields have been
replaced with a single `general.public_url` field. Be sure
to use a valid url with the new configuration field.

PR #1774, opened 2022-04-18
feat: move allowed cors setting to configuration file

jooola:
Depends on #1770


jooola:
I want to move the allowed_hosts (cors) to the configuration file, but the ini config format doesn't allow lists (php and python ini lists format are really different...). I need this change to be able to remove the web installer and merge the new installer/new migration tool/new config file format.

I could use a string and split the string at each `,` though.

It starts getting annoying to make perfect PR, while I know the next PR will prune the files I just updated. I have enough of the web installer and the config file, @paddatrapper is that ok if I put many changes in a single PR and we test this one before merging all the individual PRs ? I want to change the config file format, the migrations tool, the installer but they all have some kind of chicken/egg problem and it always boils down to the web installer...

paddatrapper:
Yeah that sounds reasonable

jooola:
Closing in favor of #1821

PR #1777, opened 2022-04-20
chore: unzip package is used by composer

jooola:


paddatrapper:
Is removing zip intentional?

jooola:
Yes it is intentional. I am running without this package (at least I didn't requested any installation of it) on production, maybe I didn't use any feature that required it, but I doubt we really require it.

https://github.com/libretime/ansible-role-libretime/blob/96cbdb01da2f0e74813745055e797f3de12f4bdd/tasks/legacy.yml#L21-L24

PR #1781, opened 2022-04-20
chore(api): move uswgi config in api/install

jooola:


paddatrapper:
Nit (not crucial): the message should be to api/install or into api/install

Issue #1719, opened 2022-03-31
Replace uwsgi with a ASGI replacement


jooola:
Moving towards the modern ASGI protocol, to support async and websockets.

https://asgi.readthedocs.io/en/latest/introduction.html#what-s-wrong-with-wsgi

I propose to use uvicorn + gunicorn https://docs.djangoproject.com/en/4.0/howto/deployment/asgi/uvicorn/

paddatrapper:
Sounds good. They are both in the Debian repos and have been for a while. Looks like packaging is consistent and continuing. With that happening in Debian, it should trickle into Ubuntu nicely.

jooola:
It is recommended to use ninx next to gunicorn +uvicorn, so we might want to wait for #1186 

That said, I am already experimenting in the ansible role with nginx and I should be able to add gunicorn as well.

PR #1785, opened 2022-04-21
feat(api): replace uwsgi with gunicorn

jooola:
Related to #1719

BREAKING: The uwsgi packages and configuration files (`/etc/airtime/libretime-api.ini`) should be removed from the system.


api/packages.ini
```diff
@@ -2,10 +2,8 @@
 [python]
 python3 = buster, bullseye, bionic, focal
 python3-pip = buster, bullseye, bionic, focal
-
-[django]
-uwsgi = buster, bullseye, bionic, focal
-uwsgi-plugin-python3 = buster, bullseye, bionic, focal
+gunicorn = buster, bullseye, bionic, focal
+python3-gunicorn = buster, bullseye, bionic, focal
```
jooola, line 6: I wonder if we prefer to install/pin a version from pip for now and let the package maintainer install those packages.

paddatrapper: I'd rather use distro versions of packages where possible

PR #1786, opened 2022-04-21
ci: add codeql analysis

jooola:


paddatrapper:
CodeQL is not open source and has a very limiting set of terms and conditions...

jooola:
I thought it was, haven't done any research in that regard.

PR #1803, opened 2022-04-25
chore: allow deps scope for commit messages

jooola:


jooola:
Do we need a deps-dev scope as well ?

paddatrapper:
I don't think we need that as well, they are still deps

Issue #1787, opened 2022-04-21
Change password hashing algorithm


jooola:
**Is your feature request related to a problem? Please describe.**

We urgently need to change to password hashing algorithm to something secure, md5 is not secure.

**Describe the solution you'd like**

We need something supported by both python (django) and php.

I think bcrypt or argon* is the way to go until we drop the php code. Then we can let django choose the best password hash algorithm.

Django handles password very well, the required changes are mostly php related to comply with the django stored password format.

https://docs.djangoproject.com/fr/3.2/topics/auth/passwords/#how-django-stores-passwords

We should update the password during a future user login, check if we use the old format of the password, and update accordingly if the password is valid.

jooola:
https://github.com/rumeau/zf1-authadapter-dbtablebcrypt/blob/master/Jr/Auth/Adapter/DbTableBcrypt.php
https://github.com/zf1s/zf1/blob/master/packages/zend-auth/library/Zend/Auth/Adapter/DbTable.php

jooola:
In addition, once we have #1644 we could simply wrap all the md5 passwords inside the new hashing algorithm so we don't have to wait for all the user to log in to update there passwords. 

https://docs.djangoproject.com/fr/3.2/topics/auth/passwords/#password-upgrading-without-requiring-a-login

PR #1804, opened 2022-04-25
ci: reduce usage

### LINKED ISSUE #1798: Reduce CI/CD usage
We should reduce the CI/CD usage by running some workflow only for some changed files.

This is not related to any pricing limitation, but simply make the workflows more reactive as we do have limitations on the number of concurrent workflows/jobs, and reduce some CPU usage (its free but still eats energy).

I propose to split the test workflow into multiple workflow per component (analyzer/api/api-client/playout/legacy/...) so we can trigger some test only if some files changed.

Having workflow per component would move the CI/CD in a monorepo like structure, which seems to fit our needs.

We can use reusable workflow to reduce duplication. I already have a working example here https://github.com/jooola/libretime/pull/86

This does mean a total rebuild of our test workflows.

jooola:
Fixes #1798

Please squash!


jooola:
The required workflow names changed.

.github/workflows/legacy.yml
```diff
@@ -0,0 +1,77 @@
+name: Legacy
+
+on:
+  push:
+    branches: [main]
+    paths:
+      - .github/workflows/**
+      - api/**
```
paddatrapper, line 8: Why run on api changes? Currently there are no tests that would verify things breaking in legacy if the API changed

jooola: I expect both to be tightly coupled together during the transition to django. We already add more coupling with #1644 or #1787 this was more a way to make sure we don't break legacy when changing something on the api side. 

I can remove this path though.

paddatrapper: Not necessary with #1644. With that, we will need to test both

paddatrapper:
I've actually dropped the required tests - in general all of our tests are required for merging. We need to check if that does what is expected for auto-merging though

PR #1819, opened 2022-04-28
feat: move storage path to config file

jooola:
Move the storage path from the database to the configuration file.


jooola:
Closing in favor of #1821

PR #1840, opened 2022-05-05
feat(worker): don't run with a dedicated user

jooola:
BREAKING: The worker service no longer uses a dedicated `celery` user to run. The old `celery` user can be removed from the system.


jooola:
The check command doesn't work, I think we should think to a way to make the check either platform specific (docker/systemd) or way more generic (using http health check or something else).

jooola:
I changed to check to use systemctl, following #1848 

This is ready to be reviewed/merged.

PR #1841, opened 2022-05-05
fix(legacy): check libretime-api service using systemctl

jooola:


jooola:
Check if the service is running using systemctl, according to #1848

PR #1821, opened 2022-04-29
feat: rebuild install process

jooola:
- Fixes #1852
- Fixes #1849
- Fixes #1377
- Fixes #1372
- Fixes #1204
- Fixes #1026
- Fixes #420
- Fixes #1642
- Fixes #889
- Fixes #1378
- Fixes #1300

PR #1847, opened 2022-05-08
feat: rename default stream mountpoint

jooola:
While I would like to change the mount point to something more generic, I feel this change will be a pain for a lot of people who relied on the default stream.

BREAKING: The default stream mount point changed from `airtime_128` to `main`. Be sure to updates your clients accordingly.


paddatrapper:
This maybe makes sense on new installs. So don't change existing upgrades. Is live.ogg a default for some clients? What is the motivation behind the change?

jooola:
> This maybe makes sense on new installs. So don't change existing upgrades. Is live.ogg a default for some clients? What is the motivation behind the change?

Mostly is to get finally rid of the airtime name in the mount point. and have something more generic.

I addition I though it would be good to propose more that a single mount point by default, and have some kind of hierarchy between them. I personally like the following mount point schema, with `low` being for low quality streams.
- `main.ogg`
- `main-low.ogg`
- `main.mp3`
- `main-low.mp3`

paddatrapper:
As far as I know, the default is in the database. The change between default and custom for stream 1 should just be a UI thing and configuration on the toggle. Other than that, the DB details are used.

Having a high and low quality stream is a nice idea, though maybe we want to look at HLS and replace Icecast completely?

For naming, do we even need the file extension? Most players don't determine the format from the extension (or at least shouldn't), so could do live and live-low or main and main-low. That would be consistent between formats and make documentation simpler. The actual name (live/main/something) I would need to think about a bit more

jooola:
> As far as I know, the default is in the database. The change between default and custom for stream 1 should just be a UI thing and configuration on the toggle. Other than that, the DB details are used.

This is something I had trouble understanding. I think the naming doesn't help.

It the default stream is hardcoded here https://github.com/libretime/libretime/blob/6289b5be6f460beb8bd3bc072c3e2e3f36462d65/legacy/application/models/StreamSetting.php#L281-L294

and overrides the s1 stream when not using the custom streams.

So I think we could move the default stream mountpoint to the preference table just like the `default_icecast_password`.

> Having a high and low quality stream is a nice idea, though maybe we want to look at HLS and replace Icecast completely?

I really want to have HLS, but I think ditching Icecast is a bad idea. Many diffusion (FM/DAB) provider expect to have an icecast endpoint to listen to.

> For naming, do we even need the file extension? Most players don't determine the format from the extension (or at least shouldn't), so could do live and live-low or main and main-low. That would be consistent between formats and make documentation simpler. The actual name (live/main/something) I would need to think about a bit more

I personally include the extension/format, so I can make the difference and swap the streams depending on the browser capabilities (mostly older browser I agree...) or client capabilities.

jooola:
In the long term I would actually ditch the default stream and directly use the configured s1 stream

paddatrapper:
I agree that is the best way to go. Though maintaining a list of the default settings is helpful if a station wants to be able to revert back to the default streaming output (e.g. after testing or for testing)

jooola:
I will only be able to implement a decent migration once we merge the Django based migrations.

We depend on #1821

jooola:
For now the default stream mount point is not editable from the UI, I wonder if we should make it editable. If so I'll have to understand how the Zend forms are working.

jooola:
> I agree that is the best way to go. Though maintaining a list of the default settings is helpful if a station wants to be able to revert back to the default streaming output (e.g. after testing or for testing)

Maybe we should rather provide a "reset" button for people willing to restore the default settings.

jooola:
This is ready to be merged.

PR #1868, opened 2022-06-09
fix: use constrained foreign key for files track_type

### LINKED ISSUE #1366: cc_files track_types is not a foreign key, but a unconstrained varchar
**Describe the bug**

In the db, the cc_file table use track type to do some extra sorting. But the column is an unconstrained varchar and should be a foreign key to the track_types table.

This doesn't allow easy updates on existing files with track types.

jooola:
Fixes #1366


jooola:
@paddatrapper I think we can ship this one in alpha.13 too. I think it needs proper testing, but from what I've tested upload/smartblocks are working.

paddatrapper:
I've wanted to test this, but my vagrant setup is currently broken thanks to https://bugs.debian.net/1014380. Happy to merge

PR #1876, opened 2022-06-14
docs: add french user-manual translations

jooola:


jooola:
The screenshot aren't translated, and I don't know how to best handle them. Should we change the paths pointing to them, or should we create symlinks to the existing files ?

I will try changing the paths, hope docusaurus handle well the multiple path computation.

jooola:
It seems we cannot have a partially translated documentation.

Seems like if we only translate the user-manual, some link from the admin-manual to the user-manual will be broken.

image


I already use a really long relative path to get the untranslated screenshots, which seem a bit dirty.
image

jooola:
In addition, keeping track of the changes between multiple languages seem also to be a complex topic.

We might want to extract the translated documentation to it's own repository, and use a bot to raise issues if we detect changes on the default documentation.

jooola:
I think this should go in a dedicated repository, with it's own name spaced URL e.g. `/fr/`.

PR #1877, opened 2022-06-15
feat: add bulk upload cmd

jooola:


jooola:
This bulk upload command is showing some of the weaknesses of the upload pipeline and the analyzer when under heavy load.

Re-uploading a file that hasn't been processed by the analyzer yet, will fail, because the second file goes exactly to the same path than the first  one and if moved by the analyzer during the first file analysis.

I am unsure if there is some throttling for the analyzer ingestion. Moving to celery would help this.

Probably some other things.

legacy/application/models/airtime/CcFiles.php
```diff
@@ -147,6 +147,9 @@ private static function createAndImport($fileArray, $filePath, $originalFilename
 
             self::validateFileArray($fileArray);
 
+            // Early md5dum processing
+            $md5 = md5_file($filePath);
+
```
jooola, line 152: I am unsure if this is smart, when uploading a large file it might timeout, but I don't how large a file has to be to timeout.

api/libretime_api/storage/management/commands/bulk_import.py
```diff
@@ -0,0 +1,157 @@
...
+
+logger = logging.getLogger(__name__)
+
+DEFAULT_ALLOWED_EXTENSIONS = [
+    ".flac",
+    ".m4a",
+    ".mp3",
+    ".ogg",
+    ".opus",
+]
```
jooola, line 18: Maybe we should extend this list to something more realistic. Open for suggestions

paddatrapper: wav at least

jooola: Ok, but we should highly recommend people not to use wav files as they don't store metadata.

jooola: Maybe this can be moved to the configuration file actually, this will be used by the future django upload endpoint to filter unwanted files extension ?

legacy/application/models/airtime/CcFiles.php
paddatrapper: MD5 is O(n), so on modern processors it is probably a fairly large file. I think the benefit of having it here is higher than the potential timeout. If stations encounter timeouts, they (or we) can always increase the timeout or re-look at this

jooola:
I'll write a test and should be good to go.

paddatrapper:
Is this ready for review?

jooola:
Feel free to review, but haven't found a clean way to test this code yet. This requires a lot of mocking (DB + Legacy api + local filesystem with fixtures to upload).

jooola:
Ok, I wrote some tests for the importer class. I think it's good enough.

PR #1883, opened 2022-06-16
docs: remove packages based install

### LINKED ISSUE #1879: Remove documentation about package install ?
The libretime debian package repository is out of date with the current libretime version.

Should we remove the documentation about it until we upgrade the package repository ?

https://libretime.org/docs/admin-manual/setup/install/#using-packages

@paddatrapper

jooola:
Fixes #1879


paddatrapper:
Maybe remove from the alpha.12 docs too? There is no packaged version for that version any way

PR #1886, opened 2022-06-17
feat: add internal url config

### LINKED ISSUE #515: Feature Request: Use a different "base_url" for API services than the web front-end
So, I'm running Libretime in a multi-container docker setup that I've built ready for production use (see here for the repo: https://github.com/ned-kelly/docker-multicontainer-libretime), and ideally I'd like to request that there be two base_url variables added to the airtime.conf file:

 - One for the actual Web UI (So that you can specify an external public facing endpoint for the IFrames that are loaded)

 - The other would be the endpoint that API's can hit - This way it can be another local docker containers hostname (joined in a bridged network), or just "localhost" etc...

Hope this makes sense and isn't too much to ask.

jooola:
Fixes #515

### DIFF
--- a/api-client/libretime_api_client/version1.py
+++ b/api-client/libretime_api_client/version1.py
@@ -129,7 +129,7 @@ class AirtimeApiClient:
         self.logger = logger or logging
 
         config = Config(filepath=config_path)
-        self.base_url = config.general.public_url
+        self.base_url = config.general.get_internal_url()
         self.api_key = config.general.api_key
 
         self.services = RequestProvider(
--- a/api-client/libretime_api_client/version2.py
+++ b/api-client/libretime_api_client/version2.py
@@ -32,7 +32,7 @@ class AirtimeApiClient:
         self.logger = logger or logging
 
         config = Config(filepath=config_path)
-        self.base_url = config.general.public_url
+        self.base_url = config.general.get_internal_url()
         self.api_key = config.general.api_key
 
         self.services = RequestProvider(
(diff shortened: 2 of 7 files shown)


jooola:
This requires to remove the default stream settings as it rely on guessed values from the public URL and only use the "custom stream settings". Otherwise we would have to provide an extra settings for the default stream settings icecast host.

paddatrapper:
The streaming defaults are only an issue until we move streaming settings into the settings file - then the default can live in there. So I suggest we leave the default as is and document that it needs changing before streaming in a docker environment will work. i.e after creating the environment, log in and update the streaming configuration to use the icecast container

jooola:
> The streaming defaults are only an issue until we move streaming settings into the settings file - then the default can live in there. So I suggest we leave the default as is and document that it needs changing before streaming in a docker environment will work. i.e after creating the environment, log in and update the streaming configuration to use the icecast container

I'll assume the default stream is always on localhost, and as you suggested, docker user will have to use custom streams.

jooola:
I am tempted to remove the default stream url, this should make this PR simpler and maybe after that I can push some more to get this merged.

Breaking things to prepare for docker is a good idea before releasing beta.0.

paddatrapper:
Remove changing it or remove it as a setting?

jooola:
> Remove changing it or remove it as a setting?

Remove the entire SAAS default stream settings, from the UI and the DB, leaving only the current "custom stream" settings.

After that  we can separate the icecast connection config from the stream public URL. Having a dedicated host to configure icecast would allow us to keep the public url for the public facing url of the streams.

I am currently doing some dirty hack were the icecast host is a domain name pointing to my public IP, and I need a output port redirect in my firewall on the libretime guest to hit the icecast server.

paddatrapper:
Ah I see. Yes that makes sense

PR #1889, opened 2022-06-18
chore(api): pin django dependencies

### LINKED ISSUE #1888: Pin django dependencies
We should pin the dependencies in order to have an explicit upgrade mechanism with PR + testing + renovate.

jooola:
Fixes the failing api tests.
Fixes #1888


api/setup.py
```diff
@@ -32,11 +32,11 @@
...
-        "django~=3.0",
-        "djangorestframework",
-        "django-filter",
-        "drf-spectacular",
+        "coreapi~=2.3.3",
+        "django~=3.2.13",
+        "djangorestframework~=3.13.1",
+        "django-filter~=21.1.0",
+        "drf-spectacular~=0.22.1",
         "markdown",
```
jooola, line 40: I was wondering why we need the markdown package here ?

paddatrapper: I can't remember. Think it is a dependency somewhere?

paddatrapper:
I'm not sure that this sem-ver range is going to work in the long term. Especially if we want the Debian/Ubuntu packages to work - there we will need to support the versions within the various repos

api/setup.py
jooola: It isn't used in our code at least. We could probably remove it and add it back if we need markdown.

jooola:
What do you have in mind ? I don't want to be too open about the version range and I want to prevent any version bump, so renovate to pick it up.

jooola:
Opened a new PR, feel free to write or commit what you have in mind there.

PR #1900, opened 2022-06-21
chore(api): update preferences models

jooola:


jooola:
Oups, totaly forget to update the commands using the pref models

PR #1905, opened 2022-06-22
feat(api): rename storage models fields

jooola:


jooola:
I don't if we should fix anything ins this PR, but maybe take the issue into consideration https://github.com/libretime/libretime/issues/526

api/libretime_api/storage/models/file.py
```diff
@@ -28,7 +28,14 @@ class ImportStatus(models.IntegerChoices):
     is_hidden = models.BooleanField(db_column="hidden", blank=True, null=True)
 
     is_accessed = models.IntegerField(db_column="currentlyaccessing")
-    edited_by = models.ForeignKey(
+    is_scheduled = models.BooleanField(db_column="is_scheduled", blank=True, null=True)
+    is_part_of = models.BooleanField(db_column="is_playlist", blank=True, null=True)
```
paddatrapper, line 32: Maybe is_playlisted? Is part of implies that the value is the object it is a part of

api/libretime_api/storage/models/file.py
```diff
@@ -28,7 +28,14 @@ class ImportStatus(models.IntegerChoices):
...
+
+    uploaded_at = models.DateTimeField(db_column="utime", blank=True, null=True)
```
paddatrapper, line 34: maybe on instead of at?

api/libretime_api/storage/models/file.py
```diff
@@ -2,82 +2,122 @@
...
+    comment = models.TextField(db_column="comments", blank=True, null=True)
     composer = models.CharField(max_length=512, blank=True, null=True)
+    conductor = models.CharField(max_length=512, blank=True, null=True)
+    copyright = models.CharField(max_length=512, blank=True, null=True)
+    language = models.CharField(max_length=512, blank=True, null=True)
     encoder = models.CharField(max_length=64, blank=True, null=True)
-    checksum = models.CharField(max_length=256, blank=True, null=True)
-    lyrics = models.TextField(blank=True, null=True)
+    encoded_by = models.CharField(max_length=255, blank=True, null=True)
+    isrc_number = models.CharField(max_length=512, blank=True, null=True)
```
jooola, line 96: Many fields could be renamed to something simpler, when obvious, here for example `isrc_number` => `irsc`

api/libretime_api/storage/models/file.py
```diff
@@ -2,82 +2,122 @@
...
+    url = models.CharField(max_length=1024, blank=True, null=True)
+    mood = models.CharField(max_length=64, blank=True, null=True)
     orchestra = models.CharField(max_length=512, blank=True, null=True)
-    conductor = models.CharField(max_length=512, blank=True, null=True)
+    lyrics = models.TextField(blank=True, null=True)
     lyricist = models.CharField(max_length=512, blank=True, null=True)
     original_lyricist = models.CharField(max_length=512, blank=True, null=True)
-    radio_station_name = models.CharField(max_length=512, blank=True, null=True)
+
+    rating = models.CharField(max_length=8, blank=True, null=True)
```
jooola, line 105: Again, unsure where this is used and how it is filled. It was empty in the databases I queried.

api/libretime_api/storage/models/file.py
```diff
@@ -2,82 +2,122 @@
...
     buy_this_url = models.CharField(max_length=512, blank=True, null=True)
-    isrc_number = models.CharField(max_length=512, blank=True, null=True)
     catalog_number = models.CharField(max_length=512, blank=True, null=True)
-    original_artist = models.CharField(max_length=512, blank=True, null=True)
-    copyright = models.CharField(max_length=512, blank=True, null=True)
+    subject = models.CharField(max_length=512, blank=True, null=True)
+    contributor = models.CharField(max_length=512, blank=True, null=True)
+
+    radio_station_url = models.CharField(max_length=512, blank=True, null=True)
+    radio_station_name = models.CharField(max_length=512, blank=True, null=True)
```
jooola, line 116: Maybe used for podcasts ? Or is a SAAS related feature ?

api/libretime_api/storage/models/file.py
```diff
@@ -2,82 +2,122 @@
...
+
     report_datetime = models.CharField(max_length=32, blank=True, null=True)
     report_location = models.CharField(max_length=512, blank=True, null=True)
     report_organization = models.CharField(max_length=512, blank=True, null=True)
```
jooola, line 120: What is this used for ?

api/libretime_api/storage/models/track_type.py
```diff
@@ -3,9 +3,18 @@
 
 class TrackType(models.Model):
     code = models.CharField(max_length=16, unique=True)
```
jooola, line 5: Tempted to rename this as key

api/libretime_api/storage/models/file.py
jooola: It can also be a block, maybe `is_part_of_list` ? Maybe `is_playlisted` is good enough, hope the spell checker doesn't scream !

api/libretime_api/storage/models/file.py
jooola: Maybe, I try to follow a convention and I am used to `created_at`, `modified_at`. I'll check if I already started to use `_at` somewhere else.

api/libretime_api/storage/models/file.py
paddatrapper: `is_part_of_list` would work

api/libretime_api/storage/models/file.py
paddatrapper: I don't really mind either way, but I agree that keeping to a standard convention is important

api/libretime_api/storage/models/file.py
```diff
@@ -2,82 +2,122 @@
...
+        SUCCESS = 0, "Success"
+        PENDING = 1, "Pending"
+        FAILED = 2, "Failed"
+
+    import_status = models.IntegerField(
+        choices=ImportStatus.choices,
+        default=ImportStatus.PENDING,
+    )
+
+    file_path = models.TextField(db_column="filepath", blank=True, null=True)
```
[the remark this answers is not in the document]
paddatrapper: If that doesn't create issues with how it is read when legacy writes a record or vice versa, yes. Probably will have to be FilePathField

api/libretime_api/storage/models/file.py
```diff
@@ -2,82 +2,122 @@
...
-    utime = models.DateTimeField(blank=True, null=True)
-    lptime = models.DateTimeField(blank=True, null=True)
+
     md5 = models.CharField(max_length=32, blank=True, null=True)
-    track_title = models.CharField(max_length=512, blank=True, null=True)
-    artist_name = models.CharField(max_length=512, blank=True, null=True)
+    checksum = models.CharField(max_length=256, blank=True, null=True)  # Duplicate ?
+    cue_in = models.DurationField(db_column="cuein", blank=True, null=True)
+    cue_out = models.DurationField(db_column="cueout", blank=True, null=True)
+    silan_check = models.BooleanField(blank=True, null=True)  # Unused, to remove ?
```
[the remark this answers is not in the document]
paddatrapper: I agree

api/libretime_api/storage/models/file.py
```diff
@@ -2,82 +2,122 @@
...
+
+    # https://github.com/libretime/libretime/issues/1366
+    track_type = models.CharField(max_length=16, blank=True, null=True)
+
+    artwork = models.CharField(max_length=512, blank=True, null=True)
```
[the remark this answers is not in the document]
paddatrapper: Depends on how the legacy code uses this. Same considerations are required as for `file_path`

api/libretime_api/storage/models/file.py
paddatrapper: Probably more general metadata fields that mutagen provides if it is set in the ID3 tags?

jooola: Yup, we will have to let the analyzer fill this field in that case.

jooola:
The thing is that analyzer only passes some metadata back to the DB, using a mapping dict: https://github.com/libretime/libretime/blob/cbfd9e51fdf9e2fb806645a9665e78f1e2237d20/analyzer/libretime_analyzer/pipeline/analyze_metadata.py#L72-L97

In addition some fields in the analyzer are not in the database, or use a different name, for example the comment field or the website field. See https://github.com/libretime/libretime/issues/526#issuecomment-427600267

Though, some fields might be only used by LibreTime for podcasts or something else. So it is difficult to know the origin of a field.

Anyway, something is definitely not working as expected, we should have a mapping for ALL the fields in the DB.

jooola:
https://dev.funkwhale.audio/funkwhale/funkwhale/-/blob/develop/api/funkwhale_api/music/metadata.py

api/libretime_api/storage/models/file.py
jooola: https://stackoverflow.com/questions/15038726/ruby-on-rails-convention-for-naming-date-time-fields

paddatrapper: I'm happy with that

api/libretime_api/storage/models/file.py
```diff
@@ -2,82 +2,122 @@
...
+        PENDING = 1, "Pending"
+        FAILED = 2, "Failed"
+
+    import_status = models.IntegerField(
+        choices=ImportStatus.choices,
+        default=ImportStatus.PENDING,
+    )
+
+    file_path = models.TextField(db_column="filepath", blank=True, null=True)
+    file_type = models.CharField(db_column="ftype", max_length=128)
```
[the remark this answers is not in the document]
jooola: I removed the ftype field from the api entirely.

api/libretime_api/storage/models/file.py
```diff
@@ -2,82 +2,122 @@
...
+    last_modified_at = models.DateTimeField(db_column="mtime", blank=True, null=True)
+
+    edited_by = models.ForeignKey(  # Unused ?
         "core.User",
         models.DO_NOTHING,
         db_column="editedby",
         blank=True,
         null=True,
         related_name="edited_files",
     )
```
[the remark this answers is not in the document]
jooola: Let's leave it like this for now.

api/libretime_api/storage/models/file.py
```diff
@@ -2,82 +2,122 @@
...
-    mtime = models.DateTimeField(blank=True, null=True)
-    utime = models.DateTimeField(blank=True, null=True)
-    lptime = models.DateTimeField(blank=True, null=True)
+
     md5 = models.CharField(max_length=32, blank=True, null=True)
-    track_title = models.CharField(max_length=512, blank=True, null=True)
-    artist_name = models.CharField(max_length=512, blank=True, null=True)
+    checksum = models.CharField(max_length=256, blank=True, null=True)  # Duplicate ?
```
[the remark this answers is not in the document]
jooola: Removed the checksum field from the api.

api/libretime_api/storage/models/file.py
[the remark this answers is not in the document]
jooola: This will happen in a future PR.

jooola:
Oh completely forgot that playout have to handle both the legacy files fields and the new api files fields.

jooola:
We should actually not have any problems as the get_schedule function creates the events from the data collected on the API, so we have a compatibility layer there

paddatrapper:
LGTM, just needs the conflicts resolved

PR #1914, opened 2022-06-24
chore: move api client get_schedule to playout

jooola:


jooola:
This introduce testing for playout.

The api client v2 is really thin now :tada:.

Issue #1915, opened 2022-06-24
Setup coverage report and check for regressions


jooola:
I think we should setup a way to see the coverage report generated by pytest and coverage, on PR changing Python code. This should show if we make a bit regression in terms of testing coverage.

jooola:
We could use https://github.com/marketplace/codecov for example.

paddatrapper:
This is a very good idea

paddatrapper:
We can do it without a third-party app using the pytest coverage data file in a PR comment

PR #1916, opened 2022-06-24
ci: report pytest coverage in PR

paddatrapper:
Fixes: #1915


paddatrapper:
Need to find a way to get the permissions correct so that the CI can post a comment

jooola:
The tird party app is actually here to work around this problem.
I don't mind using coverall.

jooola:
If we do it without third party app, we might also need to save the coverage reports from the main branch to make a diff with the PR report.

Or we can try https://about.codecov.io/ or Code

jooola:
Woups, we forgot to setup flags to split each testing component https://docs.codecov.com/docs/flags

PR #1930, opened 2022-07-01
chore: clean schedule messages

jooola:


paddatrapper:
Are those the only values used?

jooola:
> Are those the only values used?

From the code I read yeah, I'll do some testing.

paddatrapper:
Makes sense - the artist and title are the only file metadata items that can be displayed as stream metadata

jooola:
Ok, just schedules some files, metadata show up in icecast. This is ready.

PR #1936, opened 2022-07-04
chore(legacy): update locales from code

jooola:


jooola:
Maybe this should run with pre-commit so we don't forget to update the strings.

jooola:
Well, nevermidn I'll remove the pre-commit hook in CI, too much complexity to handle multiple version of gettext that could output different results.

PR #1939, opened 2022-07-05
docs: first blog post, v3 beta.0

jooola:
@paddatrapper Feel free to write or dump ideas in the blog post.

This should be merged once we release beta.0, I am already starting to write it now so we have time to dump all our thoughts and correct everything.

I think this will be a good way to get more testing and prepare people for a coming v3 release.

We will be able to explain what happen in the past months, what didn't happen and what should happen in the future.


jooola:
I'll finish writing this post some other day. As I want to add a small call to sustain Libretime

PR #1941, opened 2022-07-06
feat: use dedicated 'libretime' user with nginx + php-fpm

### LINKED ISSUE #1186: Nginx suport
I would like you to give it support for nginx

### LINKED ISSUE #1637: Dedicated user that run libretime
We should consider adding a dedicated user to run the multiple LibreTime services.

For now we use `www-data` to run all the services. I think `www-data` is a generic user that might have already too much permission and might be used by other programs.

`www-data` will still  be used to serve files with the web server and handle uploads, but I would go towards a php-fpm solution, with the php process running as `libretime` and `www-data` simply having read only permission where it is needed.

We also have a dedicated `celery` user, but I think we should remove this celery user and use this new `libretime` user.

Regarding the backend services such as playout or analyzer, they definitely don' require the `www-data` user.

Depending on web server stack structure, we should have:
```bash
# Running user: libretime
/etc/systemd/system/libretime-analyzer.service

# Running user: libretime
# what about the uwgsi package user?
/etc/systemd/system/libretime-api.service

# Running user: libretime
# what about the celery package user?
/etc/systemd/system/libretime-celery.service

# Running user: libretime
/etc/systemd/system/libretime-liquidsoap.service

# Running user: libretime
/etc/systemd/system/libretime-playout.service

# Owner: libretime
# once liquidsoap does not write here anymore, we could change the permission to root
/etc/airtime

# Owner: libretime
# add www-data to the libretime group, for read only permission
/etc/airtime/airtime.conf

# Owner: libretime
/var/log/libretime

# Owner: libretime
/var/lib/libretime

# Owner: libretime
# how does the webserver write in here ?
/srv/airtime
```

### LINKED ISSUE #1719: Replace uwsgi with a ASGI replacement

jooola:
- Fixes #1186
- Fixes #1637
- Fixes #1719

### DIFF
--- a/analyzer/README.md
+++ b/analyzer/README.md
@@ -23,16 +23,16 @@ rabbitmqctl set_permissions -p /airtime airtime .\* .\* .\*
 ## Usage
 
 This program must run as a user with permissions to write to your Airtime music library
-directory. For standard Airtime installations, run it as the www-data user:
+directory. For standard Airtime installations, run it as the libretime user:
 
 ```bash
-sudo -u www-data libretime-analyzer --daemon
+sudo -u libretime libretime-analyzer --daemon
 ```
 
 Or during development, add the --debug flag for more verbose output:
 
 ```bash
-sudo -u www-data libretime-analyzer --debug
+sudo -u libretime libretime-analyzer --debug
 ```
 
 To print usage instructions, run:
--- a/analyzer/install/systemd/libretime-analyzer.service
+++ b/analyzer/install/systemd/libretime-analyzer.service
@@ -8,8 +8,8 @@ Environment=LIBRETIME_CONFIG_FILEPATH=@@CONFIG_FILEPATH@@
 WorkingDirectory=@@WORKING_DIR@@/analyzer
 
 ExecStart=/usr/local/bin/libretime-analyzer
-User=libretime-analyzer
-Group=libretime-analyzer
+User=libretime
+Group=libretime
 Restart=always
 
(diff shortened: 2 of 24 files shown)


jooola:
@paddatrapper I am all for using unix socket between nginx and php-fpm or gunicorn, but I wanted to have your input if you see anything bad with it ?

paddatrapper:
It will work for general use, but it will need to change for Docker - ideally php-fpm would be in a different container, so connecting via http would make more sense there

jooola:
> It will work for general use, but it will need to change for Docker - ideally php-fpm would be in a different container, so connecting via http would make more sense there.

I had this in mind, so I thought we might want to use tcp sockets, but I feel tweaking the listen socket isn't a bit problem as we will have to edit it anyway to specify the containers names.

Unix socket it is ! (This is going faster than I expected to upstream those differences https://github.com/libretime/ansible-role-libretime/issues/15)

jooola:
Tested fresh install and upgrade from main to this branch, it is working as expected if we properly follow the upgrade procedure.

jooola:
Add the uvicorn asgi server. Tested again on bullseye !

PR #1945, opened 2022-07-07
fix(legacy): keep playlists/blocks/podcasts on user removal

### LINKED ISSUE #1944: Deleting user also remove his smartblocks and playlists
Deleting user also remove his smartblocks and playlist.
I don't know if is a bug, or desired function. But if it is it would be good to add warning about it to the dialog message. Tracks ownership was transferred to default admin account. I would expect same behavior for playlist and smartblocks.

**To Reproduce**

1. Delete user
2. All his smartblocks and playlists are deleted automatically

**LibreTime version:**
1924919e1

jooola:
I don't know if this is good enough. Reassign to admin seem dirty as well.

Ideally we should have dialog to check what to do with the content of the user to remove.

Fixes #1944


jooola:
Maybe we could have a "lost & found" user to reassign the content ?

PR #1946, opened 2022-07-07
style(legacy): format files

jooola:


jooola:
This should reduce the amount of styling commits I'll have to squeeze before actually code changes in my future PRs.

PR #1951, opened 2022-07-07
feat(legacy): update php config

jooola:


legacy/application/models/StreamSetting.php
```diff
@@ -283,7 +283,7 @@ public static function getDefaults($prefix)
         $config = Config::getConfig();
 
         return [
-            'host' => $config['baseUrl'],
+            'host' => 'localhost',
```
paddatrapper, line 286: Is this used to generate the listening url on the mini site or from the UI?

jooola: Hmm, I guess I'm going a bit too fast.

I'll have a look, but I think you are right.

jooola: Reverting back, if we change this I'll make a dedicated PR.

PR #1952, opened 2022-07-08
chore: clean default data.sql

jooola:


jooola:
No changes were made, I only sorted the lines so we group inserts by features instead of when it was added.

PR #1954, opened 2022-07-08
feat: allow 4 streams

jooola:


jooola:
This should be merged after #1953

jooola:
Too much risks for the no real benefit. I don't know if the 4th stream is fully working and if it was only disabled or some kind of error/mismatch. Anyway, let's stay safe and close this for now.

PR #1955, opened 2022-07-08
chore(playout): move liquidsoap package in a module

jooola:


jooola:
This simply moves the libretime_liquidsoap package in the libretime_playout.liquidsoap module to simplify the structure of the modules.

We now can properly mirror the tests structure of the package layout, like other python projects in the repo.

This makes it also more tempting to build library like function that can be reuse across the playout package.

PR #1956, opened 2022-07-11
chore: add changelog generator

jooola:
I am not totally happy with the workflow, we will have to fine tune this along the way.

This only generate a specific tag changelog and prepend it to the existing changelog. This allow us to fix spelling errors and keep the fixes by not regenerating the entire changelog. 

The tag will always be the latest tag in the current branch, unless passed as env var `tag=3.0.0-alpha.12 make changelog`.

The changelog does not contain breaking changes as they are in the upgrade procedure. The BREAKING CHANGE commit messages foot notes doesn't allow later editing, so we should keep using them but we should document upgrade procedure in the release notes.

~~This puts a hard dependency on git-chlog, don't know if we have some better alternatives. Maybe we want to embrace a fully featured tool such as https://github.com/commitizen-tools/commitizen, which could handle commit message formatting, changelog generation, version bump, and CI checks.~~


paddatrapper:
Should the CHANGELOG link to the release notes on the website?

jooola:
Yeah, good idea, I'll edit the template

jooola:
Done, I've generated the changelog for 3.0.0-beta.0 as well.

PR #1970, opened 2022-07-17
feat(installer): remove allow-restart flag

jooola:
Services management should be handled by the admin. You can use the following command `systemctl restart  libretime.target` after the install/upgrade script.

Start/Stop/Restart of the services should be handled by the admin, we don't want to mess with missing migrations (This is a problem for legacy though, because the files are swapped and take effect immediately).

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -80,14 +80,13 @@ Vagrant.configure('2') do |config|
     LIBRETIME_RABBITMQ_PASSWORD=libretime \
     bash install \
       --listen-port 8080 \
-      --allow-restart \
       --in-place \
       http://192.168.10.100:8080
 
     libretime-api migrate
     libretime-api set_icecast_passwords --from-icecast-config
 
-    systemctl start libretime.target
+    systemctl restart libretime.target
     SCRIPT
 
     config.vm.provision 'install', type: 'shell', inline: $script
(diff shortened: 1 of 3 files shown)


paddatrapper:
A production system that didn't use `--allow-restart` before would be broken until the restart thanks to legacy then right? With this change that would be all systems until the sysadmin restarts the service?

jooola:
> A production system that didn't use --allow-restart before would be broken until the restart thanks to legacy then right? With this change that would be all systems until the sysadmin restarts the service?

Not sure I fully understand your comment.

This flag only allowed to restart the LibreTime services during upgrade. And was mainly used to quickly update the vagrant setup without having to restart the service manually (this is now handle somewhere else).

The flag was introduced when I rewrote the installer (in `alpha.13`), I don't think anyone will heavily rely on it, maybe some will, but it will be a problem if we restart the services before some migrations are run.

jooola:
> A production system that didn't use --allow-restart before would be broken until the restart thanks to legacy then right? With this change that would be all systems until the sysadmin restarts the service?

I might have understood: Yes, but it depends if the migrations are already run or not. If they aren't run the legacy is ahead, if they are, the python apps are behind. Though, the upgrade documentation recommends stopping the python services before running the migrations, so our only problem is with legacy. But I think we can let it as it is for now and let legacy log few error.

Regarding the flag, it is set to false by default, so if someone relied on the worse that could happen is that the installer will print the usage next time he will upgrade/install Libretime

PR #1976, opened 2022-07-20
docs: add releases distributions support

### LINKED ISSUE #1966: Define support distributions matrix
We currently advertise that we require either Ubuntu LTS or Debian stable to install LibreTime.

In reality, we also try to support Bionic and Buster.

Now that v3 is nearly out, I would like to raise this distro version matrix discussion again. I think LibreTime should develop on the current stable distro only, and provide patch versions for oldstable distros.

The main reason I see LibreTime running on old stable distros is that the software was installed at the time old stable was stable, and upgrading requires maintenance work that might not be available. In the same idea, LibreTime is not be upgraded to a next "major" or even "minor" version to prevent downtime or introduce bugs in the radio station, which again requires maintenance work.

So supporting bionic and buster until there EOL for those specific use case might be sensible. Though, we have been breaking a lot of stuff between alpha's, and the reasoning above doesn't completely stand with regards to stability when we talk about the pre 3.0.0 versions.

So I propose to support the current stable distros for fresh installs, and only support old stable distros with patches upgrades.

Here is a idea of what I could see:

|              | Ubuntu 18.04 | Debian 10  | Ubuntu 20.04 | Debian 11  | Ubuntu 22.04 |
| ------------ | ------------ | ---------- | ------------ | ---------- | ------------ |
| Release date | 2018-04-26   | 2019-07-06 | 2020-04-23   | 2021-08-14 | 2022-04-21   |
| End of life  | 2023-04      | 2024-06    | 2025-04      |    2026-06        | 2027-04      |
| Versions     |              |            |              |            |
| 3.0.x        | supported    | supported  | supported    | supported  | WIP          |
| 3.1.x        |              |            | supported    | supported  | WIP          |

This should allow use to reduce the version matrix on some parts of the project, such as Liquidsoap ( I really want to drop liquidsoap < 1.4.x)

This kinda mimics a distro packaging behavior, where the software is maintained with security/bug fixes only for the time the distro is not EOL.

jooola:
Fixes #1966

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,23 +37,6 @@ provided here.
 The full text of the contract is licensed under the GPL and available at
 the above link courtesy of the [ZeroMQ community](https://zeromq.org/).
 
-## Releasing
-
-The LibreTime maintainers release new versions of LibreTime at regular
-intervals. We follow the [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
-standards.
-
-In a nutshell, given a version number MAJOR.MINOR.PATCH we increment the:
-
-1. MAJOR version when we make incompatible API changes,
-2. MINOR version when we add functionality in a backwards-compatible manner, and
-3. PATCH version when we make backwards-compatible bug fixes.
-
-Given the history of the project we will be starting with version 3.0.0.
-
-Stay tuned for early alphas and lots of betas as we get development
-up and running.
-
 ## Support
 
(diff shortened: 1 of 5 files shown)


paddatrapper:
Maybe we should refer to `supported distribution releases` because the distros we support aren't changing (Ubuntu and Debian)?

PR #1979, opened 2022-07-22
chore(api-client): restructure modules

jooola:


paddatrapper:
Why are you using a module for v1, but a file for v2?

jooola:
I expect to move more files to the v1 folder, while v2 will be self contained in the future changes I'll propose.

jooola:
Moved the file out of the folder, I'll do the changes in the future PR, not to mix everything.

PR #1980, opened 2022-07-22
feat(playout): use single clients instance

jooola:
- Use legacy_client across playout code to make the difference between the old and new clients.
- Setup clients during initialization, and pass clients down to the different apps.


paddatrapper:
@jooola please don't merge your own commits - the C4 has guidance on the topic

jooola:
Yup, sorry trying to prevent massive PRs.

paddatrapper:
If they are urgent, ping me and I'll try review as quickly as possible

PR #1984, opened 2022-07-22
feat(api): don't use hyperlinked serializers

jooola:


paddatrapper:
As far as I remember three was a reason I went with Hyperlinked serializers, but I can't remember it off hand. I'll have to see if I can find it

jooola:
> As far as I remember three was a reason I went with Hyperlinked serializers, but I can't remember it off hand. I'll have to see if I can find it

Sure, we haven't been using the urls in the code at all, and it is bothering me more than it benefits. Hence the change.

We can go the other way around and keep the hyperlinked models while also adding all the missing IDs by hand.

paddatrapper:
No, I think doing IDs instead of hyperlinking is the better solution. I'll see if I can find if there is a technical reason to keep the hyperlinks tomorrow

paddatrapper:
I can't find why I went with hyperlinked serializers. So happy to merge this

Issue #1537, opened 2022-01-25
Fix resources hardcoded in http://


valerio-bozzolan:
**Describe the bug**
I've seen some LibreTime resources hardcoded with `http://`. This is (at least for my installation) a protocol downgrade problem.

Some of these resources:

* https://github.com/LibreTime/libretime/blob/main/legacy/application/models/Preference.php#L1143
* https://github.com/LibreTime/libretime/blob/main/legacy/application/models/Preference.php#L1128
* https://github.com/LibreTime/libretime/blob/main/legacy/application/controllers/ApiController.php#L679
* ....

**Expected behavior**

Do not hardcode `http://` and adopt instead something like this:

* https://stackoverflow.com/a/14270161/3451846

As suggested, an extra option like `public_url` would be amazing to determine the schema/host/port/base URL.

---

Thank you for everyone reporting this kind of resources here. I will update the list.

jooola:
The 2 lines in Preference.php are tricky. It would mean the liquidsoap/icecast input harbor accept secure connection (as it should) but this isn't a small change as the legacy code didn't plan for this in the past.

The third line seem easy to fix. I think the provided solution on stackoverflow is the right way to go, but that would mean we rely on a possible bad configured reverse proxy somewhere. This needs documentation.

An extra option would be to use a `public_url` field in the config to determine the schema/host/base url.

An extra note is that our apache configuration is missing a ServerName, I think we should add this configuration.

valerio-bozzolan:
I think having the StackOverflow solution as default, together with a `public_url` as override configuration, is an amazing idea.

paddatrapper:
Liquidsoap SSL is tracked in #1032

Issue #513, opened 2018-09-06
no ssl available for output stream


drh8ball:
Hi at all,
I want to integrate the player widget in my site, but my browser shows me mixed content because of the stream from http. I have a working setup for icecast with ssl, but I am not able to set right address. I can set port, but not protocoll.
![bildschirmfoto_2018-09-06_16-52-10](https://user-images.githubusercontent.com/43039898/45165841-f0cef900-b1f5-11e8-8383-11a25763ef04.png)
Where can I change this?
Thank you.

Robbt:
checkout the files in airtime_mvc/public/widgets - I'm not sure if that is where it actually is coming from but there is a js file.

Like much of this project unfortunately whomever wrote the code in the first place is no longer around and a lot of the knowledge we have is dependent upon us digging through the codebase to try to find answers.

Looking at the path in the url embeddablewidgets I did `grep -ir 'embeddablewidgets' *`
in the source directory and I found application/controllers/EmbeddablewidgetsController.php & airtime_mvc/application/views/scripts/embeddablewidgets/player.phtml 

I'd check those out for a possible answer to your problem. I'd do more digging for you myself but I'm trying to get as much done while my daughter is taking a nap. Let us know if you find a good solution/hack. It is something I need to fix as well.

drh8ball:
My hack is very very dirty, but it works.
I will search for a clean way, but I am no php coder.
In /usr/share/airtime/php/airtime_mvc/application/models/StreamSetting.php on line 81 i did this:
$url = "http://$host:$port/$mount"; --> $url = "https://$host:8443/$mount";

I think about making a new checkbox "alternate url" under settings output stream stream X.
If it is checked, there will be a textbox for the string, this must be written to database, from there we can use it directly in my file, without this dirty hack.

JohnnyC1951:
I suspect you have set the SSL flags in Icecast and assumed it's working?
You can use Icecast 2.4 binaries FROM Icecast.org NOT the ones LT installed.
The standard distros before Bionic do NOT have ssl compiled in.
Upcoming Icecast 2.5 has SSL in by default.

hairmare:
I started work on this in https://github.com/radiorabe/libretime/commit/d0357a47d3489dfac2bdc0c9ac79a853d66f1e12 but didn't get around to really testing any of it.

maxlinux2000:
Hi all, 
I have seen that you also have the problem that basic icecast does not have SSL support.
[...]

PR #1986, opened 2022-07-24
feat: use file based stream configuration

### LINKED ISSUE #1977: Use Jinja2 to template the liquidsoap config
For now this should allow to easily edit some configs of the liquid soap script using data from the config file or elsewhere.

In the future, this should allow cleaner setup with ability to enable / disable features more easily.

jooola:
Fixes  #1977 #1537
Related to #1451 #1032 #513


jooola:
I am having a hard time allowing to load env variables for Union types and arrays of objects (an array of Union of objects). We could really use allowing to override the output stream host names using env variables, so I am unsure if we can work around this.

It's an interesting problem though, I might need some extra time.

jooola:
This changes is quite big and many decisions needs to be taken. I am currently stuck at decision making and not being able to move forward. This is also defines the config schema, so I prefer not miss anything.

We have to keep in mind that:
- We can leverage the power of yaml anchors to help users to configure there streams.
- Settings that cannot be changed after Liquidsoap is started should not be editable from the UI.

Some of thinking I have done:
- I am not sure if we want to drop this 'editable' feature from the UI for now and re implement it once we released v3, because it requires a lot of changes to properly be integrated in legacy. My problem would be that we might have to deprecated fields from the config file and ask people the change again. Having a way to handle both config+UI seem handy though, but not a trivial change.
- The offline message should be editable from the UI, we should be able to edit after liquidsoap started, but it isn't implemented yet.
- Streams name/description/url/genre settings seem not changeable once liquidsoap started. So we need them in the config file.  Those are annoying to configure though, so I wonder if we should provide a default template in case of undefined values or leave it to the user to set a yaml anchor. I am unsure of there purpose too, they add details to the icecast main page, and might be used in the yellow pages listing, but I don't if this feature is really used ? I don't know if user use Icecast to get the proper metadata to there website, they might use the /api instead ?
- The legacy stream settings view needs changing, but I don't know in which direction. Having the stream status available for checking is helpful and reporting errors to the UI is a good idea. We also need to present details in the UI for users to setup there streaming software, such as the input settings. Some settings will still be editable such as replaygain/offair message/banner template. I probably need a list of output stream with there status below, and some key configuration details for those streams such as mount point/format/quality/name and description/and public URL. The rest should probably be kept private. (Maybe this is the place for starting to embed a new UI :rocket: ?)
- Some stream settings are spread across the stream settings table and the preferences table, I wonder if we should merge everything together in the pref table or move the few stream settings in the pref table to the stream settings.
- For now, we will probably require an API call from liquidsoap to the API to get some of the editable settings, so we might not need to try to move every setting to the config file.

For now, what is working:
- liquidsoap starts, generate the entrypoints properly and streams properly
- playout properly fetch the listeners stats, while I noticed it will only get the stats for a single endpoint.

Work left todo is mostly on the legacy/api side, which might require large changes to the stream settings view/model.

@paddatrapper I could use your thoughts on each of these points.

paddatrapper:
> I am not sure if we want to drop this 'editable' feature from the UI for now and re implement it once we released v3, because it requires a lot of changes to properly be integrated in legacy. My problem would be that we might have to deprecated fields from the config file and ask people the change again. Having a way to handle both config+UI seem handy though, but not a trivial change.

Which editable features are you wanting to keep? As far as I know, all stream settings require a liquidsoap restart to reflect. If so, then having them in the config file only makes sense and we can drop the editable feature

> The offline message should be editable from the UI, we should be able to edit after liquidsoap started, but it isn't implemented yet.

I agree, but this can be implemented at a later point I think. Removing a single line from the config isn't a major breaking change. If we can find a way for the config line to remain, that is great, but I'm not too concerned about removing it later.

> Streams name/description/url/genre settings seem not changeable once liquidsoap started. So we need them in the config file. Those are annoying to configure though, so I wonder if we should provide a default template in case of undefined values or leave it to the user to set a yaml anchor. I am unsure of there purpose too, they add details to the icecast main page, and might be used in the yellow pages listing, but I don't if this feature is really used ? I don't know if user use Icecast to get the proper metadata to there website, they might use the /api instead ?

They also populate the stream metadata viewable in things like VLC. Perhaps we provide a set of templated values and the configuration file maintains a string that the user can configure with those variables? E.g. `metadata: '{ artist } - { song }'`

> The legacy stream settings view needs changing, but I don't know in which direction. Having the stream status available for checking is helpful and reporting errors to the UI is a good idea. We also need to present details in the UI for users to setup there streaming software, such as the input settings. Some settings will still be editable such as replaygain/offair message/banner template. I probably need a list of output stream with there status below, and some key configuration details for those streams such as mount point/format/quality/name and description/and public URL. The rest should probably be kept private. (Maybe this is the place for starting to embed a new UI rocket ? Only kidding haha)

I agree - show as read-only the useful settings like connection status, public URLs, input streams, and as read/write the database-kept settings. Keep the rest private

> Some stream settings are spread across the stream settings table and the preferences table, I wonder if we should merge everything together in the pref table or move the few stream settings in the pref table to the stream settings.

What settings are they?

> For now, we will probably require an API call from liquidsoap to the API to get some of the editable settings, so we might not need to try to move every setting to the config file.

We have this already from playout to get the DB values, don't we?

> How do we validate the streams settings such as audio bitrate depending on the stream format ? The stream settings UI had some validation that helped the user. We will need good documentation for users, but validation seem required somewhere for inexperienced users. Ideally we need the service/container to fails hard if stream settings are wrong.

Probably on playout startup - have explicit errors if things are wrong and fail hard

Thanks for all the work!

jooola:
Depends on #2039

jooola:
Sorry if I didn't answer you properly. I had to let this sit some day, and had the change to tackle one small problem at the time. In the end my questions found there answers.

This isn't the cleanest PR for the legacy code as I didn't want to dive too much into the php code, which scared me when I first started working on this. The StreamSettings class now query the config file instead of the database, everything else is untouched. I've removed most of the write operation in the stream settings and around (forms/...).

What is left to do is:
- update the messages sending from legacy to playout. 
- update some left over from models schema updates. 
- probably some extra cleanup.

legacy/application/models/RabbitMq.php
```diff
@@ -57,6 +57,32 @@ public static function SendMessageToPypo($event_type, $md)
...
+        $md = ['transition_fade' => $fade];
+
+        $eventType = 'update_stream_format';
+        $md = ['stream_format' => $type];
+
+        $exchange = 'airtime-pypo';
+        $data = json_encode($md, JSON_FORCE_OBJECT);
+        self::sendMessage($exchange, 'direct', true, $data);
+    }
+
```
jooola, line 85: NOTE: To remove, I'll improve this in another PR.

legacy/application/models/StreamSetting.php
```diff
@@ -2,70 +2,94 @@
...
+            $prefix . 'admin_user' => $output['admin_user'] ?? 'admin',
+            $prefix . 'admin_pass' => $output['admin_password'] ?? '',
+            $prefix . 'channels' => $output['audio']['channels'] ?? 'stereo',
+            $prefix . 'bitrate' => $output['audio']['bitrate'] ?? 128,
+            $prefix . 'type' => $output['audio']['format'],
+            $prefix . 'name' => $output['name'] ?? '',
+            $prefix . 'description' => $output['description'] ?? '',
+            $prefix . 'genre' => $output['genre'] ?? '',
+            $prefix . 'url' => $output['url'] ?? '',
+            $prefix . 'mobile' => 'false',
```
jooola, line 64: This will always disable a feature that I am unaware off, I could use some input to know what this is really for and if we want to keep it !

playout/libretime_playout/config.py
```diff
@@ -26,3 +31,4 @@ class Config(BaseConfig):
     general: GeneralConfig
     rabbitmq: RabbitMQConfig = RabbitMQConfig()
     playout: PlayoutConfig = PlayoutConfig()
```
jooola, line 33: The liquidsoap host and port will be moved in in own configuration schema.

playout/libretime_playout/liquidsoap/entrypoint.liq.j2
```diff
@@ -0,0 +1,61 @@
...
+{%- endif %}
+
+set("server.telnet", true)
+set("server.telnet.port", 1234)
+
+{% if version >= (1, 3, 3) -%}
+set("harbor.bind_addrs", ["0.0.0.0"])
+{%- else -%}
+set("harbor.bind_addr", "0.0.0.0")
+{%- endif %}
```
jooola, line 53: NOTE: Allow custom telnet server host from configuration file.

playout/libretime_playout/liquidsoap/entrypoint.liq.j2
```diff
@@ -0,0 +1,61 @@
...
+
+off_air_meta = "{{ settings.message_offline }}"
```
jooola, line 55: NOTE: Make this variable dynamic, and allow change from playout.

playout/libretime_playout/liquidsoap/entrypoint.py
```diff
@@ -1,64 +1,82 @@
...
-    max_attempts = 10
-    successful = False
+class StreamSettings(BaseModel):
+    station_name: str
+    input_fade_transition: float
+    message_format: MessageFormatKind
+    message_offline: str
+    source_input_main_streaming: bool
+    source_input_show_streaming: bool
+    source_schedule_streaming: bool
```
jooola, line 43: Sharing those models across api and playout is not a recommended pattern, we will have to maintain models on both sides.

playout/libretime_playout/timeout.py
```diff
@@ -14,27 +12,15 @@ def __init__(self):
...
-                fetch.PypoFetch.ref.restart_liquidsoap()
-            else:
-                raise Exception("Thread did not terminate")
-        else:
-            return thread.result
-
-        first_attempt = False
+            raise Exception("Thread did not terminate")
+
+        return thread.result
```
jooola, line 23: This whole timeout decoration might have become overkill, could probably be handled with a context manager and socket timeout.

legacy/application/configs/conf.php
```diff
@@ -89,6 +93,77 @@ public function getConfigTreeBuilder()
...
+            /*  */->scalarNode('mount')->defaultValue("show")
+            /*      */->validate()->ifString()->then($trim_leading_slash)->end()
+            /*  */->end()
+            /*  */->integerNode('port')->defaultValue(8002)->end()
+            /**/->end()->end()
+            ->end()->end()
+
+            // Stream outputs
+            ->arrayNode('outputs')->arrayPrototype()->children()
+            /**/->booleanNode('enabled')->defaultFalse()->end()
```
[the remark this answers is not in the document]
jooola: Postponed for a future PR.

installer/config.yml
```diff
@@ -61,3 +61,46 @@ playout:
...
+    name: LibreTime!
+    description: LibreTime Radio!
+    url: https://libretime.org
+    genre: various
+
+    admin_user: admin
+    admin_password: hackme
+
+  # Output streams, max outputs is 3
+  outputs:
```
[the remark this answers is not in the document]
paddatrapper: Hmm... I like the idea of rolling everything into a single outputs section. We could split it up under that or add a `type` field. I've seen it done both ways and I'm not sure which is better. I think the cleanest is probably

```yaml
outputs:
  hls:
    - server:
  icecast:
    - server:
  shoutcast:
    - server:
```

Yeah, I don't know who actually uses shoutcast...

legacy/application/models/StreamSetting.php
paddatrapper: No idea what that is...

playout/libretime_playout/liquidsoap/entrypoint.py
paddatrapper: Can they not go in `shared`?

jooola: This would be great, but I am afraid of few things :
-  If we want to modify a model in one service, we have to apply the change to the other of the services using it. That is doable but commits might not represent what they supposed to changes as many other parts will be edited. Being able to modify an api and only map old => new field on model setup seem easier. This is a really important in micro services architecture, but I think it can be applied in our case, as we shared if for both the api and playout.

- Sharing logic might be a good idea for small things, but I intend to use some of those models to build advanced features that are tightly integrated in the services using it, pushing everything in the shared lib will just recreate a api-client like library were we had the playout code is spread in multiple packages.

We might be able to share those models as data structure only and inherit the classes to write some extra features, but we will have to be careful not to make it hard to use in the long run.

Also, maybe I am overthinking this and we can share it and see how it goes.

legacy/application/models/StreamSetting.php
jooola: Ok, lets just leave it disabled for now.

playout/libretime_playout/liquidsoap/entrypoint.py
paddatrapper: Maybe we add it to shared for now and if things expand or get more complicated we can refactor things out?

jooola: Ok !

playout/libretime_playout/timeout.py
jooola: Will be refactored on a future PR

playout/libretime_playout/liquidsoap/entrypoint.py
jooola: Back to this, I think we are already hitting a too complex scenario, for example the api preferences. 

For example:
https://github.com/jooola/libretime/blob/4b17894c6e98ff76fabebdc4987ef7bb09d1f4ec/api/libretime_api/core/models/preference.py#L17-L29

will be extend really soon, when we will uncomment fields. Playout don't need some of the extra fields but with the validation system, the values will still be required as they aren't marked as optional.

So there is illusion of duplication for now, but both models will diverge really soon. I don't think we should merge this into the shared lib.

In addition preferences models will be used all over the place, maybe we should not add complexity from the start.

If we want to make sure the models are in sync for the few fields they share, we should probably create tests instead.

paddatrapper: Ok, sounds reasonable to keep them separate then

installer/config.yml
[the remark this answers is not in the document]
jooola: Do we agree on the following config ?
```yml
  # Default output
  .default_output: &default_output
    kind: icecast
    source_user: source
    source_password: hackme
    admin_user: admin
    admin_password: hackme

    name: LibreTime!
    description: LibreTime Radio!
    website: https://libretime.org
    genre: various

  outputs:
    - <<: *default_output
      enabled: true
      mount: main.ogg
      audio:
        format: ogg
        bitrate: 256

    - <<: *default_output
      enabled: true
      mount: main.mp3
      audio:
        format: mp3
        bitrate: 256

    - kind: shoutcast
      enabled: true
      source_password: password
      audio:
        format: mp3
        bitrate: 256

    - kind: alsa
      enabled: true
```

The outputs field should never be used in python but properties filtering this field can. This is specially required when listing outputs and guessing legacy ids based on there index, as we have to build up the `s1`, `sX` values from these lists.

For PHP, we will have to preprocess the outputs list, as the symfony validation component does not handle [Union type with a discriminator field](https://github.com/jooola/libretime/blob/4b17894c6e98ff76fabebdc4987ef7bb09d1f4ec/shared/libretime_shared/config/_models.py#L191-L194) to sort them out.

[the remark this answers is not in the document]
jooola: Hmm, going back and forth.

Why don't we want to have `stream.line_outputs`, `stream.icecast_outputs` or `stream.hls_outputs` ?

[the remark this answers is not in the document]
paddatrapper: That was what I was suggesting, except that `_outputs` is unnecessary in the name because they are under the `outputs` key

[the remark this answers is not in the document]
jooola: I might be bike shedding here, but I was hoping not to add too much nesting.

api/libretime_api/permissions.py
```diff
@@ -98,10 +115,12 @@ def has_permission(self, request, view):
...
+
+        return super().has_permission(request, view)
 
     def has_object_permission(self, request, view, obj):
         if request.user and request.user.is_authenticated:
             perm = get_permission_for_view(request, view)
             return request.user.has_perm(perm, obj)
-        return check_authorization_header(request)
+
+        return super().has_object_permission(request, view, obj)
```
jooola, line 126: I am a bit lost with our custom permission management system. IS there a way to simplify it ?

I also wonder if we currently expose way too much data to the public, shouldn't we only allow the api key auth and admin auth for now and slowly grant permissions when needed ?

jooola:
This is ready for review and lots of testing !

api/libretime_api/permissions.py
paddatrapper: Currently this mimics the legacy permissions - ensuring that users have access to things via the API that they would have access to via the UI. Nothing sensitive should be publicly accessible

jooola:
While the PR is "working", here is what is left todo:
- review each change and do some extra cleaning/refactoring.
- test this PR with weird configs to find uncovered edge cases.
- write more documentation once we are happy with the code.
- write an upgrade procedure.

installer/config.yml
```diff
@@ -85,3 +85,170 @@ playout:
...
   record_sample_size: 16
+
+stream:
+  # Inputs sources.
+  inputs:
+    # Main harbor input.
+    main:
+      # Harbor input public url. If not defined, the value will be generated from
+      # the [general.public_url] hostname and the input port and mount.
+      url:
```
jooola, line 96: Unsure about the naming of this fields, maybe url is too generic. Maybe we should use public_url or stream_url ?

installer/config.yml
```diff
@@ -85,3 +85,170 @@ playout:
...
+      - <<: *default_icecast_output
+        enabled: false
+        mount: extra
+
+      - # Whether the output is enabled.
+        # > default is false
+        enabled: false
+        # Output public url, If not defined, the value will be generated from
+        # the [general.public_url] hostname and the output port and mount.
+        url:
```
jooola, line 151: Same as for the input url, maybe we should have public_url ?

installer/config.yml
```diff
@@ -85,3 +85,170 @@ playout:
...
+        # Shoutcast stream name.
+        name: LibreTime!
+        # Shoutcast stream website.
+        website: https://libretime.org
+        # Shoutcast stream genre.
+        genre: various
+
+    # Line outputs.
+    # > max items is 1
+    line:
```
jooola, line 247: Is line outputs a good name ? Should we consider hardware output ? Unsure which name is the best to group all the system outputs.

installer/config.yml
paddatrapper: `public_url` I think is a good name here

installer/config.yml
jooola: Maybe `device` or `system`. We have to group the following alsa, ao, oss, portaudio, pulseaudio.

This list could probably be extended at some point, with gstreamer, and other outputs supported by liquidsoap.

paddatrapper: I like `system` , think it describes what it is better than device

jooola:
Unsure if we can provide much help for the upgrade procedure, hope this is enough.

Anyway, I tested this on bullseye, its ready to be reviewed.

docs/admin-manual/setup/configuration.md
```diff
@@ -221,6 +221,240 @@ In order to apply the changes made in this section, please restart the following
 libretime-liquidsoap
 ```
 
+## Stream
+
+The `stream` section configure anything related to the inputs and outputs streams of the project.
```
paddatrapper, line 226: configures

docs/admin-manual/setup/configuration.md
```diff
@@ -221,6 +221,240 @@ In order to apply the changes made in this section, please restart the following
...
+In order to apply the changes made in this section, please restart the following services:
+
+```
+libretime-liquidsoap
+libretime-playout
+```
+
+:::info
+
+To help you simplify your stream configuration, you can use yaml anchors to define a common properties and reuse them in your outputs definition:
```
paddatrapper, line 243: output definitions

docs/admin-manual/setup/configuration.md
```diff
@@ -221,6 +221,240 @@ In order to apply the changes made in this section, please restart the following
...
+        audio:
+          format: "mp3"
+          bitrate: 256
+```
+
+:::
+
+### Inputs
+
+The `stream.inputs` section configure anything related to the inputs streams of the project.
```
paddatrapper, line 277: configures

input streams

of LibreTime maybe?

docs/admin-manual/setup/configuration.md
```diff
@@ -221,6 +221,240 @@ In order to apply the changes made in this section, please restart the following
...
+
+```yml
+stream:
+  # Inputs sources.
+  inputs:
+    # Main harbor input.
+    main:
+      # Harbor input public url. If not defined, the value will be generated from
+      # the [general.public_url] hostname and the input port and mount.
```
paddatrapper, line 286: hostname, the input port and mount

docs/admin-manual/setup/configuration.md
```diff
@@ -221,6 +221,240 @@ In order to apply the changes made in this section, please restart the following
...
+      # > default is main
+      mount: "main"
+      # Listen port for the main harbor input.
+      # > default is 8001
+      port: 8001
+
+    # Show harbor input.
+    show:
+      # Harbor input public url. If not defined, the value will be generated from
+      # the [general.public_url] hostname and the input port and mount.
```
paddatrapper, line 298: hostname. the input port and mount

docs/admin-manual/setup/configuration.md
```diff
@@ -221,6 +221,240 @@ In order to apply the changes made in this section, please restart the following
...
+      # > default is show
+      mount: "show"
+      # Listen port for the show harbor input.
+      # > default is 8002
+      port: 8002
+```
+
+### Outputs
+
+The `stream.outputs` section configure anything related to the outputs streams of the project.
```
paddatrapper, line 310: output streams of LibreTime

docs/admin-manual/setup/configuration.md
```diff
@@ -221,6 +221,240 @@ In order to apply the changes made in this section, please restart the following
...
+stream:
+  outputs:
+    # Icecast output streams.
+    # > max items is 3
+    icecast:
+      - # Whether the output is enabled.
+        # > default is false
+        enabled: false
+        # Output public url, If not defined, the value will be generated from
+        # the [general.public_url] hostname and the output port and mount.
```
paddatrapper, line 335: hostname, the output port and mount

docs/admin-manual/setup/configuration.md
```diff
@@ -221,6 +221,240 @@ In order to apply the changes made in this section, please restart the following
...
+        # Icecast source user.
+        # > default is source
+        source_user: "source"
+        # Icecast source password.
+        # > this field is REQUIRED
+        source_password: "hackme"
+        # Icecast admin user.
+        # > default is admin
+        admin_user: "admin"
+        # Icecast admin password. If not defined, stats will not be collected.
```
paddatrapper, line 355: statistics?

docs/admin-manual/setup/configuration.md
```diff
@@ -221,6 +221,240 @@ In order to apply the changes made in this section, please restart the following
...
+        name: "LibreTime!"
+        # Shoutcast stream website.
+        website: "https://libretime.org"
+        # Shoutcast stream genre.
+        genre: "various"
+```
+
+#### System
+
+The `stream.outputs.system` section configure the system outputs.
```
paddatrapper, line 441: configures

docs/admin-manual/setup/configuration.md
```diff
@@ -221,6 +221,240 @@ In order to apply the changes made in this section, please restart the following
...
+        description: "LibreTime Radio!"
+        # Icecast stream website.
+        website: "https://libretime.org"
+        # Icecast stream genre.
+        genre: "various"
+```
+
+#### Shoutcast
+
+The `stream.outputs.shoutcast` section configure the shoutcast output streams.
```
paddatrapper, line 388: configures

docs/admin-manual/setup/configuration.md
```diff
@@ -221,6 +221,240 @@ In order to apply the changes made in this section, please restart the following
...
+  # Output streams.
+  outputs:
+    icecast: # See the [stream.outputs.icecast] section.
+    shoutcast: # See the [stream.outputs.shoutcast] section.
+    system: # See the [stream.outputs.system] section.
+```
+
+#### Icecast
+
+The `stream.outputs.icecast` section configure the icecast output streams.
```
paddatrapper, line 323: configures

docs/admin-manual/setup/configuration.md
paddatrapper: configures

docs/admin-manual/setup/configuration.md
paddatrapper: configures

docs/releases/unreleased.md
```diff
@@ -36,6 +36,24 @@ Along with the Debian Buster deprecation, the following dependencies versions ar
 - [php7.3](https://packages.debian.org/buster/php7.3)
 - [python3.7](https://packages.debian.org/buster/python3)
 
+## :arrow_up: Before upgrading
+
+:::caution
+
+Please run this **before the upgrade procedure**!
```
paddatrapper, line 43: Maybe `follow this`, as there aren't actually commands to run?

paddatrapper:
I've done some vagrant testing on this too and looks like it is working consistently

paddatrapper:
Do we want to leave the settings in the DB for a release? I.e. do not use them, but do not drop the table either. On the other hand people should be making a backup and reading release notes before upgrading...

jooola:
> I've done some vagrant testing on this too and looks like it is working consistently

Which distro did you test ? I also started testing again this morning. With the recent bionic compatibility problems I decided to properly test on it.

> Do we want to leave the settings in the DB for a release? I.e. do not use them, but do not drop the table either. On the other hand people should be making a backup and reading release notes before upgrading...

I'm fine either way, we can postpone the table cleanup for 3.0.0. We could also dump the table in a file "in case of emergency" before the upgrade procedure.

It doesn't require much work and might help those who don't read the documentation...

paddatrapper:
I tested on focal and bullseye.

I think either of those options are good

jooola:
- Did a test on bionic, nothing obvious is broken !
- Fixed legacy to only use the config `inputs.*.public_url` if they are truthy (not empty strings)
- Remove the database migration that flushes the stream settings, and added commands that helps user to view this data.

https://github.com/libretime/libretime/compare/dd97598d74efff42bf48940cae5762015a69dd61..1f65fc9baadb245ad3afd6b3f55a8557c25bc578

paddatrapper (approved):
Sounds good!

jooola:
I guess we are good for merging ?

PR #2001, opened 2022-07-26
fix(api): update set_icecast_passwords StreamSetting fields

jooola:


jooola:
This is an urgent fix, gonna merge it.

paddatrapper:
Can we add tests for this?

jooola:
Yup, second time I forget about it, I'll work on the stream config file instead. The command will be useless by then.

PR #2004, opened 2022-07-27
chore: install local python deps in editable mode

jooola:


paddatrapper:
Why do you want to do this?

jooola:
The previous way of installing the deps was by building a wheel and installing the wheel in the app venv.
This does not allow updating the dependency code and have the changes reflected on the app without reinstalling the wheel.

Now we can update the shared lib for example and the changes will directly be reflected.

For the chdir removal, it was introduced when the setup tool was installing the sysv init scripts and didn't used the entrypoint option (they used the script option which seemed to use a relative path).

PR #2009, opened 2022-07-30
feat(shared): load env config using jsonschema

jooola:


jooola:
The code could be refactored some more, but I'll try to do this some other time.

It is working with some limitations that won't bother us for now. For example it doesn't support unions of different shapes `Union[List[Any], Dict[str, Any]]` or `Union[str, Dict[str, Any]]`. But we don't have any of such use case for now, so we can always implement it once we need it.

jooola:
I'll try to do more refactoring this evening.

jooola:
I refactored the code enough, the few duplication are now removed and the code is comprehensible.

Ready to be reviewed/merged.

PR #2011, opened 2022-07-31
chore: generate a config.schema.json

jooola:
I am unsure about this, it is handy to have the config file checked against a schema in an editor, but I might be the only one using it.

I wonder how much users will edit there config files from a editor that supports validating using a schema.


jooola:
Maybe generating the documentation from the json schema is a good idea.

jooola:
This is not worth it.

PR #2012, opened 2022-07-31
fix(shared): remove unused field from rabbitmq config

jooola:


paddatrapper:
Will parsing fail if that field is there?

jooola:
No it should not, I was thinking we might want to forbid extra fields for submodels, but that not enabled now.

PR #2013, opened 2022-08-01
feat(installer): use ed for config update

jooola:
This allow to edit deeply nested values in the config file, which will be required for setting the icecast passwords.

### DIFF
--- a/install
+++ b/install
@@ -234,14 +234,15 @@ set_config() {
   value="${1}" && shift
 
   # Build sed query
-  query="/^${1}:/"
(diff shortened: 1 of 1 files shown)


paddatrapper:
We may want to look at using [yq](https://github.com/mikefarah/yq) eventually

paddatrapper:
It's on my list to eventually package

jooola:
> It's on my list to eventually package

https://salsa.debian.org/go-team/packages/yq

PR #2027, opened 2022-08-08
refactor(playout): improve exceptions handling

jooola:
Mainly exception handling improvements

### DIFF
--- a/playout/libretime_playout/liquidsoap/entrypoint.py
+++ b/playout/libretime_playout/liquidsoap/entrypoint.py
@@ -1,7 +1,6 @@
 import os
 import sys
 import time
-import traceback
 from pathlib import Path
 from typing import Optional
 
@@ -54,10 +53,8 @@ def generate_entrypoint(log_filepath: Optional[Path]):
             ss = legacy_client.get_stream_setting()
             generate_liquidsoap_config(ss, log_filepath)
             successful = True
-        except Exception as e:
-            print("Unable to connect to the Airtime server.")
-            logger.error(str(e))
-            logger.error("traceback: %s", traceback.format_exc())
+        except Exception:
+            logger.exception("Unable to connect to the Airtime server")
             if attempts == max_attempts:
                 logger.error("giving up and exiting...")
                 sys.exit(1)
--- a/playout/libretime_playout/player/liquidsoap.py
+++ b/playout/libretime_playout/player/liquidsoap.py
@@ -64,9 +64,9 @@ class PypoLiquidsoap:
             try:
                 self.telnet_liquidsoap.queue_push(available_queue, media_item)
                 self.liq_queue_tracker[available_queue] = media_item
-            except Exception as e:
-                logger.error(e)
-                raise
+            except Exception as exception:
+                logger.exception(exception)
+                raise exception
         else:
             logger.warning(
                 "File %s did not become ready in less than 5 seconds. Skipping...",
@@ -87,13 +87,12 @@ class PypoLiquidsoap:
 
     def find_available_queue(self):
         available_queue = None
-        for i in self.liq_queue_tracker:
-            mi = self.liq_queue_tracker[i]
-            if mi == None or self.is_media_item_finished(mi):
+        for queue_id, item in self.liq_queue_tracker.items():
+            if item is None or self.is_media_item_finished(item):
                 # queue "i" is available. Push to this queue
-                available_queue = i
+                available_queue = queue_id
 
-        if available_queue == None:
+        if available_queue is None:
             raise NoQueueAvailableException()
 
         return available_queue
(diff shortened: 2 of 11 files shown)


playout/libretime_playout/player/liquidsoap.py
```diff
@@ -192,8 +191,8 @@ def verify_correct_present_media(self, scheduled_now):
                 # something is playing and it shouldn't be.
                 self.telnet_liquidsoap.stop_web_stream_buffer()
                 self.telnet_liquidsoap.stop_web_stream_output()
-        except KeyError as e:
-            logger.error("Error: Malformed event in schedule. " + str(e))
+        except KeyError as exception:
+            logger.error("Error: Malformed event in schedule. " + str(exception))
```
paddatrapper, line 195: Log the message and then the exception or log it all as exception?

playout/libretime_playout/player/push.py
```diff
@@ -49,9 +48,9 @@ def main(self):
         while True:
             try:
                 media_schedule = self.queue.get(block=True)
-            except Exception as e:
-                logger.error(str(e))
-                raise
+            except Exception as exception:
+                logger.error(exception)
```
paddatrapper, line 52: logger.exception

playout/libretime_playout/player/push.py
```diff
@@ -119,17 +118,15 @@ def stop_web_stream_all(self):
             tn.write("exit\n")
             logger.debug(tn.read_all())
 
-        except Exception as e:
-            logger.error(str(e))
+        except Exception as exception:
+            logger.error(exception)
```
paddatrapper, line 122: logger.exception

playout/libretime_playout/recorder.py
```diff
@@ -161,8 +158,8 @@ def run(self):
 
                 self.upload_file(filepath)
                 os.remove(filepath)
-            except Exception as e:
-                logger.error(e)
+            except Exception as exception:
+                logger.error(exception)
```
paddatrapper, line 162: logger.exception

playout/libretime_playout/recorder.py
```diff
@@ -185,8 +182,8 @@ def __init__(self, q, config: Config, legacy_client: LegacyClient):
             try:
                 self.legacy_client.register_component("show-recorder")
                 success = True
-            except Exception as e:
-                logger.error(str(e))
+            except Exception as exception:
+                logger.error(exception)
```
paddatrapper, line 186: logger.exception

jooola:
Fixed !

playout/libretime_playout/liquidsoap/entrypoint.py
```diff
@@ -54,10 +53,8 @@ def generate_entrypoint(log_filepath: Optional[Path]):
             ss = legacy_client.get_stream_setting()
             generate_liquidsoap_config(ss, log_filepath)
             successful = True
-        except Exception as e:
-            print("Unable to connect to the Airtime server.")
-            logger.error(str(e))
-            logger.error("traceback: %s", traceback.format_exc())
+        except Exception:
+            logger.exception("Unable to connect to the Airtime server")
```
paddatrapper, line 57: Should we log the exception traceback?

jooola: The traceback will always be present in the context of an exception handling. The string parameter will be title of the log entry with the traceback below.

```py
try:
    data["something"]
except KeyError:
    logger.exception(f"something key is missing from data")
```
Gives
```
$ python3 ./test.py   
2022-08-09 20:34:18.948 | ERROR    | __main__:<module>:12 - something key is missing from data

> File "/home/jo/git/github.com/libretime/libretime/shared/./test.py", line 10, in <module>
    data["something"]
    └ {}

KeyError: 'something'
```

```py
try:
    data["something"]
except KeyError as exception:
    logger.exception(f"something key is missing from data: {exception}")
```
Gives
```
$ python3 ./test.py
2022-08-09 20:34:41.934 | ERROR    | __main__:<module>:12 - something key is missing from data: 'something'

> File "/home/jo/git/github.com/libretime/libretime/shared/./test.py", line 10, in <module>
    data["something"]
    └ {}

KeyError: 'something'
```

I've been using the 2nd way in many other places, probably because it gives better context to the log message.

paddatrapper: Do you think we should standardise on a way to do this?

jooola: I don't think we need a "standard" way of doing this, it should be handled case by case.

Once we removed all the `broad exception` ™  to specific exceptions, we will be able to know whether an exception message should be printing within the log entry or not, to print the traceback or not. But for now I have to admit it is difficult.

PR #2029, opened 2022-08-09
refactor(analyzer): fix linting errors

jooola:

PR #2030, opened 2022-08-09
feat(analyzer): override paths using env variables

jooola:


paddatrapper:
This depends on #2029

jooola:
Rebased

PR #2039, opened 2022-08-10
style(legacy): format

jooola:


jooola:
This is a preparation for #1986

jooola:
I'll try to format all the files once and for all.

PR #2040, opened 2022-08-11
feat(legacy): setup config schema validation

jooola:
BREAKING CHANGE: Unrecognized values in the configuration file will
raise validation errors, please make sure to cleanup your configuration
file.


legacy/.php-cs-fixer.php
```diff
@@ -2,6 +2,7 @@
 
 $finder = PhpCsFixer\Finder::create()
     ->in(__DIR__)
+    ->notPath('application/configs/conf.php')
```
jooola, line 5: ... This is only used to allow the weird indent I made using comments. Otherwise its difficult to understand where in the schema you are.

PR #2051, opened 2022-08-12
feat(playout): create liquidsoap client

jooola:
Create both a telnet based client and a unix socket based client for communication with liquidsoap.

The connections class will only handle IO, the client should implement any playout <> liquidsoap logic, like skipping tracks and changing liquidsoap variables.


jooola:
Would be nice to be able to run the same test matrix locally. Having to wait on the CI to complex is verbose and slow.

jooola:
Depends on #2055

jooola:
This still needs some extra testing, but I could use some of your input @paddatrapper as this is mostly done.

tools/python.mk
```diff
@@ -74,3 +74,19 @@ install: $(VENV)
...
+			--workdir /src/$(APP) \
+			ghcr.io/jooola/libretime-dev:$(DISTRO)
+
+docker-dev:
+	$(MAKE) clean
+	$(DOCKER_RUN) bash
+
+docker-test:
+	$(MAKE) clean
+	$(DOCKER_RUN) make test
```
jooola, line 92: I've been using this for testing on other distros, not sure we should merge it though.

playout/libretime_playout/liquidsoap/client/_connection.py
```diff
@@ -0,0 +1,132 @@
...
+            raise exception
+
+    def __exit__(self, _exc_type, _exc_value, _traceback):
+        self.close()
+
+    def connect(self):
+        # A connection shouldn't need longer than a few seconds to send,
+        # receive and close. If for some reason another thread does not release the
+        # lock, we make sure this timeout.
+        self._lock.acquire(timeout=self._timeout + 5)
```
jooola, line 68: NOTE: this does not raise but returns whether the lock was acquired.

playout/libretime_playout/player/fetch.py
```diff
@@ -267,66 +199,15 @@ def update_liquidsoap_connection_status(self):
...
-            command = ("vars.stream_metadata_type %s\n" % stream_format).encode("utf-8")
-            logger.info(command)
-            tn.write(command)
-            tn.write(b"exit\n")
-            tn.read_all()
-        except Exception as exception:
-            logger.exception(exception)
-        finally:
-            self.telnet_lock.release()
+        self.liq_client.settings_update(message_format=stream_format)
```
jooola, line 202: All those changes unwrap liquidsoap calls from the broad exception handling, I wonder how and where this should be handled.

We should probably catch any ConnectionError and TimeoutError.

I don't know if we should log and ignore or fail hard ? Current behavior is log and ignore. 

Where this should be handled ? Maybe in the client context __exit__ exception handling.

tools/python.mk
paddatrapper: Would it be helpful for the development of other features?

jooola: Probably anytime we want to run tests on another distro, without spinning a full vagrant setup.

jooola: We will always be able to remove it.

playout/libretime_playout/main.py
```diff
@@ -144,18 +92,21 @@ def cli(log_level: str, log_filepath: Optional[Path], config_filepath: Optional[
...
+    liq_client = LiquidsoapClient(
+        host=config.playout.liquidsoap_host,
+        port=config.playout.liquidsoap_port,
+    )
 
-    liquidsoap_startup_test(telnet_lock, liquidsoap_host, liquidsoap_port)
+    logger.debug("Checking if Liquidsoap is running")
+    liq_version = liq_client.wait_for_version()
+    if not LIQUIDSOAP_MIN_VERSION <= liq_version:
+        raise Exception(f"Invalid liquidsoap version {liq_version}")
```
jooola, line 103: We do not loop indefinitely on invalid version or connection anymore. We fail hard and let the service manager restart the service if wanted.

More generally, I think we should aim for a fast and disposable app that can catch up its previous stats in case of restart. Catching up its state would mean without having to reset everything as it actually does, but simply read the current state and apply a diff update if necessary.

paddatrapper: I agree with that - it also aligns with HA practices where you cannot guarantee any single instance's state over a period of time

jooola:
Ok, I've tested based scheduling, stream settings update, everything is working as expected.

This is ready for review !

playout/Makefile
```diff
@@ -2,6 +2,7 @@ all: lint
 
 include ../tools/python.mk
 
+APP := playout
```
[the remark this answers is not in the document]
jooola: That's for the docker-dev and docker-test makefile target in `tools/python.mk`, should be added to the other apps if we want to ouse the docker based tasks.

PR #2060, opened 2022-08-16
chore: widen version range on python deps update

jooola:


jooola:
@paddatrapper You might have a better idea if this is smart. I was hoping that for example that the click deps upgrade widen the version range instead of forcing 8.1.3 ? This would let pip decide if the python version can install 8.1.x or should stick to 8.0.x.

paddatrapper:
Yeah, I think this is a good idea. We just need to be careful when we make changes that we continue to support all the versions in the range or bump the range. Maybe we should have a matrix of tests with the oldest and newest version of dependencies?

jooola:
> Yeah, I think this is a good idea. We just need to be careful when we make changes that we continue to support all the versions in the range or bump the range. Maybe we should have a matrix of tests with the oldest and newest version of dependencies?

That might generate really a lot of tests. Maybe we can have only an extra test matrix with the python3-* packages installed in a dev containers.

But I don't think we should worry too much, reading the change log and check when the feature we want to use was added should be enough.

paddatrapper:
Those python3-* packages probably won't be within the supported range, especially at this point.

jooola:
We agreed that only bullseye and onward should be in that range (with few exceptions), older distros will have to install from pip.

paddatrapper:
Ah right, yes. I forgot the docker image was based on Bullseye

jooola:
> Ah right, yes. I forgot the docker image was based on Bullseye

Haha, you lost me on this one, what do you mean ?

paddatrapper:
I mean I was working on the assumption that testing python3- packages would bring in versions older than the oldest version we support. This assumption was based on the understanding that the docker image was based on buster, not bullseye

jooola:
> I mean I was working on the assumption that testing python3- packages would bring in versions older than the oldest version we support. This assumption was based on the understanding that the docker image was based on buster, not bullseye

The docker images will always install from pip, we explicitly exclude the python packages from the docker build.

For testing I meant to have an extra `libretime-dev:bullseye` image with the python3-* packages installed and add it to the test matrix.

Is something missing in this PR ? Do you want me to setup this `libretime-dev:bullseye` packages testing job in the CI ?

Issue #2043, opened 2022-08-11
Setuptools 64.0.0 break pip editable install


jooola:
Setuptools 64.0.0 introduced the now PEP 660 editable feature, which seem at bit too fresh and buggy.

https://github.com/pypa/setuptools/pull/3488

I'll pin the setuptools version to ^63.0.0 for development.

PR #2066, opened 2022-08-21
chore: revert setuptools <64.0.0 pinning

jooola:


jooola:
upstream issue might be https://github.com/PyCQA/pylint/issues/7306

paddatrapper:
I assume they fixed the cause of #2043 in a later version, so this should probably blacklist 64.0.0?

jooola:
> I assume they fixed the cause of https://github.com/libretime/libretime/issues/2043 in a later version, so this should probably blacklist 64.0.0?

I was wrong saying 64.0.0 breaks the editable install, it only break pylint that raises an import error. The code is working, but pylint does not. I trusted pylint on this one and said everything is broken.

So only have to wait for pylint to handle the pep660 editable install and not raise this editable install error.

jooola:
I think we can work around this by forcing the legacy-editable feature in setuptools. We should be notified once https://github.com/PyCQA/pylint/issues/7306 is resolved.

Ready to be merged.

PR #2072, opened 2022-08-24
feat: optionally load systemd env from /etc/libretime/env

jooola:
This allow to override some configuration globally in the systemd services.

Useful to enable debug logs in all the python apps.


jooola:
This might be a dev only feature so maybe we should not blot the "production" services with that.

paddatrapper:
I definitely think that this is a development feature - it introduces another config file essentially, which we don't want for end-users

PR #2078, opened 2022-08-26
fix(installer): on bionic install uvicorn from pip

jooola:
I am in the train, so I cannot test this PR.


jooola:
Tried to run a test on vm at hetzner, seems like the bionic setup is completely broken. This will require more work.

PR #2090, opened 2022-09-06
feat: flush db stream settings data

jooola:
Remove data that has been deprecated in #1986

Migration filename will need renaming.


jooola:
Should we schedule this for 3.0.0 or 3.1.0 ?

paddatrapper:
Release next+1

jooola:
I think we should stop versioning the legacy schema with the libretime version. We cannot guess which version number will be next. Maybe we should use dates, `20221010.0.1`

paddatrapper:
Can do either date or just numerical increments

jooola:
Yes, you are right, we will use the same number as the migration file prefix

jooola:
Done, this is ready.

PR #2096, opened 2022-09-07
feat: move timezone preference to config file

jooola:
BREAKING CHANGE: The timezone preference moved to the configuration file.


paddatrapper:
Do we want to document what format the timezone should be in? We don't support UCT+2 for example, but do support Africa/Johannesburg

jooola:
Done, though I think timezone is explicit, hope nobody confuses timezone and zone offset.

PR #2107, opened 2022-09-08
feat(playout): improve graceful shutdown

jooola:


jooola:
Depends on #2156

jooola:
Tested on docker and vagrant working as expected !

PR #2116, opened 2022-09-12
chore: update version bump procedure

jooola:


Makefile
```diff
@@ -19,8 +19,8 @@ shell-check:
 VERSION:
 	tools/version.sh
 
-bump-python-version: VERSION
-	tools/bump-python-version.sh
+bump-python-version:
```
paddatrapper, line 22: Shouldn't this still depend on VERSION?

jooola: Yup, but we should actually not bump the python version during the CI job but in the release commit/tag. I removed the target.

PR #2147, opened 2022-09-15
fix: nginx depends on legacy in docker-compose

jooola:
Nginx is failing because of the missing php-fpm upstream server.

Do we prefer to have a 3rd layer of dependency (nginx > legacy > postgres and rabbitmq) or should we configure nginx to retry multiple times before failing ?


paddatrapper:
I think docker should resolve that level of dependency without too much issue

PR #2166, opened 2022-09-20
docs: add pulseaudio output in containers tutorial

jooola:


jooola:
This setup is not persisting a reboot. We should probably provide documentation that saves settings in a persisting way (e.g. the pulse audio server).

PR #2167, opened 2022-09-20
fix(legacy): rebuild favicon

jooola:


jooola:
This is the new favicon, but I zoomed in.
image
image

PR #2174, opened 2022-09-21
fix: include version variable inside containers

### LINKED ISSUE #2162: Missing VERSION file in the libretime-legacy container
**Describe the bug**
The legacy app relies on the VERSION file to provide this information to the users via the status page. This file is missing from the container image.

I am not sure if we should keep using the VERSION file. This information is already in the container tags or in the python packages, and the VERSION file is actually only used by the legacy app.

Now that the assets caching does not rely on the version number anymore, the requirement on the VERSION file is reduced, so we might want to hard code it into app as we do for the python setup.py files.

Passing the VERSION file inside the docker container is doable, but I wonder how to do this from within the Dockerfile, as we ignore any git related files. I would prefer not to have to run make VERSION before building the containers.

On the same note, we might also want to provide a version flag to the python apps to print the current version.

Not sure how cleanly we can handle both docker tags and app version inside the container. Leaving the version information outside the app makes it easier, as we can use `pip show libretime-playout` to get an app version, and `docker-compose images` to list current tags.

Anyhow, the version handling needs some improvements.

jooola:
Fix #2162

### DIFF
--- a/.github/workflows/actions/build-container/action.yml
+++ b/.github/workflows/actions/build-container/action.yml
@@ -22,6 +22,8 @@ runs:
         context: .
         pull: true
         push: ${{ github.event_name == 'push' }}
+        build-args: |
+          LIBRETIME_VERSION=${{ env.LIBRETIME_VERSION }}
         target: ${{ inputs.target }}
         tags: ${{ steps.meta.outputs.tags }}
         labels: ${{ steps.meta.outputs.labels }}
--- a/.github/workflows/container.yml
+++ b/.github/workflows/container.yml
@@ -30,6 +30,11 @@ jobs:
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
 
+      - name: Guess LIBRETIME_VERSION
+        run: |
+          make VERSION
+          echo "LIBRETIME_VERSION=$(cat VERSION | tr -d [:blank:])" >> $GITHUB_ENV
+
       - name: Build python-builder
         uses: docker/build-push-action@v3
         with:
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,3 +1,4 @@
+ARG LIBRETIME_VERSION
 #======================================================================================#
 # Python Builder                                                                       #
 #======================================================================================#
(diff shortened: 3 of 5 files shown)


jooola:
I am a bit confused on how we should generate the version. We should make it consistent and this PR isn't currently going in that direction.

We generate the following versions:

from tools/version.sh:
  - 3.0.0-beta.0 for a tag
  - 3.0.0-beta.0-dev+29ed76ac3 for a branch

and now in containers CI:
  - 3.0.0-beta.0-29ed76ac3dd1705dd3fdd99c120825d3dee93397 for a tag
  - main-6b16df8966bfb42ef248b8e8e33d7b5ffc80dd48 for a branch
 
I don't necessarily want to pull the whole repository to be able to properly find the latest tag in the container CI job, so we might need to agree on easier to guess versions for tagged release, versions for the main branch, and development versions.

jooola:
Maybe we should simplify everything and include the latest tagged version, the current branch, and the current sha.

For example changing the VERSION file to:
For a tagged release:

```json
{
  "version": "3.0.0-beta.0",
  "branch": null,
  "commit": "bdc2e924547e06f80be72b0a6097acb04cd30088"
}
```

For a branch version:

```json
{
  "version": null,
  "branch": "main",
  "commit": "6b16df8966bfb42ef248b8e8e33d7b5ffc80dd48"
}
```

Those information could also be read from env variables:

```
LIBRETIME_VERSION=3.0.0-beta.0
LIBRETIME_BRANCH=main
LIBRETIME_COMMIT=6b16df8966bfb42ef248b8e8e33d7b5ffc80dd48
```

For dev environment we could simply specify `LIBRETIME_VERSION=dev` or none

With that we should have everything to properly track the version run by users, while we don't have to hack script to build complex version strings.

jooola:
I don't know I'm trying to over engineer this. The idea might be good, but doesn't bring much. Back to a simple git describe.

jooola:
For this pull request the LIBRETIME_VERSION is build with the GITHUB_REF_NAME and the GITHUB_SHA: `LIBRETIME_VERSION=2174/merge-81232017fec6056499debad03648bc6a2b7cee0c` but those versions are not tagged/published.

On the LibreTime main branch we will have `LIBRETIME_VERSION=main-81232017fec6056499debad03648bc6a2b7cee0c`
On LibreTime releases we will have  `LIBRETIME_VERSION=3.0.0-beta.1`

paddatrapper:
Ok. I'm happy with that string

PR #2185, opened 2022-09-26
chore: add icecast stress script

jooola:
Unsure if this belongs here. Should we create a community repo ? For scripts and helpers ?


paddatrapper:
It's fine here until we have a lot of scripts that need to be managed separately

PR #2190, opened 2022-09-26
fix(legacy): add play button to stream player

### LINKED ISSUE #1027: Listen pop-up uses autoplay which is blocked by Firefox w/o permission
**Describe the bug**
This isn't so much a bug as something I noticed that may lead people to not be able to hear the stream. Newer versions of Firefox block autoplay of audio and you need to go and click "Allow autoplay on this site". The Listen pop-up uses this and thus audio will not be heard. There is also no play button to start the audio so there is a good chance that some people will think that the Listen button is broken or that their Libretime instance isn't producing sound.

**To Reproduce**
Open up LibreTime in the newest version of Firefox
Click Listen
No audio.
Click button on URL to check permissions and allow autoplay - it works

**Expected behavior**
We should either have some kind of pop-up that informs people they need to enable autoplay or have a fall-back where there is a play button that will circumvent this for people who don't hear anything and thus are inclined to hit play. We can keep autoplay enabled for who it works for or replace it by a play button if autoplay is just a bad way to do this sort of thing now.

jooola:
Fix #1027

### DIFF
--- a/legacy/application/controllers/DashboardController.php
+++ b/legacy/application/controllers/DashboardController.php
@@ -93,10 +93,6 @@ class DashboardController extends Zend_Controller_Action
 
     public function streamPlayerAction()
     {
-        $CC_CONFIG = Config::getConfig();
-
-        $baseUrl = Config::getBasePath();
-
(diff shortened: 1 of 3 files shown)


jooola:
image

PR #2198, opened 2022-09-28
chore: use forms for issue templates

jooola:
You can test it at https://github.com/jooola/libretime/issues/new/choose


paddatrapper:
I like it!

PR #2220, opened 2022-10-10
feat: drop distros support

jooola:
So here a big cleanup, but I was wondering if we want to split this into multiple PRs ?


jooola:
I'll force a CI run on each commit and slowly stack them in this branch.

jooola:
This is ready to be merged.

paddatrapper:
Yay for cleaner code :tada:

jooola:
We deprecated python 3.7 but I didn't drop it as I am unsure if we should bump the python version to >=3.8. The only reason would be to keep backward compatibility for people that have custom system, so advanced users.

We are not a library so I would go for raising the minimal version to 3.8, and follow our deprecation notice.

paddatrapper:
That is reasonable. I agree we should drop it

PR #2234, opened 2022-10-12
feat: add hls output stream

jooola:


jooola:
This POC is working with docker-compose, really cool to have HLS ! But I think we should rewrite the liquidsoap script before further additions

paddatrapper:
Hiding this behind a feature-flag is a good idea. Having some sort of feature-flag framework for experimental features in general may be something that we should look into

PR #2274, opened 2022-12-07
feat(legacy): enable strict_types=1

jooola:


jooola:
This is a small step towards php 8, feel free to contribute on my branch as I might not be the fastest with php and time available.

PR #2277, opened 2022-12-09
ci: enable renovate for 3.0.x

jooola:
Actually enable renovate for 3.0.x, the change must be done on the default branch, not on the target branch as I did here: 3694b316105a88c73b630a43304e55bc3c92f1a1


jooola:
Wouw, sorry for the spam

PR #2318, opened 2022-12-16
fix(legacy): add log entry on task run (#2316)

jooola:


jooola:
This is a back-port of an approved PR, merging.

PR #2319, opened 2022-12-16
fix(legacy): log errors on connect check failure (#2317)

jooola:


jooola:
This is a back-port of an approved PR, merging.

PR #2339, opened 2022-12-27
docs: add secure installation tutorial

jooola:


jooola:
We are now missing the PR preview command ^^

Here are some screenshots: 
image


image

paddatrapper:
The commands that require root should be prefixed with `sudo` to be consistent with the other documentation pages

jooola:
> There is a large section about how to install LibreTime, this feels like a duplicate of the general install page. Maybe point people there and include the differences here? e.g. the public_url. The same with the firewall and date/time stuff. Maybe we replace the default installation instructions with the secure, no reverse proxy instructions?

Yes, the duplication was intended, at first I wanted to create a blog post, to describe step by step how to install libretime on a specific system from start to finish a specific point in time. So it was meant (just like the digital ocean tutorials) to be frozen, and if updated to a newer system, duplicated and corrected.

The duplication from the real documentation is to have on one side a standard reference, and in the tutorial something really specific with command output and steps that we don't want to be part of the reference documentation such as the firewall setup.

This would be the one tutorial to guide newcomers by the hand.

You're probably right, maybe we cannot afford such duplication since the reference documentation is quite poor ^^.  I'll try to fix this in some way.

jooola:
I think we should not necessarily merge the 2 documentation, I am having a hard time merging the 2 simply because on is supposed to be generic, where you can pick the tools you need (reverse proxy/firewall/docker), while the other is really specific.

I don't know how properly handle this, but I would like to keep the generic aspect of the reference docs, but also provide help for those who need a step by step guide to a fully installed and secure server. This is where I like the django docs sstructure, that we try to follow https://docs.djangoproject.com/en/4.2/#how-the-documentation-is-organized

PR #2347, opened 2023-01-12
fix(legacy): move forked deps to the libretime namespace

### LINKED ISSUE #2346: Move forked php dependencies to the libretime org

A few php libraries that LibreTime is depending on are currently under my username namespace on Github.

While I don't plan to delete them as it would break a lot of things, I think it's better to move them under the LibreTime organization.

 

@paddatrapper What do you think is cleaner:
- fork under the Libretime org
- transfer them, as Github will probably automatically redirect composer to the new repository path

_No response_

### Additional context

https://github.com/jooola/propel1
https://github.com/jooola/celery-php

jooola:
Fix #2346

### DIFF
--- a/legacy/composer.json
+++ b/legacy/composer.json
@@ -16,11 +16,11 @@
   "repositories": [
     {
       "type": "vcs",
-      "url": "https://github.com/jooola/propel1"
+      "url": "https://github.com/libretime/propel1"
     },
     {
       "type": "vcs",
-      "url": "https://github.com/jooola/celery-php"
+      "url": "https://github.com/libretime/celery-php"
     }
   ],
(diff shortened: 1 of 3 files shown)


legacy/composer.json
```diff
@@ -16,20 +16,20 @@
   "repositories": [
     {
       "type": "vcs",
-      "url": "https://github.com/jooola/propel1"
+      "url": "https://github.com/libretime/propel1"
```
paddatrapper, line 19: Does this require a lock file update?

jooola: Yes it does. I have update the lock file as well.

PR #2362, opened 2023-02-03
fix(installer): only setup nginx on first install

### LINKED ISSUE #2338: Don't override configs on upgrade
### Describe the bug

Currently, a user could destroy there nginx configuration file by rerunning the installer. This can cause trouble once a user has configured certbot directly on libretime vhost, and the installer rewrite the configuration.

### To reproduce

1. Install LibreTime
2. Install and setup an ssl certificate using Certbot
3. Rerun the installer
4. Loose you certbot edited nginx configuration

### Expected behavior

We should either:
- document that certbot should use a dedicated vhost so that we can update the libretime nginx config on upgrade
- prevent upgrading the nginx config file unless explicitly requested.

_No response_

### LibreTime version

3.0.1

Irrelevant

### Installation details

_No response_

### Client Environment

_No response_

### Screenshots

_No response_

jooola:
Fixes #2338

@paddatrapper What do you think about this change ? Does it adds to much complexity ?

### DIFF
--- a/docs/admin-manual/setup/install.md
+++ b/docs/admin-manual/setup/install.md
@@ -150,6 +150,12 @@ When upgrading be sure to run the installer using the same arguments you used du
 
 :::
 
+:::warning
+
+To update the LibreTime nginx configuration file, for example to change the `--listen-port`, make sure to add the `--update-nginx` flag to allow overwriting the existing configuration file.
+
+:::
+
 If you need to change some configuration, the install script can be configured using flags or environment variables. Changing the listening port of LibreTime or whether you want to install some dependency by yourself, you could run the following:
 
 ```bash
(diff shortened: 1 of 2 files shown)


paddatrapper (approved):
I think this is a good idea and doesn't necessarily add too much complexity. We will need to add to the release notes if a change requires updating nginx configuration, but we need to do that anyway

jooola:
Here is an example when changing the --listen-port without --update-nginx:
image

And here is an example when adding --update-nginx and we update the file:
image

jooola:
> I think this is a good idea and doesn't necessarily add too much complexity. We will need to add to the release notes if a change requires updating nginx configuration, but we need to do that anyway

Cool, merging in that case.

PR #2363, opened 2023-02-03
feat: allow disabling cuepoint analysis per track type

### LINKED ISSUE #1818: Import causes Files to be shorter than they are
**Describe the bug**
I found that an Install of the newest Version (as of April. 27th 2022) causes uploaded Files to be shorter (time wise) than they really are.
I have Files from EXACTLY 1 hour and if i import it it gets down to around 57:49 thats a huge issue.
We need to have it play the files 1:1 as they come since theese are recorded Radio Shows.

Probably i should mention this is an Issue with Settings.
LT sets the Cue Out point way to early.

**To Reproduce**
Upload a MP3 with a known length

**Expected behavior**
I expect LibreTime to import my File as is.

**Installation method:**

- OS: Ubuntu
- Method: `./install` script

jooola:
Fix #1818

~~I am not happy with how I implemented this. But I guess I don't want to play too much with PHP...~~

The disable cuepoint_analyze option should go in track_type table as field, but I was a bit lazy going that route as used the preference table.

### DIFF
--- a/api/libretime_api/legacy/migrations/0044_add_track_types_analyzer_options.py
+++ b/api/libretime_api/legacy/migrations/0044_add_track_types_analyzer_options.py
@@ -1,0 +1,27 @@
+# pylint: disable=invalid-name
+
+from django.db import migrations
+
+from ._migrations import legacy_migration_factory
+
+UP = """
+alter table "cc_track_types" add column "analyze_cue_points" boolean default 't' not null;
+"""
+
+DOWN = """
+alter table "cc_track_types" drop column if exists "analyze_cue_points";
+"""
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("legacy", "0043_remove_cors_preference"),
+    ]
+    operations = [
+        migrations.RunPython(
+            code=legacy_migration_factory(
+                target="44",
+                sql=UP,
+            )
+        )
+    ]
--- a/api/libretime_api/legacy/migrations/0044_add_track_types_analyzer_options.py
+++ b/api/libretime_api/legacy/migrations/0044_add_track_types_analyzer_options.py
@@ -5,7 +5,9 @@ from django.db import migrations
 from ._migrations import legacy_migration_factory
 
 UP = """
-alter table "cc_track_types" add column "analyze_cue_points" boolean default 't' not null;
+alter table "cc_track_types" add column "analyze_cue_points" boolean default 'f' not null;
+
+update "cc_track_types" set "analyze_cue_points" = 't';
 """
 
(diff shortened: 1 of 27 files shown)


jooola:
I will re target this for 3.1.0

api/libretime_api/legacy/migrations/0044_add_track_types_analyzer_options.py
```diff
@@ -0,0 +1,27 @@
+# pylint: disable=invalid-name
+
+from django.db import migrations
+
+from ._migrations import legacy_migration_factory
+
+UP = """
+alter table "cc_track_types" add column "analyze_cue_points" boolean default 't' not null;
+"""
```
jooola, line 9: In the migration, the option is enabled by default, this is to prevent breaking with the previous behavior for existing track types, but it feels more 'normal' to have this disabled by default from now on ?

So we could set the column default to "disabled" but we enable all existing rows.

api/libretime_api/legacy/migrations/sql/schema.sql
```diff
@@ -95,6 +95,7 @@ CREATE TABLE "cc_track_types"
     "visibility" BOOLEAN DEFAULT 't' NOT NULL,
     "type_name" VARCHAR(64) DEFAULT '' NOT NULL,
     "description" VARCHAR(255) DEFAULT '' NOT NULL,
+    "analyze_cue_points" BOOLEAN DEFAULT 't' NOT NULL,
```
jooola, line 98: Haha I ended up doing this the right way ^^.

Are we ok with the naming ?

jooola (commented):
I think this is almost ready, I just have a few question with regards to naming and behavior.

api/libretime_api/legacy/migrations/0044_add_track_types_analyzer_options.py
paddatrapper: I wondered the same thing, but I think having it disabled by default is better. Especially if it doesn't really work anyway

jooola:
Oh you merged this a bit too fast

jooola:
The latest should have been squashed with the previous one.

paddatrapper:
No we won't. Just have to leave it as is. I read a PR that isn't in Draft state as ready to review and merge.

jooola:
> No we won't. Just have to leave it as is. I read a PR that isn't in Draft state as ready to review and merge. Should I rather wait for a 'Ready to merge' comment?

Ah ok, no I think it's been a while, I'll mark as ready when it's ready to merge, like you said, my bad.

Issue #2375, opened 2023-02-19
Setup a remote logging framework


jooola:
### Is your feature request related to a problem?

We previously removed all the code in PHP for sending exceptions to a remote server.

And we are now lacking those advanced monitoring/tracking capabilities. 

### Describe the solution you'd like

I propose to re-enable this feature but for python code, the sentry sdk works great with python and the django integration, this would allow devs to track exceptions using sentry or selfhosted/lightweight alternatives such a glitchtip.

One mistake I made is to force using loguru as logging framework, as sentry doesn't have any integration for this library and we would have to write custom code, I am unsure if we want this, I prefer not.

This would streamline the logging system with the django api with has to use the std logging library anyway. 

https://github.com/Delgan/loguru/issues/509
https://github.com/getsentry/sentry-python/issues/653#issuecomment-788854865

But the libretime-shared library does provide a nice level of abstraction that should allow us to do a smooth transition/integration.

PR #2377, opened 2023-02-19
refactor: don't use f-string on logging statements

jooola:
The strings are now be formatted if the logging level is enabled.

### DIFF
--- a/analyzer/libretime_analyzer/message_listener.py
+++ b/analyzer/libretime_analyzer/message_listener.py
@@ -98,7 +98,9 @@ class MessageListener:
         Here we parse the message, spin up an analyzer process, and report the
         metadata back to the Airtime web application (or report an error).
         """
-        logger.info(f" - Received '{body}' on routing_key '{method_frame.routing_key}'")
+        logger.info(
+            " - Received '%s' on routing_key '%s'", body, method_frame.routing_key
+        )
 
         audio_file_path = ""
         # final_file_path = ""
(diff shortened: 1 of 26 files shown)


jooola:
Whoups, not sure this PR is clever https://loguru.readthedocs.io/en/stable/resources/migration.html#replacing-style-formatting-of-messages

paddatrapper:
Looks like loguru supports f-strings no problem - https://loguru.readthedocs.io/en/stable/overview.html#modern-string-formatting-using-braces-style

jooola:
> Looks like loguru supports f-strings no problem - https://loguru.readthedocs.io/en/stable/overview.html#modern-string-formatting-using-braces-style

The goal was to move away from f-strings for logging statements. While we won't have a huge performance gain, it's a good practice to delay the formatting until the very last moment, so that the logger filter most of the logs by level for example.

All the debug logs with some huge dict/objects, will not have to be formatted unless we explicitly asked the debug log level.

So 
```py
logger.debug(f"hello {world}") # format the message, then send the log record
logger.debug("hello %s" % world) # format the message, then send the log record
logger.debug("hello %s", world) # send the log record, and print the message only if the level is correct 
```

In addition, I propose to add a remote logging facility such as sentry (#2375) and it currently does not support loguru yet, so I have the feeling we might want to move away from loguru.

I feel a bit dump about proposing this loguru in the first place Haha.

PR #2378, opened 2023-02-20
chore: add security policy

### LINKED ISSUE #1438: Handling security vulnerability reports
We need to decide on, and publish, how we handle security vulnerabilities.

The key things around that are:

1. Notification process - how do the maintainers get notified in a confidential way
2. Embargo length - how long after notification is the vulnerability allowed to be made public by the reporter
3. Release coordination - how do we distribute the fix

Point 1 I don't know how to address. Maybe we can setup a mail-forwarding address that we can then point people to?

On point 2, I think maybe 2-3 weeks (14-21 days). This should give us enough time to fix the issue, but not too long

For Point 3, currently we only support the latest release because LT is still in alpha. Thus, I suggest we cut a new release as the method of distributing the fix. Once 3.0 is out, we will probably need to define which versions we support and for how long, then make point releases for each.

jooola:
Fixes #1438

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,14 @@
+# Security Policy
+
+## Reporting a Vulnerability
+
+**Please do not use GitHub issues for security-sensitive communication.**
+
+The LibreTime maintainers ask that known and suspected vulnerabilities to be privately and responsibly disclosed by:
+
+- sending all the required detail to [security@libretime.org](security@libretime.org),
+- or by filling a [security advisory on Github](https://github.com/libretime/libretime/security/advisories/new).
+
+A LibreTime maintainer will acknowledged the report within 3 working days.
+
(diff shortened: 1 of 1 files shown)


paddatrapper:
Do we want to use https://docs.github.com/en/code-security/security-advisories/guidance-on-reporting-and-writing/privately-reporting-a-security-vulnerability?

jooola:
Yeah, sure. I have no idea how it works, but Github surely have a good documentation for it.

I would keep the security email address up though, and have 2 way of reaching out. If we receive an email we can always fill the security report in Github ourselves.

paddatrapper:
Yup, sounds good. I'll enable it for the repo and we can have a look at what the process is

jooola:
I included a link to open a new security advisory in the security policy. I think this is ready.

PR #2380, opened 2023-02-20
chore: release 3.0.2

jooola:
I would like to release another patch version.
I am only waiting for:
- #2379 
- #2381


jooola:
I've just installed LibreTime form the 3.0.x branch on both debian 11 and ubuntu 20.04 using the installer. Everything is working as expected.

Once the CI is passing I think we will be ready to merge.

jooola:
Great! let's go!

PR #2383, opened 2023-02-20
feat(playout): use liquidsoap interactive variables

jooola:
Please do not squash


paddatrapper (approved):
I haven't had a chance to test this in a dev instance, but overall looks good

jooola:
While working on the next iteration, I fixed the few bugs I encountered from this PR.

What would be awesome though, is to find a way to do unit testing for the liquidsoap library.

PR #2387, opened 2023-02-21
fix(installer): bump setuptools to ~=67.3

jooola:


jooola:
Well, let's keep it to ~=67.3 for now. This should prevent downgrading setuptools on the bookworm and jammy distro.

PR #2390, opened 2023-02-22
refactor(playout): remove unused code and improve code quality

jooola:
- rename some variables to be consistent with the config keys
- rename some non obvious variables names
- remove dead code (mainly related to the now unused "restart liquidsoap after settings change" feature)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,7 +8,11 @@
 *~
 VERSION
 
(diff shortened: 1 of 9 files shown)


jooola:
I tested this with normal tracks schedule, the main input harbor, a web stream, everything is working and streaming as expected.

PR #2395, opened 2023-02-24
fix(legacy): do not delete audio file when removing artwork

togir2:
### Description

While removing a custom artwork the audio file is also deleted.

**This is a new feature**:

No

No

### Testing Notes

This bug does only appear with custom cover images.
The default ones are not affected because they are in a different folder than the custom ones

1. Edit track
2. Add custom cover image and save
3. remove cover image => audio_file deleted

#### After the patch:

1. Edit track
2. Add custom cover image and save
3. remove cover image => audio_file is not deleted

### DIFF
--- a/legacy/application/common/FileDataHelper.php
+++ b/legacy/application/common/FileDataHelper.php
@@ -305,12 +305,18 @@ class FileDataHelper
 
         $fp = Config::getStoragePath();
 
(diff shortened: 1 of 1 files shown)


jooola:
Ooh, nice catch!

Now that I see this, the glob really scares me. I think we should be more explicit to what we delete here.

I've push an extra commit to replace the glob and fixed the CI checks.

Thanks for the patch!

PR #2400, opened 2023-02-26
docs: adapt c4 to our workflows

### LINKED ISSUE #2037: Setup backport tools to cherry pick fixes into version branches
Since we are going to maintain multiple versions of LibreTime, we will need to backport fixes to some of the version branches. This could be automated using back porting tools / bots.

Ideally, we could have a command that take a branch name as argument and opens a PR to backport the changes to this branch:

```
/backport 3.0.x
```

Some interesting tools: 
- https://github.com/sqren/backport

Here is how Nextcloud is handling it [https://github.com/nextcloud/server/pull/33495](https://togithub.com/nextcloud/server/pull/33495)

jooola:
Fixes #2037

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -6,9 +6,12 @@ LibreTime, your time is valuable, and your contributions mean a lot to us.
 Before any contribution, read and be prepared to adhere to our
 [code of conduct](https://github.com/libretime/organization/blob/main/CODE_OF_CONDUCT.md).
 
-In addition, LibreTime follow the standardized
-[C4 development process](https://rfc.zeromq.org/spec:42/c4/), in which you can
-find explanation about most of the development workflows for LibreTime.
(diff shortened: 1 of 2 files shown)


jooola:
From https://github.com/libretime/libretime/issues/2037#issuecomment-1442711330

> I am fine with a single release branch, in that case we need to change our supported versions policy.

I guess we are dropping the maintenance releases.

We could supersede the C4 requirement about the single release branch, but I have to admit I don't wanna loose time on "community rules" while there are so few contributions from the community. Unless you want to work on this I'll simply remove the maintenance branch and cherry pick the changes in 3.0.x that aren't in main yet.

> We use conventional commits and I don't want to use the format defined by C4. While we ask for a reason/problem before implementing a feature, we don't define everything as being a problem. And we will loose some of the tooling build around conventional commits.

I updated the Contributing document to reflect some of the LibreTime specific workflows.

PR #2402, opened 2023-02-26
fix(legacy): on air light fails when no shows are scheduled

jooola:
Fixes https://discourse.libretime.org/t/onairlightaction-errors-out-but-light-is-on-in-the-ui/1421


jooola:
Depends on #2403 so pre-commit doesn't fail.

paddatrapper (approved):

PR #2404, opened 2023-02-26
feat: add sentry sdk

### LINKED ISSUE #2375: Setup a remote logging framework

jooola:
Fixes #2375

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -76,7 +76,7 @@ repos:
       - id: requirements.txt
         name: requirements.txt
         description: Generate requirements.txt
-        entry: tools/extract_requirements.py dev
+        entry: tools/extract_requirements.py dev sentry
         pass_filenames: false
         language: script
         files: setup.py$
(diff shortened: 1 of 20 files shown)


jooola:
Cool, we have remote exception reporting!, This will be great to track exception we don't ever notice in the logs! 
image

paddatrapper:
Is there documentation around how to enable this?

PR #2405, opened 2023-02-26
chore: rename branch 3.0.x to stable

jooola:


jooola:
Hmm, I am still torn about this branch management problem. I am not sure it is smart to remove the ability to publish patch releases while having a development branch, just to please C4. Maybe we can supersede C4 for this particular point.

Maybe 3.0.x is a bad branch name, but rather have 'stable' as I described in https://github.com/libretime/libretime/issues/2037#issuecomment-1414096662

paddatrapper:
How do other projects handle this?

jooola:
> How do other projects handle this?

I like to compare our project to [funkwhale](https://dev.funkwhale.audio/funkwhale/funkwhale), they have a "stable" branch for patches, and a "develop" branch for major/minor changes. The stable branch in merged every once in a while inside the develop branch. The idea I described in my comment is pretty much the gist of what funkwhale is doing.

I see that other projects (e.g. [mattermost](https://github.com/mattermost/mattermost-server/branches/all?query=release-)) maintain multiple branches and cherry-pick changes when they have more than 2 versions to maintain (more than develop and stable), but this seem to work only if you have a large community/workforce, and have to maintain a lot of versions.

Looks like we initially started to do something similar to what mattermost does, but we could have used something similar to what funkwhale does.

paddatrapper:
Ok, sounds good. Let's go the way of Funkwhale then

jooola:
Further documentation and setup will be done in a future PR on the main branch.

paddatrapper (approved):

PR #2407, opened 2023-02-27
ci: pull code before updating locales

jooola:
Since we have a protection on the main branch the locales update could not push the changes to the main branch this night.

https://github.com/libretime/libretime/actions/runs/4278223004/jobs/7455414545

We should probably use the libretime-bot token to clone and push the changes.

The job succeeded on my fork, but I think this was not my intention haha. I limited the job to only run on the repo.

### DIFF
--- a/.github/workflows/analyzer.yml
+++ b/.github/workflows/analyzer.yml
@@ -1,6 +1,7 @@
 name: Analyzer
 
 on:
+  workflow_dispatch:
   push:
     branches: [main]
     paths:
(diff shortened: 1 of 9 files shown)


jooola:
So, I created a libretime-bot token with read/write access to public repositories, configured it in the repository secrets, and we should not be able to push to the main branch.

If the branch protection is still blocking libretime-bot from pushing, I'll add an exception if the CI fails after this is merged.

Issue #2070, opened 2022-08-24
When shows ends, next shows starts without fade-in/fade-out


esbobkov:
**Describe the bug**
![image](https://user-images.githubusercontent.com/5690261/186404133-a1904d69-4877-49e6-9a03-17a648d69651.png)

I have this settings and my shows ends without fade out. How to fix it?
Shows programmed with autoloading playlists with dynamic SmartBlocks. "Allow last track to exceed time limit” enabled.

I don’t hear any difference when change “Default Fade Out” time

**To Reproduce**

1. Setup schedule and shows with autoloading playlists with dynamic SmartBlocks
2. Listen
3. No fade-out track in prev. show, when next show starts

**Expected behavior**
Hear fade-out track in prev. show, when next show starts

**LibreTime version:**
3.0.0-alpha.12, 3.0.0-alpha.13
liquidsoap 1.4.2, liquidsoap 1.4.4

**Installation method:**

- OS: Ubuntu
- OS Version: 20.04.4 Focal Fossa
- Method: `./install` script

**Additional context**
logs
```
2022/08/10 06:00:00 [server:3] New client: 127.0.0.1.
2022/08/10 06:00:00 [lang:3] vars.show_name
2022/08/10 06:00:00 [server:3] Client 127.0.0.1 disconnected.
2022/08/10 06:00:01 [cue_cut_9659:3] Cueing out...
2022/08/10 06:00:01 [cross_9665:3] Analysis: -17.615671dB / -nandB (0.00s / 0.00s)
2022/08/10 06:00:01 [crossfade:3] No transition: using default.
2022/08/10 06:00:01 [schedule_noise_switch:3] Switch to map_metadata_9743 with forgetful transition.
2022/08/10 06:00:01 [lang:3] transition called...
2022/08/10 06:00:01 [decoder:3] Method "FFMPEG" accepted "/var/lib/libretime/playout/scheduler/30.mp3".
2022/08/10 06:00:01 [dummy(dot)2:3] Source failed (no more tracks) stopping output...
2022/08/10 06:00:01 [dummy(dot)4:3] Source failed (no more tracks) stopping output...
2022/08/10 06:00:01 [s4:3] Prepared "/var/lib/libretime/playout/scheduler/30.mp3" (RID 0).
2022/08/10 06:00:01 [schedule_noise_switch:3] Switch to map_metadata_9732 with transition.
2022/08/10 06:00:01 [lang:3] transition called...
[...]

PR #2412, opened 2023-02-28
fix(playout):  when shows ends, next shows starts without fade-in/fade-out 

togir2:
### Description
Liquidsoap did a harsh cut at the end of every show.
This MR fixes this an (re)enables settings for the crossfading by
- adding the values "default_crossfade_duration", "default_fade_in" and "default_fade_out" from the "preference" page to the StreamPreferences api.
- The new values are then used while generating the liquidsoap entrypoint, and can also be changed  via telnet.

### Testing Notes

Created a VM, installed my Libretime branch, scheduled some shows to hear the fade at the end of the show. 
I also run "make test" in the "playout" folder.

### **Links**

Closes: #2070 

https://discourse.libretime.org/t/how-to-fade-out-shows/700
https://discourse.libretime.org/t/no-fadestop-at-the-show-end/1412

### DIFF
--- a/playout/libretime_playout/liquidsoap/1.4/ls_script.liq
+++ b/playout/libretime_playout/liquidsoap/1.4/ls_script.liq
@@ -23,8 +23,11 @@ def create_source()
     l = cue_cut(l)
     l = amplify(1., override="replay_gain", l)
 
-    # the crossfade function controls fade in/out
-    l = crossfade(duration=0., smart=true, l)
+    # Fade the tracks to avoid hard cuts in between two tracks and at the end of the shows.
+    # Liquidsoap reads the fade in/out durations from the annotation "liq_fade_in/out" which
+    # value can be set via Libretime settings.
+    l = fade.in(l)
+    l = fade.out(l)
 
     l = on_metadata(notify_queue, l)
 


jooola:
Hey, it is amazing that you spend some time on fixing this! Thanks a lot! Ping me when this is ready for a proper review.

I had a few small question though, did you manage to get it fixed ? Or are you still searching ?

Did you consider that liquidsoap is supposed to read the fade in/out from the play request annotations ?
https://github.com/libretime/libretime/blob/32e0c2a15e355e92a46c5005d68e5b272ca37480/playout/libretime_playout/player/liquidsoap_gateway.py#L11-L40

https://www.liquidsoap.info/doc-1.4.4/reference.html#fade.in
https://www.liquidsoap.info/doc-1.4.4/reference.html#fade.out
https://www.liquidsoap.info/doc-1.4.4/reference.html#cue_cut

togir2:
> I had a few small question though, did you manage to get it fixed ? Or are you still searching ?
In my test it did work after applying the patch. Is it not working for you?

> Did you consider that liquidsoap is supposed to read the fade in/out from the play request annotations ?
Thanks, I didn't saw that. I will remove the "default_fade_in" and "default_fade_out".

jooola:
> In my test it did work after applying the patch. Is it not working for you?

I didn't test yet, I was just curious about the reason why it was broken and why your changes fixed it.

> Thanks, I didn't saw that. I will remove the "default_fade_in" and "default_fade_out".

Hmm, I like your changes about this default_fade_in, default fade out, could you keep them around ? We might want to keep them. But I have to think a bit more about whether it is a good idea.

togir2:
> I didn't test yet, I was just curious about the reason why it was broken and why your changes fixed it.

I do think it was broken because the "duration" for the crossfade was set to 0. I assume the "fade_in" and "fade_out" are  supposed to happen during the crossfade "duration".
Also the "smart=true" option does in my observation sometimes decide, a cut is better than a fade.

https://github.com/libretime/libretime/blob/32e0c2a15e355e92a46c5005d68e5b272ca37480/playout/libretime_playout/liquidsoap/1.4/ls_script.liq#L25-L28


If you want to keep the "default_fade_in" and "default_fade_out" then I would say this PR is ready from my side :)

jooola:
> I do think it was broken because the "duration" for the crossfade was set to 0. I assume the "fade_in" and "fade_out" are supposed to happen during the crossfade "duration".
Also the "smart=true" option does in my observation sometimes decide, a cut is better than a fade.

The fade in/out does not happen on the same queue, this is why there are multiple queues `create_source`. With that in mind, I am unsure what the crossfade exactly does in the create_source context. But I think its not the only one at play in the transition system.

Also, the fade between shows seemed to be working before (maybe in around the 3.0.0-alpha.* release) and I probably broken something, or a newer version of liquidsoap broke it.

I am still trying to understand why we have multiple request queues. This part of the liquidsoap script was written 7/10 years ago, and I think liquidsoap didn't have all the feature for fading it had back then. So a lot of the timing management happens on the python side.

playout/libretime_playout/liquidsoap/1.4/ls_script.liq
```diff
@@ -23,8 +23,9 @@ def create_source()
     l = cue_cut(l)
     l = amplify(1., override="replay_gain", l)
 
-    # the crossfade function controls fade in/out
-    l = crossfade(duration=0., smart=true, l)
+    # The crossfade function controls fade in/out of Tracks and shows
+    # Default values are set in the entrypoint.liq / entrypoint.liq.j2 template
+    l = crossfade(duration=default_crossfade_duration(), fade_in=default_fade_in(), fade_out=default_fade_out(), smart=false, l)
```
jooola, line 28: I tested the PR, and it is not working for me. Also the fade in/out is also broken between tracks inside the same show.

I think the crossfade here happen between the tracks in the queue, but (as explained in the comment before), we use multiple queues, so maybe the crossfade duration here is irrelevant, but the fade in / fade out are.

playout/libretime_playout/liquidsoap/client/_client.py
```diff
@@ -139,3 +142,11 @@ def settings_update(
                 self._set_var("message_offline", self._quote(message_offline))
             if input_fade_transition is not None:
                 self._set_var("input_fade_transition", input_fade_transition)
+            if default_crossfade_duration is not None:
+                self._set_var(
+                    "default_crossfade_transition", default_crossfade_duration
+                )
```
jooola, line 148: This is failing with:

```
libretime-playout-1     | 2023-02-28 15:11:07,895 | ERROR    | libretime_playout.liquidsoap.client._client:_set_var:48 - Variable default_crossfade_transition is not defined.
```

playout/libretime_playout/player/fetch.py
```diff
@@ -138,6 +138,9 @@ def set_bootstrap_variables(self):
                 message_format=preferences.message_format,
                 message_offline=preferences.message_offline,
                 input_fade_transition=preferences.input_fade_transition,
+                default_crossfade_duration=preferences.default_crossfade_duration,
+                default_fade_in=preferences.default_fade_in,
+                default_fade_out=preferences.default_fade_out,
```
jooola, line 143: This is not triggered when we change the values from the general preferences panel in the UI. So it only set during boot time.

Currently the only place where the liquidsoap interactive settings get updated is when the stream settings in the UI is updated.

While we might want to move the fade settings inside the stream settings panel, those values are only used while building the schedule and are never passed directly to liquidsoap in the first place. We might want to wait that the fix really works before moving too much code to handle this particular issue.

playout/libretime_playout/liquidsoap/client/_client.py
togir2: Thanks, I missed that. Initially I had the variables all named "default_crossfade_transition" but then changed it that is is consistent with the name in the DB and the Webpage

togir2: Thanks, I missed that. 
Initially I had the variable named "default_crossfade_transition" but then changed it that is is consistent with the name in the DB and the Webpage.

playout/libretime_playout/liquidsoap/1.4/ls_script.liq
togir2: Mhh, I just rested it with a fresh install on Debian 11 and for me it is still working.
Did you change the default values? At my installation the default was "crossfade_duration: 0", "fade_in: 0.5", "fade_out:0.5". With the defaults I did also not here a fading.

When I change them to "crossfade_duration: 3", "fade_in: 1", "fade_out: 1". 
When I have changed them I can hear them fade well.

You might need to restart the server between changing an listening because, as you said in your other comment, there is no trigger.

playout/libretime_playout/player/fetch.py
togir2: I've found the trigger for the stream settings. I could add it to the "general settings" too.
Are you planing to move away form the php-framework? If so I could also try move the settings to Django and start the trigger there.

jooola: Yes, we are moving away from the php framework.

The current code to edit the setting is still in the legacy app, so lets keep it in legacy for now.

togir2:
I did some more experiments and found that replacing the crossfade function with fade.in and fade.out did get the tracks to fade in/out. Could you please test it again? I tested it on a Debian11 VM with a fresh install of this branch and the fade die work for tracks in shows as for the end of shows.

This is why I think the crossfade operator does not working here:
---
In Liquidsoap version 1.4.3. [crossfade](https://www.liquidsoap.info/doc-1.4.3/reference.html#crossfade) is implemented as a [cross](https://www.liquidsoap.info/doc-1.4.3/reference.html#cross) with a custom transition (fade_in and fade_out).
https://github.com/savonet/liquidsoap/blob/9f730f2c5f6195e37a761f7b1eb5a74395d22b18/src/libs/fades.liq#L433-L436

The "duration" argument is passed through to the cross function.
In the implementation of the cross operator the value duration is used to determine how log the crossfade should take.

It is set to the cross_lenght parameter

https://github.com/savonet/liquidsoap/blob/f075905715584bdc236fe2078e26352a5b7f1c5f/src/operators/cross.ml#L30-L34

This can be overwritten with metadata, but the current annotation does not include a "override_duration" field so in our case it is always 0.
https://github.com/savonet/liquidsoap/blob/f075905715584bdc236fe2078e26352a5b7f1c5f/src/operators/cross.ml#L186-L198

So I assume the crossfade is starting to fade.out the track but because the duration is set to 0. the "cross" is completed immediately and the next source of the queue is started. Our queues do only ever contain one track at a time so there is no next source to play.
The next queue is activated and the same happens for the fade.in.

Replacing the crossfade with a fade.in/out removes this time boundary as there is no longer a "cross" function involved.

Until the tag 3.0.0-alpha.8 there was a custom [crossfade_airtime]( https://github.com/libretime/libretime/blob/e232469551ee3fcc199818d73ad60eedc8824e7b/python_apps/pypo/liquidsoap/ls_script.liq#LL76C9-L76C26)  function. 
In tag 3.0.0-alpha.9 it was replaced with the crossfade function but was unable to find why.
https://github.com/libretime/libretime/blob/ecd302068c443af3eaa44bd3f81f45cdd88b2155/python_apps/pypo/liquidsoap/1.4/ls_script.liq#LL76C9-L76C18

The crossfade_airtime function is basicly a reimplementation of the crossfade function, except that cross has no duration set, so it will use the default value (5.0).
```
def crossfade_airtime(s)
  #duration is automatically overwritten by metadata fields passed in
  #with audio
  s = fade.in(type="log", duration=0., s)
  s = fade.out(type="log", duration=0., s)
  fader = fun (a,b) -> add(normalize=false,[b,a])
  cross(fader,s)
end
```

---

Alternatively we could also extend the annotation to include the "liq_cross_duration", but as we are not crossfading in any case I do like it more to use the the explicit fade.out and fade.in functions. Maybe the "Crossfade duration" form from the settings page could be removed as it does not have any affect on the script.

jooola:
Nice digging! I think what you wrote above should be in the commit body, it is a gold mine for future miners (I'll add it while merging).

I've tested the PR and it is working! We might be loosing some of the "smart" from the crossfade, but since we were using it wrong, I would keep your fix as it is, and rework the whole schedule queues system it in the future.

The "Crossfade duration" is used, we must keep it. I think the legacy app is using it compute the cue_in/cue_out values which in the end has an impact on the crossfade/fade duration: https://github.com/libretime/libretime/blob/32e0c2a15e355e92a46c5005d68e5b272ca37480/legacy/application/models/Scheduler.php#L485

The problem is that the streaming code is spread between legacy (php code), the new api, the playout app and liquidsoap, and it's difficult to work with. I'm slowly working on improving this, but it takes time...

I'll do some more testing, and let this PR sit for the day.

jooola:
The only missing part is that we don't have any crossing between shows, only fades. But that's unrelated and for another ticket.

Thanks a lot for the patch.

esbobkov:
Thanks for The fix!

@jooola Hello!
When do you plan new release with this fix?

jooola:
> When do you plan new release with this fix?

There is no ETA yet, it should be released in 3.1.0. I want to ship some important changes in 3.1.0 so this will probably need some more time.

PR #2419, opened 2023-03-03
feat: improve docker-compose spec

jooola:


docker-compose.yml
```diff
@@ -2,21 +2,21 @@ version: "3.9"
...
       test: pg_isready -U libretime
 
   rabbitmq:
-    image: rabbitmq:alpine
+    image: rabbitmq:3.11-alpine
     environment:
-      RABBITMQ_DEFAULT_VHOST: /libretime
-      RABBITMQ_DEFAULT_USER: libretime
-      RABBITMQ_DEFAULT_PASS: libretime # Change me !
+      RABBITMQ_DEFAULT_VHOST: ${RABBITMQ_DEFAULT_VHOST:-/libretime}
```
paddatrapper, line 17: Maybe document these in the docker setup docs?

jooola: Done, hope this is enough. The docker-setup is a bit more complex than the install script, I am unsure if we want to make it easier by scripting it. Maybe some bash command in the docs is enough.

jooola: Once legacy is removed or support env vars instead of the configuration file, we will be able to make this simpler and remove the duplication between the env file and the config file.

PR #2441, opened 2023-03-15
docs: add instructions for the sentry setup

jooola:
We are lacking an easy way to install the sentry extra using the installer, but since this is for power users I assume they know how to handle this.

I'll include an option to install the sentry extra in the ansible role.

### DIFF
--- a/docs/admin-manual/monitoring.md
+++ b/docs/admin-manual/monitoring.md
@@ -1,0 +1,26 @@
+---
+title: Monitoring
+sidebar_position: 85
+---
+
+This page provide some guidance to monitor LibreTime.
+
+## Sentry
+
+To gather and remotely monitor exceptions that may occur in your installation, you can use the Sentry library shipped in LibreTime to send reports to any Sentry compatible server ([Sentry](https://sentry.io/), [Glitchtip](https://glitchtip.com/)).
+
+To configure Sentry in LibreTime, you need to:
+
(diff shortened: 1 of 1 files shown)


docs/admin-manual/monitoring.md
```diff
@@ -0,0 +1,17 @@
...
+
+This page provide some guidance to monitor LibreTime.
+
+## Sentry
+
+To gather and remotely monitor exceptions that may occur in your installation, you can use the Sentry library shipped in LibreTime to send reports to any Sentry compatible server.
+
+To configure Sentry in LibreTime, you need to:
+
+- install the LibreTime Python packages with the `sentry` extra (the container images already ship the sentry extra),
```
paddatrapper, line 14: Maybe provide an example here?

docs/admin-manual/monitoring.md
```diff
@@ -0,0 +1,17 @@
...
+- set the `SENTRY_DSN` environment variable on each of the LibreTime services you want to monitor.
```
paddatrapper, line 15: Can you elaborate on what is expected to be in this environment variable?

jooola: I added a like to the Sentry [Data Source Name](https://docs.sentry.io/product/sentry-basics/dsn-explainer/) docs.

jooola:
image

PR #2442, opened 2023-03-15
ci: enable python tests on bookworm

jooola:


jooola:
We need to upgrade celery to >=5.0.0 in order to support bookworm. This means dropping celery-php and move the entire podcast business logic to the django api.

PR #2474, opened 2023-03-21
feat(api-client): rewrite api-client v1 using abstract client

jooola:
Cleans the api client v1 from unused function, and rewrite the client using the abstract client.

- The retry logic has been moved to a custom decorator for the few call that needs it.
- Use query params instead of path params
- Move the api key path parameter to a Api-Key header.


paddatrapper:
I assume the PHP API supports query-based parameters and the Api-Key header?

PR #2480, opened 2023-03-23
chore: backport #2461

jooola:
fix(api): cast string value to int enum (#2461)

c7381a4f809114708838f86aead0d9c769bc3b79


jooola:
Merging as stable is currently broken

PR #2489, opened 2023-03-30
feat(playout): allow secure input harbor configuration

### LINKED ISSUE #1032: WebCaster support requires SSL liquidsoap harbour

jooola:
Fix #1032

### DIFF
--- a/dev/config.yml
+++ b/dev/config.yml
@@ -15,7 +15,17 @@ playout:
 liquidsoap:
   server_listen_address: 0.0.0.0
 
+  harbor_ssl_certificate: /certs/fake.crt
+  harbor_ssl_private_key: /certs/fake.key
+
 stream:
+  inputs:
+    main:
+      public_url: https://localhost:8001/main
+      mount: main
+      port: 8001
+      secure: true
+
   outputs:
     .default_icecast_output: &default_icecast_output
       host: icecast
--- a/docker-compose.override.yml
+++ b/docker-compose.override.yml
@@ -32,6 +32,7 @@ services:
     volumes:
       - ./playout:/src
       - ./dev/playout:/app
+      - ./dev/certs:/certs
       ## See https://libretime.org/docs/admin-manual/tutorials/setup-a-pulseaudio-output-inside-containers/
       # - ./dev/pulse.socket:/tmp/pulse.socket
       # - ./docker/pulse.client.conf:/etc/pulse/client.conf
--- a/docker/config.template.yml
+++ b/docker/config.template.yml
@@ -107,6 +107,13 @@ liquidsoap:
   # > default is ["0.0.0.0"]
   harbor_listen_address: ["0.0.0.0"]
 
+  # Input harbor tls certificate path.
+  harbor_ssl_certificate:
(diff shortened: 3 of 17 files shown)


dev/config.yml
```diff
@@ -15,7 +15,17 @@ playout:
...
+  harbor_ssl_certificate: /certs/fake.crt
+  harbor_ssl_private_key: /certs/fake.key
+
 stream:
+  inputs:
+    main:
+      public_url: https://localhost:8001/main
+      mount: main
+      port: 8001
+      secure: true
```
jooola, line 27: According to #1032 we talked about adding a `secure_port` field to be able have both a secure and insecure port, but this requires to add an extra input harbor and require more handling, for instance how does the input behave if both input port are used at the same time.

Only allowing for a single input makes things simpler for the time being, and maybe we want to allow a list of inputs in the future to allow more advanced use cases.

So this adds a simple way to secure the input stream, until we draft a proper spec on how we want to handle more than 1 input stream for both main and show inputs.

My only question is whether the secure field is future proof, and if we will have to change it again when adding extra features for the input streams ?

Makefile
```diff
@@ -10,7 +10,18 @@ setup:
...
+certs:
+	rm -f dev/certs/fake.*
+	openssl req -x509 \
+		-newkey rsa:2048 \
+		-days 365 \
+		-nodes \
+		-subj "/CN=localhost" -addext "subjectAltName=DNS:localhost,IP:127.0.0.1" \
+		-keyout dev/certs/fake.key \
+		-out dev/certs/fake.crt
+	cat dev/certs/fake.{key,crt} > dev/certs/fake.pem
```
jooola, line 22: Let's always use a secure setup during development, for now only the liquidsoap inputs and the icecast stream are secure, but maybe we can also setup nginx to load the certificates.

docker-compose.override.yml
```diff
@@ -31,9 +31,10 @@ services:
       - 1234:1234
     volumes:
       - ./playout:/src
-      - ./docker/data/playout:/app
+      - ./dev/playout:/app
+      - ./dev/certs:/certs
```
jooola, line 35: We started packing everything also dev related in the docker directory, I felt we had to put a clear separation for what the dev files are and what the docker resources are.

jooola:
This is lacking documentation, but I think it's ready to be reviewed.

Maybe I manage to finish #2339 before this so I have a base to document "how to secure liquidsoap ports".

dev/config.yml
paddatrapper: I don't think we'll have to change it again. I agree that a single input is much easier, and makes it a binary choice - are you using SSL for your LT deployment or not. We could change the result of the flag if needed later, but I can't foresee needing to do that at the moment

PR #2495, opened 2023-04-03
chore: release 3.1.0

jooola:
I'd like to start thinking about releasing 3.1.0.

There are some issues still open on the 3.1.0 milestone, I've moved some of them to the 3.2.0 milestone. Is there anythingelse that must go into 3.1.0 ?

### DIFF
--- a/analyzer/setup.py
+++ b/analyzer/setup.py
@@ -2,7 +2,7 @@ from setuptools import find_packages, setup
 
 setup(
     name="libretime-analyzer",
-    version="3.0.2",
+    version="3.1.0",
     description="Libretime Analyzer",
     author="LibreTime Contributors",
     url="https://github.com/libretime/libretime",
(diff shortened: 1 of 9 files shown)


paddatrapper:
This is blocked by #2498 I guess?

jooola:
I was hoping to close the entire milestone, but maybe that's too much and we need to release sooner ?

paddatrapper:
I think we can hold off to complete the milestone

jooola:
Once #2567 is merged, I'll be able to prepare the 3.1.0 release, test it and maybe even publish it today!

jooola:
So, I've tested a deployment using vagrant, as well as an upgrade from stable to the current main. Everything seem to be working fine.

I've been running from the main branch in production for the different PR's that needed testing. So I think the overall state of main is stable.

I think we could release today, if there is no other blockers, @paddatrapper what do you think ?

PR #2518, opened 2023-04-24
chore(api): install drf from git repository

jooola:
Temporarily install django-rest-framework from the git repository. This allows us to test LibreTime before we release 3.1, and test Django 4.2 + django-rest-framework before it is released. We will revert this and upgrade django-rest-framework before we release 3.1.


jooola:
I'd like to send the current main branch into staging/production envs, so I can do some testing before we release 3.1.

PR #2521, opened 2023-04-24
feat: install libretime inside a venv

### LINKED ISSUE #2507: Handle PEP-668 and install LibreTime in a venv
### Describe the bug

The introduction of https://peps.python.org/pep-0668/ in Debian Bookworm, will break our current install procedure.

This PEP will prevent the installation of pip packages globally, which we heavily rely on.

Since we will not rely on the distribution python packages, the solution is to install LibreTime inside a virtual env, either using pipx or manually.

I am unsure what is best for our use case, using venv seem more flexible, but pipx removes a decent amount of boilerplate we will have to write.

While we only have to implement this to support debian bookworm, I think we can extend this setup to all the distributions we support.

Some old discussion somewhat related about this https://github.com/libretime/libretime/issues/1207

cc @paddatrapper

### To reproduce

1. Install LibreTime on debian bookworm
2. See the pip install error

### Expected behavior

The LibreTime installer should be able to install the apps dependencies and packages.

```shell
error: externally-managed-environment

× This environment is externally managed
╰─> To install Python packages system-wide, try apt install
    python3-xyz, where xyz is the package you are trying to
    install.
    
    If you wish to install a non-Debian-packaged Python package,
    create a virtual environment using python3 -m venv path/to/venv.
    Then use path/to/venv/bin/python and path/to/venv/bin/pip. Make
    sure you have python3-full installed.
    
    If you wish to install a non-Debian packaged Python application,
    it may be easiest to use pipx install xyz, which will manage a
    virtual environment for you. Make sure you have pipx installed.
    
    See /usr/share/doc/python3.11/README.venv for more information.

note: If you believe this is a mistake, please contact your Python installation or OS distribution provider. You can override this, at the risk of breaking your Python installation or OS, by passing --break-system-packages.
hint: See PEP 668 for the detailed specification.
```

### LibreTime version

3.0.2

Operating system: >= Debian Bookworm
Method: Ansible, installer, maybe docker

### Installation details

_No response_

### Client Environment

Operating system:
Browser:

### Screenshots

_No response_

jooola:
Fixes #2507

### DIFF
--- a/install
+++ b/install
@@ -177,12 +177,10 @@ while [[ $# -gt 0 ]]; do
 done
 
 # Variables
-PYTHON="python3"
-PIP="$PYTHON -m pip"
-
 DEFAULT_WEB_USER="www-data"
 
 # Paths
+VENV_DIR="/opt/libretime"
 CONFIG_DIR="/etc/libretime"
 CONFIG_FILEPATH="$CONFIG_DIR/config.yml"
 CONFIG_TMP_FILEPATH="$CONFIG_DIR/config.yml.tmp"
@@ -276,10 +274,12 @@ set_config() {
 # install_python_app <path>
 install_python_app()  {
   info "installing python app from $1"
-  $PIP install \
-    --upgrade \
-    --upgrade-strategy only-if-needed \
-    "$1"
+  $VENV_DIR/bin/pip install --upgrade "$1"
+}
+
+# link_python_app <name>
+link_python_app() {
+  ln -s -f "$VENV_DIR/bin/$1" "/usr/local/bin/$1"
 }
 
 # install_service <name> <filepath>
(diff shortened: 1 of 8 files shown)


install
```diff
@@ -177,12 +177,10 @@ while [[ $# -gt 0 ]]; do
 done
 
 # Variables
-PYTHON="python3"
-PIP="$PYTHON -m pip"
-
 DEFAULT_WEB_USER="www-data"
 
 # Paths
+INSTALL_DIR="/opt/libretime"
```
jooola, line 183: `/opt/libretime` seem like a good path for this, but maybe it doesn't comply enough with FHS ? What would be the other alternatives ?

paddatrapper: `/op/` seems to be a fairly standard place for software. I agree it isn't strictly FHS, but it's common enough to be a standard

Issue #1788, opened 2022-04-22
Shared session between Django and Zend


jooola:
In the transition to Django, I think we should setup a shared session store in the db (or in a key/value store). I am unsure how feasible this is on the PHP side, but this should simplify the authentication management.

The future Webapp will need to call the Django API, and we will probably reuse the same shared session.

Ideally we would have a single place with Write access to the store, and the rest will only have a read only permissions to the store to authenticate a request.

togir2:
I took a look at the php-code and I am afraid of changing the session store...
I do have a proposal with which we can login via Django and then create two sessions, one for Django and one for php.

1. Login to Django with Username/ Password
2. Django creates its session and an extra token for the user stored in a  cookie
  The cookie should look like this:
 ```
   message = {"db_userid": <id>, "valid_until":"<timestring>"}
   cookie = {"message" : message, "hmac": hmac(mesage, "libretime_secret_key")}
   valid_until should be short timespan
```

3. Django redirects to the legacy login
4. Custom authAdapter reads cookie-token, check if hmac is valid.
5. If valid: create php session for the user and delete token from cookie

The user is now logged in with his Django "session_cookie" and the PHP_SESSID

Advantages:
- "Easy" to implement
- We do not need an extra database
- The session handling of the legacy php is not changed
- We do need only an addition to the Django authentication flow
- The Django login and session management is not changed
- It is possible to update the password in the DB to use the Django hashfunctions (no more md5!)
- In the future  our additions can be removed and  the "stock" Django authentication is back.

Disadvantages:
- On logout Django must delete it's own session cookie + the PHP_SESSID
- Session in legacy does not get deleted on logout, only when expired (allows for longer bruteforce time)
- Deletion of only one session_cookie (e.g. session timeout) will lead to the user logged in at either Django or legacy
- When the cookie gets stolen we are "duration from now until valid_until" vulnerable to replay attacks, but if it is possible that cookies are stolen the session cookies are also not save.

jooola:
Yes, maybe not shared the session store is cleaner.

Have you considered adding some internal endpoints to legacy, and rather than redirecting the user, we make a few calls from the api to legacy to maintain the php sessions ?

After valid login in the api, we forward the request to legacy, create a session and we respond to the user with both api and legacy data ?

togir2:
> After valid login in the api, we forward the request to legacy, create a session and we respond to the user with both api and legacy data ?

This does sound fine and also not hard to implement.
[...]

PR #2523, opened 2023-04-24
feat(legacy): move session store to database

jooola:
Related to #1788


jooola:
This PR is a preparation work for #2572

paddatrapper (approved):
Looks good. I don't really understand PHP though, so I won't be able to give a deep review of the actual code

PR #1696, opened 2022-03-26
feat: adding new UI scaffolding

zklosko:
### Description

With the official release of Vuetify 3 Beta, I'm restarting development of our new UI with Vue 3, the Composition API, and Vite. Typescript, Vue Router, Cypress, Vitest, and Storybook are also included. Most of it is boilerplate, and I'm hoping for assistance, same as before.

This code does not interact with the existing PHP UI, and is not copied to production folders during our install script.

The proposed URL nesting structure is as follows:

```
/ -- Microsite or splash page
/dashboard -- Base dashboard page template, requires login
   /calendar -- Dashboard UI pages, same as with the PHP UI
   /help -- etc...
/widgets -- delivery of widgets for embedding elsewhere
   /weeklyschedule -- weekly schedule widget
   /nowplaying -- etc...
```

Todo:

- [ ] Add documentation explaining how to develop the new UI
- [ ] Integrate Vue i18n
- [ ] Automate linting with Github Actions
- [x] Remove Storybook.js
- [x] Complete skeleton routing with Vue Router

### Testing Notes

**What I did:**

This was created with

```
npm init vue@latest
```

**How you can replicate my testing:**

Enter the `ui` directory and run

```
yarn install
yarn dev
```

PR #2546, opened 2023-05-12
feat: init workspace for vue webapp

### LINKED ISSUE #1121: Proposal: Initialize Vue Env

zklosko:
### Description

Initially started in #1696, starting again from scratch following feedback from lead maintainers.

This PR creates a development environment for the rewrite of Libretime's web user interface. The environment is based on Node 18.12.0 LTS and includes Vue 3, Vuetify 3, Vite, Cypress, Storybook, vue-i18n, Eslint, and Typescript.

No documentation updates for now. Getting started notes are in Readme.MD.

### Todo

- [x]  Install packages for development (Vue, Vite, Vuetify, Cypress, etc.)
- [x]  Initialize vue-i18n: create templates, config files
- [x]  Configure linting and formatting tools
- [x] Set up Storybook
- [x] Set up and create basic testing with Cypress (e2e and unit)
- [x] Set up CI/CD with Github Actions
- [x] Copy over logo, image assets from legacy app
- [x] Add more dev instructions to the Readme file

### **Links**

Closes #1121


jooola:
Hey @zklosko, is there more work you want to put into this PR ? Have you come up with a good way to embed the Vue multipage views into the legacy app ?

I also have a branch with a POC that embeds some Vue pages into the legacy app, but it didn't feel clean enough to propose a PR yet. I am currently focusing the on the new auth system, which is required for the new UI to work.

What is your plan in regard to this PR and with more generally the web app ?

zklosko:
Hi @jooola! My goal is to set up a base dev environment with configured plugins and CI so we have a good foundation to start building the new UI.

I haven't thought about embedding into the new app. Was planning to do the widgets first since they can be served separately. Once we have the building blocks, I would want to start redesigning from the ground up. It doesn't make sense to move from Bootstrap to Material design frameworks and then endlessly mold the CSS so it looks just like the old UI. I would want to make a new UI and then release it all together as Libretime 4.0.

maxtimbo:
I've been getting very friendly with Vue 3 lately. How can I help?

zklosko:
@maxtimbo I added some todos to the PR's description. Just focusing on building out the development workspace with all the basics; we'll start building individual components next.

zklosko:
One issue that I'm running into is that many i18n plugins that work with Vue don't work with .po files. I've been able to convert the po files to json for i18next for dev purposes. We will need to adjust Weblate accordingly.

jooola:
@zklosko I hate to cut your energy every time you work on this, but this is a large change and some planning is required. You cannot implement or swap a technology for another without having a small discussion about it.

I won't merge something that hasn't been agreed on, as we will be the one maintaining it.

From what I recall, I think we agreed on the Vuetify framework, you cannot swap this with bootstrap.

You are currently going back and forth about the i18n framework, this should be planned ahead instead of putting too much work in something that will not be agreed on.

Anyway, if that PR is meant to be an experiment, please continue, I appreciate the effort.

zklosko:
We did agree on Vuetify, that's what I'm using. I said it isn't practical to make Vuetify look exactly like Bootstrap 3, so we need to redesign the UI from the ground up.

Vue-i18n doesn't support .po translation files, so we'll need to completely redo our translation pipeline to use it, which is not ideal. I'm trying to find a way so we don't have to work with two translation systems during development. I'll open an issue to discuss it.

Yes, everything WIP.

webapp/.editorconfig
```diff
@@ -0,0 +1,5 @@
+[*.{js,jsx,ts,tsx,vue}]
+indent_style = space
+indent_size = 4
+trim_trailing_whitespace = true
+insert_final_newline = true
```
jooola, line 5: Do we really need this file? I would assume IDE will pick up the prettier config instead of this one.

webapp/package.json
```diff
@@ -0,0 +1,55 @@
...
+    "@typescript-eslint/parser": "^5.59.6",
+    "@vitejs/plugin-vue": "^3.2.0",
+    "@vue/eslint-config-typescript": "^11.0.0",
+    "cypress": "^12.12.0",
+    "eslint": "^8.0.0",
+    "eslint-plugin-storybook": "^0.6.12",
+    "eslint-plugin-vue": "^9.13.0",
+    "eslint-plugin-vuetify": "^2.0.0-beta.4",
+    "react": "^18.2.0",
+    "react-dom": "^18.2.0",
```
jooola, line 48: Why do we need React ?

webapp/src/components/HelloWorld.vue
```diff
@@ -0,0 +1,65 @@
...
+                    </v-btn>
+                </v-col>
+            </v-row>
+        </v-responsive>
+    </v-container>
+</template>
+
+<script setup lang="ts">
+import { useI18n } from "vue-i18n";
+const { t } = useI18n();
```
jooola, line 63: Could we use a global `$t` function instead of the `useI18N` API ?

webapp/src/locale/po/README.md
```diff
@@ -0,0 +1,10 @@
+# Legacy locales
+
+To add a new locale, make sure to add/edit the following files:
+
+- `legacy/application/models/Locale.php`
+- `legacy/locale/<LANG>/LC_MESSAGES/libretime.po`
+- `legacy/public/js/datatables/i18n/<LANG>.txt`
+- `legacy/public/js/plupload/i18n/<LANG>.js`
+
+The `legacy/application/controllers/LocaleController.php` contains additional translations loaded by jquery i18n `$.i18n` and used with `$.i18n._`.
```
jooola, line 10: Please remove any duplicate or unneeded files.

All the PO files can be removed, the JSON files can be cleaned to only include the keys we actually use in the code.

webapp/src/plugins/vuei18n.ts
```diff
@@ -0,0 +1,56 @@
...
+import jaJP from '../locale/ja_JP.json';
+import koKR from '../locale/ko_KR.json';
+import nlNL from '../locale/nl_NL.json';
+import plPL from '../locale/nl_NL.json';
+import ptBR from '../locale/pt_BR.json';
+import ruRU from '../locale/ru_RU.json';
+import srRS from '../locale/sr_RS.json';
+import trTR from '../locale/tr_TR.json';
+import ukUA from '../locale/uk_UA.json';
+import zhCN from '../locale/zh_CN.json';
```
jooola, line 24: This is dangerous, we will end up with a huge bundle that will have to load all the locales, even when not used.

Please let the i18n message bundler handle this.

webapp/src/plugins/webfontloader.ts
```diff
@@ -0,0 +1,17 @@
...
+  const webFontLoader = await import(
+    /* webpackChunkName: "webfontloader" */ 'webfontloader'
+  );
+
+  webFontLoader.load({
+    google: {
+      families: ['Roboto:100,300,400,500,700,900&display=swap'],
+    },
+  });
+}
```
jooola, line 17: We don't want to use webfontloader, I'd prefer directly importing local fonts using the `fontsource` packages.

webapp/src/router/index.ts
```diff
@@ -0,0 +1,23 @@
...
+  children: [{
+    path: '',
+    name: 'Home',
+    // route level code-splitting
+    // this generates a separate chunk (about.[hash].js) for this route
+    // which is lazy-loaded when the route is visited.
+    component: () =>
+      import(/* webpackChunkName: "home" */ '@/views/Home.vue'),
+  },],
+},];
```
jooola, line 16: We will use a Multi Page Application structure during the transition, so I think we can remove the router for this initial setup.

It is unsure if we will be using the router for smaller routing feature.

webapp/src/stories/Button.stories.ts
```diff
@@ -0,0 +1,52 @@
+import type { Meta, StoryObj } from '@storybook/vue3';
+
+import Button from './Button.vue';
+
+// More on how to set up stories at: https://storybook.js.org/docs/vue/writing-stories/introduction
+const meta = {
+  title: 'Example/Button',
+  component: Button,
+  // This component will have an automatically generated docsPage entry: https://storybook.js.org/docs/vue/writing-docs/autodocs
+  tags: ['autodocs'],
```
jooola, line 10: Let's remove storybook for now, we will essentially be building entire pages using vuetify components, so I don't really see a use case right now, as we won't create a lot of small components.

webapp/vite.config.ts
```diff
@@ -0,0 +1,39 @@
...
+    template: { transformAssetUrls }
+  }),
+  // https://github.com/vuetifyjs/vuetify-loader/tree/next/packages/vite-plugin
+  vuetify({
+    autoImport: true,
+  }),
+  VueI18nPlugin({
+    include: [path.resolve(__dirname, './src/locale/*.json')],
+  })],
+  define: { 'process.env': {} },
```
jooola, line 23: Why do we need this ?

webapp/vite.config.ts
```diff
@@ -0,0 +1,39 @@
...
+      '.json',
+      '.jsx',
+      '.mjs',
+      '.ts',
+      '.tsx',
+      '.vue',],
+  },
+  server: {
+    port: 3000,
+  },
```
jooola, line 38: I assume this is the default ? Why do we specify it ? Is this for docker setup ?

webapp/src/plugins/vuetify.ts
```diff
@@ -0,0 +1,26 @@
...
+import { createVuetify } from 'vuetify';
+
+// https://vuetifyjs.com/en/introduction/why-vuetify/#feature-guides
+export default createVuetify({
+  theme: {
+    themes: {
+      light: {
+        colors: {
+          primary: '#1867C0',
+          secondary: '#5CBBF6',
```
jooola, line 21: Where did you pick those colors ?

webapp/src/components/HelloWorld.vue
```diff
@@ -0,0 +1,65 @@
...
+                        href="https://vuetifyjs.com/introduction/why-vuetify/#feature-guides"
+                        min-width="228"
+                        rel="noopener noreferrer"
+                        size="x-large"
+                        target="_blank"
+                        variant="flat"
+                    >
+                        <v-icon icon="mdi-speedometer" size="large" start />
+
+                        {{ t("Track Types") }}
```
jooola, line 39: I think we should use keys instead of `en_US` source strings. Or at least use some kind of structure to sort the strings.

For example:
```js
t('the-path.to-the-file.TheComponentName.the-translation-key')
```

webapp/package.json
```diff
@@ -0,0 +1,55 @@
...
+    "storybook": "storybook dev -p 6006",
+    "build-storybook": "storybook build",
+    "prettier": "prettier --write src",
+    "prettier:check": "prettier --check src",
+    "cypress:open": "cypress open",
+    "cypress:run": "cypress run"
+  },
+  "dependencies": {
+    "@mdi/font": "7.0.96",
+    "axios": "^1.4.0",
```
jooola, line 18: We don't yet use axios, let's only add things we really use.

jooola (commented):
I would like to focus on how we can embed the multipage application inside legacy before we ship a bunch of tools we are not yet using.

Let's start with 2 pages, the main one, we could leave it blank, and an About page that we embed inside legacy.

webapp/src/stories/Button.stories.ts
zklosko: Agreed

webapp/vite.config.ts
zklosko: Yes, it's the default config. I haven't worked on a Docker setup for it; I thought we were serving the generated pages through Django.

webapp/src/components/HelloWorld.vue
zklosko: Agreed. We would need to change how the files appear in Weblate.

paddatrapper: This app will appear as a different component in Weblate. Don't worry about matching legacy - the strings will all need re-translating anyway

zklosko: Okay. I've removed all the HTML from the keys in the translation files.

zklosko:
I'm having trouble getting Vite to split the locale files into separate bundles per language. It's importing all of them and the compiling all locales into the same output js file.

webapp/src/router/index.ts
jooola: To embed the new pages/views inside the legacy app, we will use the MPA architecture (multi page application), so the routing will be done server side.

It will only once we have all the views put together and that we can drop the essential block of legacy that we will be able to use the SPA architecture.

So unless I am missing something, I don't understand how this will be easier.

webapp/vite.config.ts
jooola: > I thought we were serving the generated pages through Django.

No, the compiled multi pages apps will be served by the legacy app / nginx.

> Yes, it's the default config. I haven't worked on a Docker setup for it;

According to https://vitejs.dev/config/server-options.html#server-port this is not the default, please remove this. The only thing that might need changing is the listen address, so we can reach the dev server within a docker container.

jooola:
> I'm having trouble getting Vite to split the locale files into separate bundles per language. It's importing all of them and the compiling all locales into the same output js file.

Hmm, I already spend some time on this, this is why I prefer when you coordinate your work with us, so we don't duplicate effort and waste precious time.

Anyway, feel free to copy and paste what I already built: https://github.com/jooola/libretime/tree/webui/web

zklosko:
Our lines of code regarding vue-i18n are the same, the difference is that I have all translations in my repo and you only have en_US. If I add the rest of the languages to your repo the results should be the same.

webapp/.editorconfig
zklosko: I think it's worthwhile.

zklosko:
I was able to get it to output multiple language files by using dynamic imports in Vite and bypassing the unplugin-vue-i18n plugin. The unplugin-vue-i18n plugin is recommended by vue-i18n, but doesn't seem to support dynamic imports. Vue-i18n's [Dynamic Imports page](https://vue-i18n.intlify.dev/guide/advanced/lazy.html) of the documentation doesn't mention use of the plugin at all.

webapp/src/router/index.ts
zklosko: It's substantially easier to use the router than custom-coding a MPA solution. In the research I've been doing, it's been recommended to create multiple SPAs (with only one view) and compile them all separately into multiple page apps if you really need an MPA: https://github.com/wandersonpjbkn/vue-mpa-application-example. Doing _n_ setups (where _n_ is the number of pages) seems excessive.

zklosko:
I believe I have a really basic about page set up as a demo. It mostly copies from the front page of our docs site.

webapp/src/App.vue
```diff
@@ -0,0 +1,11 @@
+<template>
+    <v-app>
+        <v-main>
+            <router-view />
+        </v-main>
+    </v-app>
+</template>
+
+<script setup lang="ts">
+import HelloWorld from "@/components/HelloWorld.vue"
```
paddatrapper, line 10: This doesn't exist in components

zklosko:
I translated the .po files using something like [Localize.biz](https://localise.biz/free/converter/po-to-json) or an npm package with a CLI. It's decoupled from our Weblate workflow.

Haven't thought too much about embedding the page, but I have it set up to generate a static site so we can serve an HTML file from `dist`.

webapp/src/App.vue
zklosko: Fixed

paddatrapper:
Do we have a way of generating a list of strings or keys from the strings used? Otherwise updating the set of translation source strings is difficult

zklosko:
I could write a node script to print all keys to a text file but I'm not sure Weblate requires that. Looking at their [documentation](https://docs.weblate.org/en/latest/formats/json.html), it seems like Weblate is fully compatible with the json files that Vue-i18n needs.

paddatrapper:
Maybe I am misunderstanding the JSON files, but they map a key used in the code to the translated string right? I'm asking about the mapping from used keys in the code to create the JSON files. For example, a new string is added to a view. How does that key end up in all the JSON files?

webapp/vite.config.ts
```diff
@@ -0,0 +1,42 @@
...
+    define: { "process.env": {} },
+    resolve: {
+        alias: {
+            "@": fileURLToPath(new URL("./src", import.meta.url)),
+        },
+        extensions: [".js", ".json", ".jsx", ".mjs", ".ts", ".tsx", ".vue"],
+    },
+    server: {
+        port: 8080,
+    },
```
jooola, line 33: We already use that port for other services, please use the default.

jooola:
eslint should be capable of checking if an i18n key used in the code is missing from the translation files. Maybe also the other way around.

But I would start with empty i18n json files, as we don't (almost) use translated strings yet. 

The development setup should also take advantages of the vite dev server:
```html
<script type="module" src="http://localhost:5173/src/main.ts"></script>
```

zklosko:
Translation strings are imported globally, then selected by key for templating':

```
<h4 class="text-h4 py-2">{{ $t("Getting Started") }}</h4>
```

or as part of `v-bind`:

```
v-bind:label="$t('Language')"
```

> I'm asking about the mapping from used keys in the code to create the JSON files. For example, a new string is added to a view. How does that key end up in all the JSON files?

I think you're thinking about it backwards. You would need to have a key:value pair in your JSON i18n translations before adding the string to a view as an i18n translatable string. It isn't automatic. All new strings would need to be added manually. For example, if we wanted to have a greeting message, we could have key:value `"Greeting": "Welcome to Libretime!"`; contrast this to what we're doing with PHP where our .po file looks like `"Welcome to Libretime!": "Welcome to Libretime!"`.

I know we have a few longer blocks of text with HTML tags that we would like to translate. Instead of having i18n strings for those, we may want to have markdown files (one per language) that we could render in the view so we can have more intricate explanations for things. For example, the following would be less practical, but not impossible, to do as a key:value pair:

```
#: application/controllers/EmbeddablewidgetsController.php:24
msgid ""
"To configure and use the embeddable player you must:<br><br>\n"
"            1. Enable at least one MP3, AAC, or OGG stream under Settings -> Streams<br>\n"
"            2. Enable the Public LibreTime API under Settings -> Preferences"
msgstr ""
```

> eslint should be capable of checking if an i18n key used in the code is missing from the translation files. Maybe also the other way around.

I think it's mainly the other way around. I found [this](https://github.com/intlify/eslint-plugin-vue-i18n/blob/master/docs/rules/no-unused-keys.md), but it cleans out unused i18n strings from your locale files.

> The development setup should also take advantages of the vite dev server, e.g.: `<script type="module" src="http://localhost:5173/src/main.ts"></script>`

There isn't a need to do that. `src="/src/main.ts"` is enough for Vite to work, and it doesn't require a fixed value to be changed later.

jooola:
> `<h4 class="text-h4 py-2">{{ $t("Getting Started") }}</h4>`

I would prefer to have keys here instead of the en_US string, I think I mentioned this already before.

e.g.: `path.to-the.file.ComponentName.translate-key`


> There isn't a need to do that. src="/src/main.ts" is enough for Vite to work, and it doesn't require a fixed value to be changed later.

I meant within the legacy page. Development will happen on the legacy page,so the php code should be able to load the webapp files and take advantages of the vite dev server.

zklosko:
> I would prefer to have keys here instead of the en_US string, I think I mentioned this already before.

That is the key. In this instance, the key is also the en_US string.

>      I meant within the legacy page. Development will happen on the legacy page,so the php code should be able to load the webapp files and take advantages of the vite dev server.

I don't understand. Why can't we build and then serve the built pages instead of running through the dev server? Running through a dev server is much less secure and could be harder to rectify for a user if it crashes. This really doesn't seem like the simplest way forward.

jooola:
> That is the key. In this instance, the key is also the en_US string.

Yes, this is the current behavior with the PO files. Now that we have json files and vue-i18n, I would like to use keys instead of the en_US source strings. See https://vue-i18n.intlify.dev/guide/essentials/syntax.html

> I don't understand. Why can't we build and then serve the built pages instead of running through the dev server? Running through a dev server is much less secure and could be harder to rectify for a user if it crashes. This really doesn't seem like the simplest way forward.

We have 2 environments, production and development. For production, we compile the multiple web app pages and serve them using some web server, the files aren't meant to change. For development, we will have to change the files often, and test the changes, we cannot rebuild the production setup every time.

Vite has a powerful development server, so we need to leverage this and use it in the dev setup while we develop the web app. This allows use to have hot reload, dev tools integration, and so on...

zklosko:
> The development setup should also take advantages of the vite dev server, e.g.: <script type="module" src="http://localhost:5173/src/main.ts"></script>

Wouldn't doing this break any test builds we would do for the new web app?

paddatrapper:
How would it break tests?

zklosko:
If we're hard coding the URLs of script sources, it won't run on its own without that exact port setup. If we did that during dev, we would need to switch everything back and test it for production.

jooola:
We leverage the app environement config to either setup the dev server or serve the production files, so this is not a problem.

The legacy dev setup will point to the webapp dev server, and the production with reference the files from the webapp build manifest.

This just needs a bit of php code, nothikg fancy

paddatrapper:
> "ui.preview": "Vorschau",

Yes, though I'm not sure this is actually a good idea with how weblate works. Using a source language helps translators know what meaning the string is supposed to have in the context that it is used.

paddatrapper:
Work continued in https://github.com/libretime/libretime/pull/3146 addressing the concerns discussed here

Issue #724, opened 2019-01-29
Email Support


zeLostchild:
Email could not be sent. Check your mail server settings and ensure it has been configured properly.

1. click forgot password
2. enter information and submit form
3. See error

ERROR There is no mail server settings in the GUI!!!!

I do not believe that there is an option to add an smtp login within the user interface. Is there a config file or something? If so please let me know. If this setting was removed, will it be replaced in a future update?

Robbt:
Yeah so for most people this probably won't work and adding the ability to use a 3rd party smtp server or what not might be handy but this is also probably a low priority in terms of development.

I think adding something in the documentation would be helpful and I'm going to change the title to reflect that this isn't a bug so much as something that we don't have good docs on how to setup and troubleshoot.

zeLostchild:
"Alternatively, you can specify an external SMTP server in the Preferences page on Airtime's System menu." _From SourceFabrics Manual for Airtime_
It seems like it was once an option. It might be buried in the code somewhere and might be just a matter of adding it to the UI.

Robbt:
I have found a non UI way of setting this up in PHP for people interested see [this link](https://websistent.com/how-to-use-msmtp-with-gmail-yahoo-and-php-mail/) - basically [msmtp](https://marlam.de/msmtp/) offers drop-in support for PHPs dependence upon sendmail. It could be configured with mailgun. Perhaps I'll write up the docs for how to configure this if people are interested.

zklosko:
I found an easy way to set up the mail() function in PHP here: https://www.guru99.com/php-mail.html. Is this something we still need to add to the docs?

marmotte32:
> I have found a non UI way of setting this up in PHP for people interested see [this link](https://websistent.com/how-to-use-msmtp-with-gmail-yahoo-and-php-mail/) - basically [msmtp](https://marlam.de/msmtp/) offers drop-in support for PHPs dependence upon sendmail. It could be configured with mailgun. Perhaps I'll write up the docs for how to configure this if people are interested.

Hi @Robbt 
Yes I'm interested on how to make work the forgot password e-mail function, could you send it to me please?

Robbt:
I am sorry I don't recollect how to do it aside from the comment above but I plan on diving back into Libretime at least briefly to upgrade my stations instance and I'll try follow up with this if I get a chance.

PR #2573, opened 2023-05-31
feat(api): add email configuration

jooola:
Related to #724 it will be fixed once we move the authentication/password reset to Django

### DIFF
--- a/api/libretime_api/settings/prod.py
+++ b/api/libretime_api/settings/prod.py
@@ -73,3 +73,20 @@ LANGUAGE_CODE = "en-us"
 TIME_ZONE = "UTC"
 USE_I18N = True
 USE_TZ = True
+
+# Email
+# https://docs.djangoproject.com/en/4.2/topics/email/
+
+EMAIL_BACKEND = "django.core.mail.backends.smtp.EmailBackend"
+
+EMAIL_HOST = CONFIG.email.host
+EMAIL_PORT = CONFIG.email.port
+EMAIL_HOST_USER = CONFIG.email.user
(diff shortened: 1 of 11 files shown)


jooola:
Woups, this is missing a lot of settings, e.g. the From email address, and maybe other stuff.

api/libretime_api/settings/prod.py
```diff
@@ -73,3 +73,20 @@
...
+# https://docs.djangoproject.com/en/4.2/topics/email/
+
+EMAIL_BACKEND = "django.core.mail.backends.smtp.EmailBackend"
+
+EMAIL_HOST = CONFIG.email.host
+EMAIL_PORT = CONFIG.email.port
+EMAIL_HOST_USER = CONFIG.email.user
+EMAIL_HOST_PASSWORD = CONFIG.email.password
+EMAIL_USE_TLS = CONFIG.email.security == "explicit"
+EMAIL_USE_SSL = CONFIG.email.security == "implicit"
```
jooola, line 87: `use_tls` or `use_ssl` are somewhat difficult to understand, so I changed this to use one of `None|implicit|explicit`.

- None is no tls encryption
- implicit is tls encryption
- explicit is tls encryption using starttls

paddatrapper: Hmm. I would maybe change the options to `None|SSL/TLS|STARTTLS` - those seem to be more consistent other email setting implementations (for example email clients, Nextcloud)

jooola: Yes, the options you proposed are better! I changed it.

PR #2654, opened 2023-08-13
feat(playout): configure device for alsa and pulseaudio system outputs

maxtimbo:
### Description
Attempting to add hardware configuration to liquidsoap so that users may set hardware output in config.yml. I'm focusing my efforts on ALSA for now, but PulseAudio should be easy enough to implement after ALSA is successful.

**This is a new feature**:

Add hardware configuration

### Testing Notes

**What I did:**

I'm using docker-compose on Ubuntu 20.04.
```
docker --version
Docker version 20.10.25, build 20.10.25-0ubuntu1~20.04.1
```

```
docker-compose --version
Docker Compose version v2.20.2
```

### Links

see my forum post for full notes:
https://discourse.libretime.org/t/hardware-configuration/1229/12

### DIFF
--- a/docker/config.template.yml
+++ b/docker/config.template.yml
@@ -328,5 +328,5 @@ stream:
         kind: pulseaudio
 
         # System output device.
-        # > only available for kind=pulseaudio or kind=alsa
+        # > only available for kind=(alsa, pulseaudio)
         device:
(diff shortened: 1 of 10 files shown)


jooola:
I rebased onto the main branch and added some changes to the shared models. This should be a good base to extend the ALSA System output.

jooola:
I took the liberty to fix the PR, but it is still missing some docs.

paddatrapper (commented):
Looks good, but I agree does need docs before we can merge

jooola:
Thanks @maxtimbo !

PR #2744, opened 2023-10-13
feat: add mobile devices stream config field

jooola:
### Description

Allow to configure streams to be used for mobile devices. The field was removed while moving to the new config file, the feature should be working :tm: .


jooola:
Some screenshots:
image
image

Issue #2047, opened 2022-08-12
Support >=php8.0


jooola:
We currently only support php7.2  -> php7.4. To install LibreTime on recent distributions using the php distribution packages, we will have to support >=php8.0.

A work around would be to install php7.4 from the [php sury repository](https://deb.sury.org/).

Most of our dependencies support php8.0, but we are using some abandoned dependencies that needs upgrading as well or replacements:
- https://github.com/jooola/celery-php needs to be tested against php8.0
- https://github.com/propelorm/Propel https://github.com/jooola/propel1 we have been tracking work from https://github.com/tworzenieweb/Propel. An interesting branch we could use https://github.com/tworzenieweb/Propel/tree/feature/php8.0

Test dependencies:

- phpunit/dbunit I wonder if we should drop or fork this.
- phpunit/phpunit zf1s seem to be maintain a phpunit version for testing their work https://github.com/zf1s/phpunit with php8.0

Probably some other dependencies (indirect) that also needs some upgrading.

Related to #1845 #777

Feel free to pick this issue and help us support php8.0 ! Help is greatly appreciated !

jooola:
I don't know if we want to only support >= php8.0 once this is implemented. As it would probably make our life easier. Though support for 7.4 and 8.x should be doable, but as php7.2 and 7.3 will be dropped with #2035 and #2036

jooola:
Some progress was made that we start trying to use:
- https://github.com/zf1s/zf1/commit/d7ba9c84463cbfd20231e51f806ac593bdb6895b
- https://github.com/tworzenieweb/Propel/tree/feature/php8.0

willvincent:
> I don't know if we want to only support >= php8.0 once this is implemented. As it would probably make our life easier. 

Considering 7.4 became EOL on January 1st, yes, only php8 should be supported moving forward. :)

paddatrapper:
Dropping 7.4 support requires dropping support for the distros that ship it, so it is more of a discussion that just drop it because upstream dropped it. Distro maintainers still provide support even if upstream doesn't

jooola:
https://github.com/propelorm/Propel/pull/1086#issuecomment-1362274606
https://github.com/dayspring-tech/Propel

paddatrapper:
A list of potential issues in our current code generated by https://github.com/PHPCompatibility/PHPCompatibility
[php8.0-report.log](https://github.com/user-attachments/files/18364473/php8.0-report.log)

Generated with
```bash
[...]

PR #2789, opened 2023-11-27
feat(legacy): update deprecated PHP code

mp3butcher:
### Description
update deprecated code. It's mergeable with master without syntax conflicts across php versions
remove deprecated((https://www.php.net/manual/fr/function.strftime.php)) and unsafe (https://cve.mitre.org/cgi-bin/cvekey.cgi?keyword=strftime)   **strftime** syntax


paddatrapper (commented):
These all look reasonable. Can you please add CI runs against PHP 8.x [here](https://github.com/libretime/libretime/blob/main/.github/workflows/legacy.yml) and fix the linting failures?

mp3butcher:
@paddatrapper Hi,
I'm not very familiar with CI but:
- these changes does't seem to involve php version conflicts (only deprecation)
- lint error is not clear at all. How do I debug these logs?

paddatrapper:
> these changes does't seem to involve php version conflicts (only deprecation)

Ah, I misunderstood the purpose of this PR and thought you were adding support for PHP 8.

> lint error is not clear at all. How do I debug these logs?

The failing 7.4 PHP linting CI shows diffs on formatting changes that need to be applied to your code. You can also test this locally via [pre-commit](https://libretime.org/docs/contributor-manual/development-workflows/#pre-commit). For example the following change highlights a space that needs removing:

```diff
--- /home/runner/work/libretime/libretime/legacy/application/services/MediaService.php
+++ /home/runner/work/libretime/libretime/legacy/application/services/MediaService.php
@@ -135,7 +135,7 @@
      */
     public static function areFilesStuckInPending()
     {
-        $oneHourAgo = gmdate( DEFAULT_TIMESTAMP_FORMAT, floor(hrtime(true)) - self::PENDING_FILE_TIMEOUT_SECONDS);
+        $oneHourAgo = gmdate(DEFAULT_TIMESTAMP_FORMAT, floor(hrtime(true)) - self::PENDING_FILE_TIMEOUT_SECONDS);
         self::$_pendingFiles = CcFilesQuery::create()
             ->filterByDbImportStatus(CcFiles::IMPORT_STATUS_PENDING)
             ->filterByDbUtime($oneHourAgo, Criteria::LESS_EQUAL)
```

legacy/application/services/MediaService.php
mp3butcher, line 1: not sure these last changes are required
- floor is mandatory to ensure int value (needed)
int is clearly awaited here
the problem is microtime is used multiple times in libretime.
Do I have to change them all to floor(microtime())?...

-microtime->hrtime is not mandatory but it's not unix only

legacy/application/models/airtime/om/BaseCcShowDays.php
mp3butcher, line 1: %formatting is deprecated
here i assumed parameter is default %x or %X
TODO proscribe % formatting in whole libretime

legacy/application/services/MediaService.php
paddatrapper: @jooola do you have any insight here?

legacy/application/models/ShowBuilder.php
```diff
@@ -283,9 +283,9 @@ private function makeScheduledItemRow($p_item)
             $formatter = new LengthFormatter(Application_Common_DateHelper::secondsToPlaylistTime($run_time));
             $row['runtime'] = $formatter->format();
 
-            $row['title'] = htmlspecialchars($p_item['file_track_title']);
-            $row['creator'] = htmlspecialchars($p_item['file_artist_name']);
-            $row['album'] = htmlspecialchars($p_item['file_album_title']);
+            $row['title'] = $p_item['file_track_title'] ? htmlspecialchars($p_item['file_track_title']) : '';
+            $row['creator'] = $p_item['file_artist_name']] ? htmlspecialchars($p_item['file_artist_name']) : '';
```
paddatrapper, line 287:
```suggestion
            $row['creator'] = $p_item['file_artist_name'] ? htmlspecialchars($p_item['file_artist_name']) : '';
```

paddatrapper: @mp3butcher this is what is causing the failure

mp3butcher: ...i'm tired....

legacy/application/models/airtime/CcShowInstances.php
```diff
@@ -12,7 +12,7 @@ class CcShowInstances extends BaseCcShowInstances
     /**
      * Get the [optionally formatted] temporal [starts] column value.
      *
-     * @param string $format The date/time format string (either date()-style or strftime()-style).
+     * @param string $format The date/time format string (date()-style).
```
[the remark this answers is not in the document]
paddatrapper: I think we can leave the exceptions - we aren't actively calling strftime and it should provide consumers a good explanation of why. The issue is that I suspect this is part of the API provided by the legacy code, used by the Python playout service, etc. Dropping it requires additional checks to see if we are not breaking anything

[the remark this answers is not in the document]
jooola: I don't know if this will break, I assume that this PR is proposing such change an investigation was made before?

jooola:
There are a lot of resources in regard to the PHP 8.0 support, I think it's wise to take a look at them and use the existing work of others: https://github.com/libretime/libretime/issues/2047

legacy/application/models/airtime/CcShowInstances.php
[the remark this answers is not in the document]
mp3butcher: I haven't spot any exception launched if it's what you mean...My main concern was to run libretime on ubuntu 22.04 with apache 8 and it work like a charm without strftime

jooola:
This PR is not testable, the docker dev setup does not work out of the box when bumping the php version to 8.1.

paddatrapper:
I thought the point of this PR was that is isn't upgrading PHP to 8.0. Rather it's a prerequisite for that to happen later

paddatrapper (approved):
LGTM

jooola:
All the changes under `/legacy/application/models/airtime` must come from the Propel Library, as those files as autogenerated. See the properl-gen make target in `/legacy/Makefile`.

Please propose a change against the propel library and we can then update the files by regenerating them here.

Other than that, the rest looks good.

jooola:
Maybe enabling https://www.php.net/manual/en/language.types.declarations.php#language.types.declarations.strict in our files is also a good idea?

mp3butcher:
i pushed a commit on https://github.com/libretime/propel1/pull/1

paddatrapper (changes requested):
https://github.com/libretime/propel1/pull/1 is now merged, can you please regenerate the files here? Then this should be ready to merge

paddatrapper (approved):
LGTM

PR #2793, opened 2023-11-30
undeclared variable

mp3butcher:
### Description
fileurl is used before being declared


legacy/application/views/scripts/playlist/update.phtml
```diff
@@ -7,7 +7,8 @@ if ($isSmartBlock && $this->showPoolCount) { ?>
         <span id='sp_pool_count' class='sp_text_font sp_text_font_bold'>
             <?php
             echo $this->poolCount;
-            echo ngettext(" track matches your search criteria.", " tracks match your search criteria.", $this->poolCount);
+            if($this->poolCount)
```
paddatrapper, line 10: If `poolCount` is undefined, then this should produce some sort of reasonable message that no tracks match the search criteria. Also should you not include the line above in this check? (i.e. `echo $this->poolCount;`)

mp3butcher: indeed

jooola:
This should have targeted the `stable` branch.

PR #2794, opened 2023-12-01
feat: add hls support

mp3butcher:
### Description

implement HLS .
rely on nginx to serve files as liquidsoap hls.harbor is not recommended for direct audience.
The cons is we loose listeners stats if hls.
a mount point have been created in nginx libretime.conf:
```
 location  /hls { 
     alias @@HLS_PLAYOUT_OUTPUT@@;
  }
```

### Note
- Player not upgraded: To test use https://github.com/libretime/libretime/pull/2797 or  vlc  open url http://localhost/hls/main 
- Will not work when moving to ubuntu 22.04 (liquidsoap 2.0 have differents hls stream definition -with %audio syntax-)

### Testing
- Tested on ubuntu 20.04 with vagrant
- docker bullseye


mp3butcher:
i don't understand ci message (sorry I have versions problems with prettier)
https://github.com/libretime/libretime/actions/runs/7059733283/job/19217904794?pr=2794
if someone does, he'd be welcome...

playout/libretime_playout/liquidsoap/1.4/ls_lib.liq
mp3butcher, line 1: make a function because streaminfos content varies accross version

paddatrapper:
I hope to properly test this in the coming weeks. Could you also please update the docs describing this feature and how to configure/use it?

mp3butcher:
> I hope to properly test this in the coming weeks. Could you also please update the docs describing this feature and how to configure/use it?

The config file template is pretty self explanatory...HLS is composed of several streams with different quality so you have to describe them via a tuple (streamname,codec,bitrate)...In the template i enable a hls output with 3 mp3 streams of differents quality....
However before to document new settings, I need to validate with you their pertinence.
The port setting is most controversal because we could use the same port as the admin/api interface (it'd makes the setting redundant)

playout/libretime_playout/liquidsoap/1.4/ls_lib.liq
mp3butcher, line 1: make a function because output.file.hls varies accross liquidsoap versions
TODO: check if 1.4 is happy with it

installer/nginx/libretime.conf
```diff
@@ -12,6 +12,10 @@ server {
   client_max_body_size 512M;
   client_body_timeout 300s;
 
+  location /hls {
+     alias /var/lib/libretime/playout/hls;
```
paddatrapper, line 16: So this assumes that there is a common volume between the nginx server and liquidsoap. This is not currently required and requires documenting. If HLS is going to be a default approach (which I would support at a later point), we will need to update the default Dockerfile to include a shared volume

shared/libretime_shared/config/_models.py
```diff
@@ -197,6 +197,31 @@ class AudioOpus(BaseAudio):
...
+    fragment_prefix: str = "mp3_128"
+    bitrate: str = "128k"
+    sample_rate: int = 44100
+    codec: str = "libmp3lame"
+
+
+class HlsOutput(BaseModel):
+    kind: Literal["hls"] = "hls"
+    enabled: bool = False
+    public_url: Optional[AnyUrl] = None
```
paddatrapper, line 210: If it isn't used, feel free to remove

shared/libretime_shared/config/_models.py
```diff
@@ -197,6 +197,31 @@ class AudioOpus(BaseAudio):
...
+    host: str = "localhost"
```
paddatrapper, line 211: This can be removed, I don't see it being used anywhere

shared/libretime_shared/config/_models.py
```diff
@@ -197,6 +197,31 @@ class AudioOpus(BaseAudio):
...
+    port: int = 80
```
paddatrapper, line 212: This can be removed, I don't see it being used anywhere

shared/libretime_shared/config/_models.py
mp3butcher: In the StreamSettings.php, public_url is constructed (if not specified) after output specific setup, so it's harmless to let it be: it allows one to explicitly setup public url instead of reconstructing from host:port/mount.
The  mobile settings seams also shared among differents outputs, but i haven't understand its usage.
My guess is to let them just in case

shared/libretime_shared/config/_models.py
mp3butcher: ok

shared/libretime_shared/config/_models.py
mp3butcher: ok

installer/nginx/libretime.conf
mp3butcher: About documenting this mount point:
do a note in hls outputs section of the admin manual can do the tricks, or do you think of an other place to mention it?
About docker:
I'm not familiar with it, but i'll check....

paddatrapper: We probably need to restructure https://libretime.org/docs/admin-manual/stream-configuration/ to have sub-pages for Icecast and HLS. Then each includes a walk through of how to set up and configure each. There are specific changes for each type that aren't generally relevant. For example changing the admin password in icecast or SSL support in HLS.

paddatrapper:
Is there a way to stream HLS to nginx instead of doing it via a file?

mp3butcher:
> Is there a way to stream HLS to nginx instead of doing it via a file?
I don't understand, HLS standard is based on serving files fragment so it's natural to serve it through a webserver (cause it serve files...)
liquidsoap can also serve these files itself but documentation recommend not to do it for direct audience for security purpose.

installer/nginx/libretime.conf
mp3butcher: i updated configuration.md but CI doesn't like it :(

mp3butcher:
> Is there a way to stream HLS to nginx instead of doing it via a file?

http://nginx-rtmp.blogspot.com/2017/07/introducing-nginx-ts-module-for-hls-and.html can be a lead
It might be possible to stream with unsafe hls.harbor on publish and serve on play

installer/nginx/libretime.conf
mp3butcher: i just added the alias in the docker/nginx.conf, .... I don't know if it's enough

legacy/application/models/StreamSetting.php
```diff
@@ -55,16 +55,26 @@ public static function getOutput($key, $add_prefix = false)
...
                 // $prefix . 'liquidsoap_error' => 'waiting',
             ];
+            if (array_key_exists('audio', $output)) {
+                $result = array_merge($result, [
+                    $prefix . 'channels' => $output['audio']['channels'] ?? 'stereo',
+                    $prefix . 'bitrate' => $output['audio']['bitrate'] ?? 128,
+                    $prefix . 'type' => $output['audio']['format'],
+                ]);
+            } elseif ($output['kind'] == 'hls') {
+                // HLS : set web server host and port
```
mp3butcher, line 72: here we could hack the mount point to add 'hls/' prefix instead of mandate it in configfile mount setting....I haven't do it because it's a bit hacky

paddatrapper:
After discussion, I think we should continue with the file-based solution instead - it is cleaner and induces less delay

installer/nginx/libretime.conf
mp3butcher: @paddatrapper  sorry i've missed some important code
I've tested on vagrant it works on 1.4

shared/libretime_shared/config/_models.py
```diff
@@ -197,6 +197,29 @@ class AudioOpus(BaseAudio):
     format: Literal[AudioFormat.OPUS] = AudioFormat.OPUS
 
 
+class HlsStream(BaseModel):
```
jooola, line 200: Please use the `HLSStream` casing.

shared/libretime_shared/config/_models.py
```diff
@@ -197,6 +197,29 @@ class AudioOpus(BaseAudio):
...
+    enabled: bool = False
+    public_url: Optional[AnyUrl] = None
+    format: str = "mpegts"
+    segment_duration: float = 2.0
+    segment_count: int = 5
+    segments_overhead: int = 5
+    streams: List[HlsStream] = Field([], max_items=10)
+    mount: str
+
+    mobile: bool = False
```
jooola, line 218: Mobile is unneeded, please remove it.

shared/libretime_shared/config/_models.py
```diff
@@ -197,6 +197,29 @@ class AudioOpus(BaseAudio):
     format: Literal[AudioFormat.OPUS] = AudioFormat.OPUS
 
 
+class HlsStream(BaseModel):
+    fragment_prefix: str = "mp3_128"
+    bitrate: str = "128k"
+    sample_rate: int = 44100
+    codec: str = "libmp3lame"
```
jooola, line 204: Could we remove the defaults? It doesn't make sens to me, as we will have the default for a single quality, but not the other ones.

legacy/application/models/StreamSetting.php
jooola: We want good default, but they should be configurable. Having docs/a comment say 'you need hls/ as prefix' is not optimal.

installer/nginx/libretime.conf
```diff
@@ -12,6 +12,10 @@ server {
   client_max_body_size 512M;
   client_body_timeout 300s;
 
+  location /hls {
```
jooola, line 15: This should be configurable, at least for our advanced installers.

installer/nginx/libretime.conf
jooola, line 16: This should use the variables from the installer.

installer/config.yml
```diff
@@ -317,6 +317,47 @@ stream:
...
+        # > segments count  (default:5)
+        segment_count: 5
+        # > segments_overhead (default:5)
+        segments_overhead: 5
+        # Output public url, If not defined, the value will be generated from
+        # the [general.public_url] hostname, the output port and mount.
+        public_url:
+        # hls mount point
+        # > this field is REQUIRED
+        mount: hls/main
```
jooola, line 338: The word "mount" is from the icecast/shoutcast vocabulary, I am not sure we want to have if in the context of HLS.

I think this should also include the playlist extension .m3u8

installer/config.yml
```diff
@@ -317,6 +317,47 @@ stream:
         # > default is false
         mobile: false
 
+    hls:
+      - # Whether the output is enabled.
+        # > default is false
+        enabled: false
+        # > must be one of ('mpegts', 'mp3', 'adts', 'mp4')
+        # > this field is REQUIRED
```
jooola, line 325: Let's reduce the amount of format for now, extra format can be added and tested in future PRs.

installer/config.yml
```diff
@@ -317,6 +317,47 @@ stream:
...
+        # the [general.public_url] hostname, the output port and mount.
+        public_url:
+        # hls mount point
+        # > this field is REQUIRED
+        mount: hls/main
+        streams:
+          - # > prefix of generated fragment
+            # > this field is REQUIRED
+            fragment_prefix: mp3_low
+            # > codec must be one of ( 'libmp3lame', 'flac', 'aac', 'libopus', 'libvorbis')
```
jooola, line 343: Let's not diverge from the spec: https://datatracker.ietf.org/doc/html/rfc8216#section-3.4

installer/config.yml
```diff
@@ -317,6 +317,47 @@ stream:
...
+            codec: libmp3lame
+            # > bitrate of the stream
+            # > this field is REQUIRED
+            bitrate: 32k
+            # > sampling rate (default: 44100Hz)
+            sample_rate: 44100
+          - fragment_prefix: mp3_med
+            codec: libmp3lame
+            bitrate: 128k
+          - fragment_prefix: mp3_hifi
```
jooola, line 354: Maybe `segments_prefix` is more consistent with the other settings?

install
```diff
@@ -628,6 +628,7 @@ link_python_app libretime-playout-notify
 
 info "creating libretime-playout working directory"
 mkdir_and_chown "$LIBRETIME_USER" "$WORKING_DIR/playout"
+mkdir_and_chown "$LIBRETIME_USER" "$WORKING_DIR/playout/hls"
```
jooola, line 631: I prefer to have playout create the "$WORKING_DIR/playout/hls" directory, the "$WORKING_DIR/playout" is writable by libretime.

legacy/application/models/StreamSetting.php
```diff
@@ -55,16 +55,28 @@ public static function getOutput($key, $add_prefix = false)
...
+                $result = array_merge($result, [
+                    $prefix . 'port' => $_SERVER['SERVER_PORT'],
+                    $prefix . 'host' => $_SERVER['SERVER_NAME'],
+                    $prefix . 'type' => 'hls',
+                    $prefix . 'bitrate' => '',
+                ]);
```
jooola, line 78: Hmm, I think we prefer to use the config public url and not the webserver configs.

jooola (changes requested):
Hi, I made first pass. 

I think we should implement a simple working HLS feature, and iterate to add more options. Supporting too many formats/codecs is also risky, let's do not diverge from the spec.

I already started working on HLS a while back, did you take a look at my PR:  https://github.com/libretime/libretime/pull/2234

legacy/application/models/StreamSetting.php
mp3butcher: in the present configuration, you can use both as with other outputs:
- default is webserver
- if public url is provided  it's use instead

jooola: The rest of the project doesn't rely on the web server configuration, this would diverge from the current expectation.

mp3butcher: so you tell me to add settings for web server host/port?

jooola: We use the `Config::get('general.public_url')` (with the _raw key you get access to the URL object (use League\Uri\Uri;).

https://github.com/libretime/libretime/blob/e2cfbf4c038f28874a206df5805f04f69a40647b/legacy/application/models/StreamSetting.php#L70-L81

jooola: But I think we create an new function and not use the backward compatible and complex Application_Model_StreamConfig::getOutput function.

installer/config.yml
mp3butcher: AC3/ADTS is in the spec, further it work as it should with hls.js so i'll let it .

I'll remove mp4 because i haven't achieve anything with liquidsoap 1.4 even if in spec...:/

ok to remove mp3

I'll will move format property as stream property to fit liquidsoap stream model (it was a mistake to put it as hls property)

legacy/application/configs/conf.php
```diff
@@ -270,7 +298,8 @@ private static function load()
         // Merge Icecast and Shoutcast outputs
         $values['stream']['outputs']['merged'] = array_merge(
             $values['stream']['outputs']['icecast'],
-            $values['stream']['outputs']['shoutcast']
+            $values['stream']['outputs']['shoutcast'],
+            $values['stream']['outputs']['hls']
```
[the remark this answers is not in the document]
mp3butcher: It's mandatory to do that, it doesn't work if i remove this line

installer/config.yml
mp3butcher: ok for m3u8 but i don't know if we need to change the mount property name it's quite coherent for a user perspective...
What would you suggest? manifest?

install
mp3butcher: the hls directory must be created before nginx starts so we can't let playout create this directory...no?

mp3butcher:
Are you okay with these settings:

```yml
stream:
  outputs:
    hls:
      - # Whether the output is enabled.
        # > default is false
        enabled: false
        # Output public url. If not defined, the value will be generated from
        # the [general.public_url] hostname and the output port.
        public_url:
        # hls manifest name.
        # > this field is REQUIRED
        manifest: "main.m3u8"
        # > segment_duration (default:2.0)
        segment_duration: 2.0
        # > segments count: segments count buffered (default:5)
        segment_count: 5
        # > segments_overhead: segments count kept past the playlist size for those listeners who are still listening on outdated segments. (default:5)
        segments_overhead: 5
        streams:
          - # > prefix of generated segments
            # > this field is REQUIRED
            segments_prefix: mp3_low
            # > format of the container must be one of ('mpegts', 'adts')
            # > this field is REQUIRED
            format: mpegts
            # > codec used for the stream, must be one of ( 'libmp3lame', 'aac')
            # > this field is REQUIRED
            codec: libmp3lame
            # > bitrate of the stream
            # > this field is REQUIRED
            bitrate: 32k
            # > audio sampling rate
            # > this field is REQUIRED
            sample_rate: 44100
          - fragment_prefix: mp3_hifi
            codec: libmp3lame
            bitrate: 256k
            sample_rate: 44100

```

legacy/application/models/StreamSetting.php
mp3butcher: i generate mount with a prefix hls now

installer/config.yml
jooola: I believe playlist is the word used in the spec.

See https://github.com/libretime/libretime/pull/2234/files#diff-63e70e96336c955e4d36d6178baa323760305a40547cdd6de14ad8e1c64c7c2e

mp3butcher: Are you okay with these settings?:

```yml
stream:
  outputs:
    # HLS output manifests.
    # > max items is 10
    hls:
      - # Whether the output is enabled.
        # > default is false
        enabled: false
        # Output public url. If not defined, the value will be generated from
        # the [general.public_url] hostname/port.
        public_url:
        # hls playlist name.
        # > this field is REQUIRED
        playlist: "main.m3u8"
        # > segment_duration (default:2.0)
        segment_duration: 2.0
        # > segments count: segments count buffered (default:5)
        segment_count: 5
        # > segments_overhead: segments count kept past the playlist size for those listeners who are still listening on outdated segments. (default:5)
        segments_overhead: 5
        streams:
          - # > prefix of generated segments
            # > this field is REQUIRED
            segments_prefix: mp3_low
            # > format of the container must be one of ('mpegts', 'adts')
            # > this field is REQUIRED
            format: mpegts
            # > codec used for the stream, must be one of ( 'libmp3lame', 'aac')
            # > this field is REQUIRED
            codec: libmp3lame
            # > bitrate of the stream
            # > this field is REQUIRED
            bitrate: 32k
            # > audio sampling rate
            # > this field is REQUIRED
            sample_rate: 44100
          - fragment_prefix: mp3_hifi
            codec: libmp3lame
            bitrate: 256k
            sample_rate: 44100

```

playout/libretime_playout/liquidsoap/templates/outputs.liq.j2
```diff
@@ -41,6 +41,44 @@
...
+{%- macro output_hls(output_id, output) -%}
+# hls:{{ output_id }}
+
+output_hls_{{ output_id }}_source = s
+
+streams = ref([])
+
+{% for stream in output.streams -%}
+
+    streamout = %ffmpeg(format="{{output.format}}" ,
```
mp3butcher: note that %ffmpeg (as well as %audio for liquidsoap>2.0) arguments can't be dynamically passed through a function we can cursomize for different lq versions (as i did for output.file.hls).....So we can't maintain hls for both lq 1.4 and 2.0

legacy/application/configs/conf.php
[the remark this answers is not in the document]
jooola: Why is it mandatory? "It doesn't work" doesn't seem like a valid argument sorry :wink:

install
jooola: Why ? We only read in this directory during a request, and response with 404 if no files are found. Sound doable to me.

playout/libretime_playout/liquidsoap/templates/outputs.liq.j2
jooola: For the futur, if you really need to handle the %functions, you may use jinja templating and check against the version during the startup of playout.

legacy/application/configs/conf.php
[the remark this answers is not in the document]
mp3butcher: Streams settings data communicated to the player are retrieved from the merged array

install
mp3butcher: if you think it's safe to launch nginx and serve a not-yet-created directory, so i'm good with removing this line

PR #2850, opened 2023-12-29

caveman99:
### Description

### Testing Notes

Playground server with new code successfully executed DELETE api call. WIP/Draft since error handling for the delete operation is not yet in place.

This is deliberately against STABLE since this is not a breaking change but implements defined API behaviour.


caveman99:
I will rebase this against main and debug the CI complaints.

PR #2876, opened 2024-01-07
docs: add root override for init

caveman99:
when following this guide, i was running inter 'permission denied' errors without adding the --user=root flag for the setup run.


jooola:
Could you provide the error message you are having, because I will not accept this unless there is a very good reason.

Development should happen as regular user, not root/sudo.

caveman99:
> Development should happen as regular user, not root/sudo.

That was the root cause, and it occured to me after other errors cropped up. Please disregard this PR:

PR #2877, opened 2024-01-08
feat(playout): record using Liquidsoap

### LINKED ISSUE #42: Reinstate record from line in from Airtime 2.5.x

GroundTruthNZ:
### Description

**This is a re-implemented feature based on existing code that was available in Airtime**:

image

This re-enables the Record and Rebroadcast functions originally available in Airtime.  Except, instead of using Ecasound to record from Line-In, it uses Liquidsoap to do the actual recording.   There are a few advantages to doing this - no extra software binary is required, and the recording will work without a soundcard on any cloud-based server.

I've managed to get the recording part working:  and it creates Ogg files in the app/recordings folder with the ID3 tags intact as Airtime used to do.  However, my Python foo is limited and the changes to Libretime subsequently mean that some coding is still required to automatically load the recordings into the tracks library.  I'm hoping that someone can help with this.

Some work will be required to handle requests to cancel the show, and recording failures.  The original functions monitored the Ecasound process whereas the Liquidsoap API doesn't give you much feedback.  An approach might be to wait say 10 seconds after the recording has finished and then delete the file if the recording is cancelled.

If we can get the basic end to end record-upload-rebroadcast functions working, I'd happy to spend more time on this testing and tweaking, writing docs, etc.

### **Links**

https://github.com/LibreTime/libretime/issues/42


GroundTruthNZ:
Now have recordings automatically loaded to library.  However they aren't being added to the schedule.

playout/libretime_playout/liquidsoap/1.4/ls_script.liq
```diff
@@ -197,6 +197,15 @@ else
...
 end
 
+def start_recording(arguments)
+    #E.g. {"filename" : "/app/recorder/test2.ogg", "length" : "25"}
+    args = of_json(default= [("error","fail")],arguments)
+    log("sources.start_recording #{args}")
+    recording_length = float_of_string(default=0.,args["length"])
+    recording = max_duration(recording_length,s)
+    output.file(%vorbis, args["filename"],fallible=true, recording)
+end
```
jooola, line 207: This does not take into account the config recorder settings:
```
    record_file_format: Literal["mp3", "ogg"] = "ogg"  # record_file_type
    record_bitrate: int = 256
    record_samplerate: int = 44100
    record_channels: int = 2
    record_sample_size: int = 16
```

playout/libretime_playout/recorder.py
```diff
@@ -0,0 +1,402 @@
...
+
+        br = self.config.playout.record_bitrate
+        sr = self.config.playout.record_samplerate
+        c = self.config.playout.record_channels
+        ss = self.config.playout.record_sample_size
+
+        liq_client = LiquidsoapClient(
+            host=self.config.playout.liquidsoap_host,
+            port=self.config.playout.liquidsoap_port,
+        )
```
jooola, line 87: Please use the existing liquidsoap client created in the main.

playout/libretime_playout/recorder.py
```diff
@@ -0,0 +1,402 @@
...
+        # # blocks at the following line until the child process
+        # # quits
+        # self.p.wait()
+        # outmsgs = self.p.stdout.readlines()
+        # for msg in outmsgs:
+        #     m = re.search("^ERROR", msg)
+        #     if not m == None:
+        #         logger.info("Recording error is found: %s", outmsgs)
+        # logger.info("finishing record, return code %s", self.p.returncode)
+        # code = self.p.returncode
```
jooola, line 118: Please remove unused code.

playout/libretime_playout/recorder.py
```diff
@@ -0,0 +1,402 @@
...
+        ss = self.config.playout.record_sample_size
+
+        liq_client = LiquidsoapClient(
+            host=self.config.playout.liquidsoap_host,
+            port=self.config.playout.liquidsoap_port,
+        )
+        # args = f'{{"filename" : "{filepath}", "length" : "{length}"}}'
+        liq_client.start_recording(dict(filename=filepath, length=f"{length}"))
+        delay = int(length) + 10
+        time.sleep(delay)
```
jooola, line 91: Can't we fire and forget this ?

jooola (commented):
This PR holds a lot of legacy code that I tried to get rid of. Our goal is to move the Django API.

- For playout, what do you think about querying the django API to list the show that needs recording?
- We need to add tests and typings so it's easier for all of us to review and maintain these changes.
- The idea around liquidsoap start recording function seem like a good idea.

playout/libretime_playout/liquidsoap/1.4/ls_script.liq
GroundTruthNZ: It turns out that you can't use a variable as a parameter in Liquidsoap's encoding format (see https://www.liquidsoap.info/doc-1.4.4/encoding_formats.html) . 

This might have been fixed in 2.1.1. 

I've updated the code so that it uses the file format (mp3 or ogg).   But not sure what to do with the rest.  Perhaps leave the Liquidsoap defaults or offer a selection of defaults?

playout/libretime_playout/recorder.py
GroundTruthNZ: Done

playout/libretime_playout/recorder.py
GroundTruthNZ: Done

playout/libretime_playout/recorder.py
GroundTruthNZ: The Liquidsoap API doesn't give you much in the way of feedback.  So yeah I was thinking just set it off and wait a little after the recording has finished.

GroundTruthNZ:
Thanks a lot for taking the time to look at the @jooola.  Regarding querying the Django API to list the show that needs recording - that sounds sensible but beyond my capabilities at the moment.  I think I'll need to leave this and hopefully someone else can pick it up.

PR #2881, opened 2024-01-09
feat(legacy): add aac/opus support to dashboard player

caveman99:
### Description

The Player on the dashbaord does not correctly play an AAC stream. The rewritng logic only takes into account the ogg format. Add rewriting from aac to m4a in the javascript calls so the player works.

**This is a new feature**:

AAC output currently only works with replacing liquidsoap, so technically this is a new feature.

The dashboard player is not documented. The player widget works with AAC streams already.

### Testing Notes

**What I did:**

Swapped liquidsoap and reconfigured the default mp3 stream as AAC

on ubuntu focal, drop in the upstream liquidsoap deb to enable aac, then configure the secondary output to use 128kbit aac instead of mp3. Restart the target, login to the dashboard and open the dashboard player. Both streams from the dropdown should work.

### DIFF
--- a/legacy/application/views/scripts/dashboard/stream-player.phtml
+++ b/legacy/application/views/scripts/dashboard/stream-player.phtml
@@ -41,6 +41,8 @@
                 $serverType = $streamData["${id}_output"];
                 if ($type == "ogg")
                     $type = "oga";
+                if ($type == "aac")
+                    $type = "m4a";
                 echo "setjPlayer('$url', '$type', '$serverType');";
             }
             ?>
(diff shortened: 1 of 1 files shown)


legacy/application/views/scripts/dashboard/stream-player.phtml
```diff
@@ -41,6 +41,8 @@
                 $serverType = $streamData["${id}_output"];
                 if ($type == "ogg")
```
jooola, line 42:
```suggestion
                if ($type == "ogg" || $type == "opus")
```

legacy/application/views/scripts/dashboard/stream-player.phtml
```diff
@@ -73,6 +75,8 @@
                     $serverType = $streamData["${id}_output"];
                     if ($type == "ogg")
```
jooola, line 76:
```suggestion
                    if ($type == "ogg" || $type == "opus")
```

jooola (approved):
I will apply the small 2 additions, I tested it myself while testing your fix.

PR #2882, opened 2024-01-10

caveman99:
Description

Testing Notes

Playground server with new code successfully executed DELETE api call.


api/libretime_api/storage/views/file.py
```diff
@@ -29,3 +29,18 @@ def download(self, request, pk=None):  # pylint: disable=invalid-name
...
+    @action(detail=True, methods=["DELETE"])
+    def delete_file(self, request, pk=None):  # pylint: disable=invalid-name
+        pk = IntegerField().to_internal_value(data=pk)
+
+        file = get_object_or_404(File, pk=pk)
+        path = file.filepath
+
+        try:
+            if os.path.isfile(path):
+                os.remove(path)
```
paddatrapper, line 42: > Right now the code will only manipulate the database but leave the file in place.

This line implies the opposite? It will remove the actual file

caveman99: I worded that poorly. What i wanted to say is, that the file deletion is introduced with this patch, and without the patch the database is updated but the file stays on storage. 

Use case is an external script that prunes old podcast episodes from the server.

mp3butcher: @caveman99  libretime/api/schema.yml may require change to pass check

jooola: A `make -C api schema` from the root of the project should fix it.

api/libretime_api/storage/views/file.py
```diff
@@ -29,3 +29,18 @@ def download(self, request, pk=None):  # pylint: disable=invalid-name
...
+        except OSError:
+            return HttpResponse(status=404)
```
jooola, line 44:
```suggestion
        except OSError as exception:
            logger.error(f"Could not delete file from storage: {exception}")
            return HttpResponse(status=404)
```
Could you please log the error here ? This should not happen on a regular basis. The logger in the package is probably missing.

api/schema.yml
```diff
@@ -298,6 +298,24 @@ paths:
       responses:
         "204":
           description: No response body
+  /api/v2/files/{id}/delete_file:
```
jooola, line 301: I'd prefer to have this at `DELETE /api/v2/files/{id}`

api/libretime_api/storage/views/file.py
```diff
@@ -29,3 +29,19 @@ def download(self, request, pk=None):  # pylint: disable=invalid-name
...
+
+        file = get_object_or_404(File, pk=pk)
+        path = file.filepath
+
+        try:
+            if os.path.isfile(path):
+                os.remove(path)
+        except OSError as exception:
+            logger.error(f"Could not delete file from storage: {exception}")
+            return HttpResponse(status=404)
```
jooola, line 45: I think this should return a 500.

api/libretime_api/storage/views/file.py
```diff
@@ -29,3 +29,19 @@ def download(self, request, pk=None):  # pylint: disable=invalid-name
...
+        pk = IntegerField().to_internal_value(data=pk)
+
+        file = get_object_or_404(File, pk=pk)
+        path = file.filepath
+
+        try:
+            if os.path.isfile(path):
+                os.remove(path)
+        except OSError as exception:
+            logger.error(f"Could not delete file from storage: {exception}")
```
jooola, line 44: The logger obj is missing in this module, please add one.

jooola (changes requested):

api/libretime_api/storage/views/file.py
caveman99: I was debating that. 4040 wuld mean "i didn't find the file i was supposed to delete" while 500 means "the database entry is there but the file isn't, there's something fishy here". Fine with 500 as well :-)

jooola: I think the behavior is a slightly different: we do check if the files exist first. So if the file is missing, we continue by deleting the database entry.

But if deleting the storage file fails because of e.g. permission denied error, then we have another problem.

So yes, we agree on the 500.

jooola:
What happens if the file is still used in a schedule? I think legacy implements a lock when the file is still used somewhere else, we might need to check too.

api/schema.yml
caveman99: That's what i had in my first attempt. I think this is the result of a brainfart on my part. lemme recheck the code.

jooola: If you look at the view class we inherited from, you have the `DestroyModelMixin`. You can override the `perform_destroy` function:
```
class DestroyModelMixin:
    """
    Destroy a model instance.
    """
    def destroy(self, request, *args, **kwargs):
        instance = self.get_object()
        self.perform_destroy(instance)
        return Response(status=status.HTTP_204_NO_CONTENT)

    def perform_destroy(self, instance):
        instance.delete()
```

caveman99: Had to override destroy itself because the perform_destroy does not return results but its working now. Need to run more manual tests today.

caveman99:
> What happens if the file is still used in a schedule? I think legacy implements a lock when the file is still used somewhere else, we might need to check too.

Inclusion in schedule is checked now, mimicking the UI/legacy behaviour.

caveman99:
i have no idea what the problem with schema.yml is. fwiw it's the correct file in the repository, made from `make schema`

caveman99:
for the record, i am giving up on this check now.

caveman99:
closing this in favor of https://github.com/libretime/libretime/pull/2960 (squashed, checks pass cleanly)

PR #2897, opened 2024-01-14
feat(legacy): trim overbooked shows after autoloading a playlist

caveman99:
### Description

Some combination of preload/postload and autoloding playlists with smartblocks generate massively overbooked shows that clutter up the interface. This addition performs a 'trim overbooked' after filling up the autoload list, and does the same as pushing the 'trim overbooked' button in the UI.

**This is a new feature**:

New feature that helps with keeping the interface tidy.

If a documentation update to reflect this change is deemed neccessary i will provide one.

### Testing Notes

**What I did:**

Define an autoloading playlist of 2 hours and schedule it for a one hour show. Without patch, you'll get entries for 2 hours, with the patch, you only get one hour and a 'overboarding' orange entry at most.

see above. UI Button press has identical behaviour.


caveman99:
My main gripe with the current system is, if you have a fixed schedule and ingest podcast episodes, filling the empty space behind the podcast with a smart block results in massively overbooked shows, if you also want to make sure the random routine doesnt pick a song twice. We don't really interact in this case, merely monitor the playout. When manually editing the autoloaded show the person doing it is responsible for making sure the airtime is fully used and theres no silence. So i took the obvious approach and added it to the autoloader in legacy. 

Your suggestion of doing it 24h after the show aired is more a point of database hygene, not UI hygene. I would rather opt for making my patch configurable/switchable. Adding database cleanup via celery 24h after airtime can be done as well. But again those are 2 different stories :-)

paddatrapper:
I like this change, but maybe it should be behind a configuration toggle? Then people who wish to keep things overbooked can

caveman99:
> I like this change, but agree maybe it should be behind a configuration toggle? Then people who wish to keep things overbooked can

Setting added. Defaults to old behaviour. If i don't have any spelling errors inside ;-)

legacy/application/models/ShowInstance.php
```diff
@@ -844,4 +844,33 @@ public function isRepeating()
...
+    {
+        // iterate through all of the showinstance items and delete those with start time greater than endtime of the show
+        $showEnd = new DateTime($this->getShowInstanceEnd(), new DateTimeZone('UTC'));
+        foreach ($this->getShowListContent('UTC') as $item) {
+            $itemStart = new DateTime($item['starts'], new DateTimeZone('UTC'));
+            if ($itemStart->getTimestamp() > $showEnd->getTimestamp()) {
+                $this->removeItem($item['item_id'], $item['starts']);
+            }
+        }
+    }
```
jooola, line 858: Would it be possible to execute all the remove Item at once? This feels like a lot for sql requests, or not?

legacy/application/models/Preference.php
```diff
@@ -1399,6 +1399,16 @@ public static function setRadioPageDisplayLoginButton($value)
         self::setValue('radio_page_display_login_button', $value);
     }
 
+    public static function getSmartblockTrimOverbooked()
+    {
+        return self::getValue('smartblock_trim_overbooked');
```
jooola, line 1404: Could we rename this key to `schedule_trim_overbooked` (and all related function/fields names), we are effectively touching the schedule table.

jooola: Also, we might want to add a default either in the code, or in the database (using the default DB data + migrations).

In the code seem easier/cleaner.

jooola (changes requested):
Looks great, I will test and do another review once the few comments are addressed.

legacy/application/models/ShowInstance.php
caveman99: I'm not sure how to make a combined sql call delete the entry that match exactly the listed tuples, so i opted for the loop approach. I'll look into the table definition if there is a primary key that can be used instead.

legacy/application/models/Preference.php
caveman99: rename sounds logical. there is no verbatim default cause the default is "false" which is equal to an unset database field or empty string. This is actually checked in the preferences dalogue after the getter is used. So all installations will default to 'false' which is the original behaviour.

legacy/application/models/ShowInstance.php
jooola: If we want to rely on the playout status (seem risky, don't know if this will break something else down the line), we can do: 

```sql
      DELETE FROM cc_schedule
           WHERE playout_status == 2 -- please make sure 2 is the right status.
        SQL;
```

But maybe a single explicit sql query can handle this.

caveman99: playout_status was inconsistent in my live db so i didn't use it. there's a primary key we can use instead.

legacy/application/forms/GeneralPreferences.php
```diff
@@ -150,6 +150,21 @@ public function init()
...
+        $scheduleTrimOverbooked = new Zend_Form_Element_Checkbox('scheduleTrimOverbooked');
+        $scheduleTrimOverbooked->setDecorators([
+            'ViewHelper',
+            'Errors',
+            'Label',
+        ]);
+        $displayScheduleTrimOverbookedValue = Application_Model_Preference::getScheduleTrimOverbooked();
+        if ($displayScheduleTrimOverbookedValue == '') {
+            $displayScheduleTrimOverbookedValue = false;
+        }
```
jooola, line 162: Please move this to `Application_Model_Preference::getScheduleTrimOverbooked()`

legacy/application/forms/GeneralPreferences.php
```diff
@@ -150,6 +150,21 @@ public function init()
...
+        $scheduleTrimOverbooked->addDecorator('Label', ['class' => 'enable-tunein']);
```
jooola, line 163: Why is there a enable-tunein class here?

legacy/application/models/ShowInstance.php
```diff
@@ -844,4 +844,30 @@ public function isRepeating()
...
+            }
+        }
+
+        // now remove the items
+        $sql = <<<'SQL'
+        DELETE FROM cc_schedule
+         WHERE id in (:ids)
+           AND instance_id = :instance_id
+           AND playout_status >= 0
+        SQL;
```
jooola, line 866: We could also make this in a single sql only call, without the loops:
```
delete
from
	cc_schedule
where
	id in (
	select
		s.id
	from
		cc_schedule s
	left join cc_show_instances si on
		s.instance_id = si.id
	where
		si.id = :instance_id
		and si.ends < s.starts
		and s.playout_status = 0 -- Double check si.ends < s.starts
);
```

legacy/application/forms/GeneralPreferences.php
caveman99: Copy/Paste from the 'Login disable' checkbox that was already there.

legacy/application/models/ShowInstance.php
caveman99: That call would work too ... for my mental sanity i try to avoid those. The loop does not do database lookups and the code is identical to what the UI button does. IIRC the crossfade is already calculated from the cuepoints of the mp3 files when the show is populated, so it will work either way.

legacy/application/models/ShowInstance.php
```diff
@@ -844,4 +844,30 @@ public function isRepeating()
...
+
+        return Application_Common_Database::prepareAndExecute($sql, [
+            ':ids' => implode($removeItems, ','),
```
jooola, line 869: For your information, this should be `implode(',', $removeItems)`.

I've already fixed it.

legacy/application/models/ShowInstance.php
```diff
@@ -844,4 +844,30 @@ public function isRepeating()
...
+
+    public function trimOverbooked()
+    {
+        // iterate through all of the showinstance items and delete those with start time greater than endtime of the show
+        $showEnd = new DateTime($this->getShowInstanceEnd(), new DateTimeZone('UTC'));
+        $removeItems = [];
+        foreach ($this->getShowListContent('UTC') as $item) {
+            $itemStart = new DateTime($item['starts'], new DateTimeZone('UTC'));
+            if ($itemStart->getTimestamp() > $showEnd->getTimestamp()) {
+                $removeItems[] = $item['id'];
```
jooola, line 856: Accessing the ID like this seem not to work.

I've already fixed it.

legacy/application/models/Preference.php
```diff
@@ -1399,6 +1399,16 @@ public static function setRadioPageDisplayLoginButton($value)
         self::setValue('radio_page_display_login_button', $value);
     }
 
+    public static function getScheduleTrimOverbooked()
+    {
+        return boolval(self::getValue('schedule_trim_overbooked'), false);
```
jooola, line 1404: For your information, the false should be an argument of the getValue function, it provides the default value.

Already fixed it.

jooola (approved):

jooola:
Next step would be to be able to run this trim on all show instance older than 1 week!

Thanks for the contribution!

caveman99:
Sheesh, that's for not paying attention when dinglast minute copy/pastechanges. Thanks for the fixes!

PR #2915, opened 2024-01-25
fix(legacy): avoid crash when lot of streams in configuration

mp3butcher:
### Description

lot of streams forbid http://libretime.example.org/preference/stream-setting to open (max=4)
this patch prevent the crash by displaying an error message and truncate streamcount to 4.


paddatrapper (approved):
LGTM

PR #2930, opened 2024-02-02
feat: playout - add internal_url parameter

unbelauscht:
This helps keeping **playout** and **libresoap** traffic internal, by specifying the **internal webservice** as the internal_url instead of the `public_url` (when behind a reverse proxy e.g.).

# Testing

I tested the change in an interactive Python 3.11.6 CLI. Unfortunately I wasn't able to build the containers myself successfully.

### DIFF
--- a/docker/config.template.yml
+++ b/docker/config.template.yml
@@ -100,6 +100,8 @@ email:
   key_file:
 
 playout:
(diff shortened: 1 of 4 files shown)


unbelauscht:
This would indeed work for me. Thank you very much 😊

Issue #2859, opened 2023-12-30
Playout volume too low and unstable


SergeDeMaseneer:
### Describe the bug

check "enable replay gain"
Move the slider to the right..... no change at all
![image](https://github.com/libretime/libretime/assets/14262795/b357536e-b36b-4e41-817f-ab292c0cb945)
I am still using version 3.1.0. but i have read some posts on discourse that it is still an issue.

### To reproduce

1. Go to Settings - Streams
2. Click on "enable replay gain" and make sure the checkbox is checked.  
3. Move the "replay gain modifier" to for example 10db
4. Listen to the stream ... nothing changes.
5. Also some songs play louder than others like the replay gain is not working at all

### Expected behavior

I expect all songs to be equally loud on the webstream.

### LibreTime version

3.1.0

### Installation method and OS / Environment

Operating system: Debian Bullseye
Method: Installer script

### Client Environment

Operating system: Debian Bullseye / Windows 11
Browser: Firefox developer edition Version 122.0b3 

### Screenshots

![image](https://github.com/libretime/libretime/assets/14262795/a67e54c0-788f-4897-9ecc-a21ab5295bce)

jooola:
I tested this myself and the replaygain is working on my side, so I cannot reproduce your issue. The currently playing song will not be updated, the replaygain changes will take effect on the next song in the queue.

SergeDeMaseneer:
I got some answers on discourse from people they are saying they have exactly the same issue. 
Is there anything I can check on my side? Is the replaygain somewhere mentioned in the log files? Are there services I can check?

jooola:
The replay gain follows this path:
https://github.com/libretime/libretime/blob/72268ad9bb1a96b24efda7143b9371d6fd98ca03/legacy/application/models/Schedule.php#L882

https://github.com/libretime/libretime/blob/72268ad9bb1a96b24efda7143b9371d6fd98ca03/legacy/application/models/Schedule.php#L905

https://github.com/libretime/libretime/blob/72268ad9bb1a96b24efda7143b9371d6fd98ca03/playout/libretime_playout/player/schedule.py#L143

https://github.com/libretime/libretime/blob/72268ad9bb1a96b24efda7143b9371d6fd98ca03/playout/libretime_playout/player/liquidsoap.py#L26-L27

[...]

PR #2936, opened 2024-02-05
bugfix(playout): fix ignored replay gain

mp3butcher:
### Description

fix https://github.com/libretime/libretime/issues/2859


mp3butcher:
see other pr

PR #2937, opened 2024-02-05
fix(playout): fix ignored replay_gain

mp3butcher:
fix [#2859](https://github.com/libretime/libretime/issues/2859)
according to doc :https://www.liquidsoap.info/doc-1.4.4/replay_gain.html
```
The replay gain metadata resolver is not enabled by default. You can do it by adding the following code to your script:
enable_replaygain_metadata()
```
## Test bug protocol
Setup and start a show
repeat until you hear a difference:
go to stream settings
set amplification to +10dB and save
restart libretime
listen volume
go to stream settings
set amplification to -10dB and save
restart libretime
listen volume difference (sometimes it require song to changed)

## Note
tested under liquidsoap 2.0 but should be the same for 1.4

### DIFF
--- a/playout/libretime_playout/liquidsoap/1.4/ls_script.liq
+++ b/playout/libretime_playout/liquidsoap/1.4/ls_script.liq
@@ -41,6 +41,8 @@ def create_source()
     source_id := !source_id + 1
 end
 
+enable_replaygain_metadata()
+
 create_source()
 create_source()
 create_source()


jooola:
This function is about enabling "computing-or-retrieving-replay-gain-information", but we already have the replay gain computed by the analyzer.

https://github.com/savonet/liquidsoap/blob/1b75452af731e6e1bf88b46554cc6d3ad9b14b4c/src/libs/replaygain.liq#L108-L134

I tested https://github.com/libretime/libretime/issues/2859 and could reproduce the issue. Did you manage to reproduce the issue before implementing this?

mp3butcher:
> Great, could you post the details of the reproduction on the issue so I can reproduce it myself?

Setup and start a show
repeat until you hear a difference:
   go to stream settings 
   set amplification to +10dB
   restart libretime
   listen volume
   go to stream settings 
   set amplification to -10dB
   restart libretime
   listen volume difference

   
> But I don't want to recompute the replay gain before we play the files.

You don't have to: metadata retrieved are the annotations you computed

caveman99:
If i understand the documentaton (https://www.liquidsoap.info/doc-1.4.4/replay_gain.html) correctly. enable_replaygain_metadata either fetches or computes the value, if it is not already stored in metadata. We compute it on import in the analyzer. The actual application is done with the amplify call. Our approach is to pass the replay_gain value to liquidsoap out-of-band, thus enable_replaygain_metadata is not needed, just the amplify call. if amplify does not work, there's a problem with parameter formatting during the passing operation.

mp3butcher:
I believed it's retrieving correct data with enable_replaygain_metadata because volume change when i touch replay gain settings.
So it's really the proof retrieved replaygain is an custom amplified data computed somewhere and not something computed by liquidsoap

mp3butcher:
@jooola  i haven'f figure func prototype are different in 1.4 and 2.0 ...I added compute=false but it fails wi

mp3butcher:
@jooola I think you're referring to an internal func the good prototype seems in [doc/content/liq/replaygain-playlist.liq](https://github.com/savonet/liquidsoap/blob/1b75452af731e6e1bf88b46554cc6d3ad9b14b4c/doc/content/liq/replaygain-playlist.liq#L4)

mp3butcher:
@jooola i tested with compute=false but CI test jobs fail. The compute argument  is implemented since 2023 https://github.com/savonet/liquidsoap/pull/3438/files and seems not present in 1.4

jooola:
I found a bugs around the replay gain settings, I'll open a PR in the coming days.

This is not the right solution, closing.

PR #2940, opened 2024-02-07
fix: support bulk upload when running from docker compose

kmahelona:
### Description
Bulk upload doesn't work when running in a docker-compose environment if the public_url is set to localhost. The fix here sets the api url to `LIBRETIME_GENERAL_PUBLIC_URL` environment variable if it exists. If it doesn't, then it defaults to the `settings.CONFIG.genera.public_url`. `LIBRETIME_GENERAL_PUBLIC_URL` is only set in docker compose (from what I can tell via search).

**This is a new feature**:
No

The documentation, docs/admin-manual/library.md, was updated to show the command for running bulk upload from docker,
```
docker-compose run --rm api libretime-api bulk_import --path PATH_THE_DIRECTORY_TO_SCAN
```

### Testing Notes
From main do,
```bash
make clean dev
docker-compose run --rm api libretime-api bulk_import --path PATH_THE_DIRECTORY_TO_SCAN
```
The command should fail with an connection error.

Checkout this PR, then run
```bash
make clean dev
docker-compose run --rm api libretime-api bulk_import --path PATH_THE_DIRECTORY_TO_SCAN
```
The command should run successfully if the given path has audio files.

### **Links**
https://matrix.to/#/!mkAnQsLsibojOwmqiw:one.ems.host/$vaVktWMuUVUmLdSmXDNbPTKg0ho09V_s2BdR6Uu9wd4?via=one.ems.host&via=matrix.org&via=freie-radios.de


kmahelona:
Not sure how to write a test for this, https://github.com/libretime/libretime/pull/2940#issuecomment-1932696642

api/libretime_api/storage/management/commands/bulk_import.py
```diff
@@ -52,7 +53,9 @@ def add_arguments(self, parser: CommandParser):
         )
 
     def handle(self, *args, **options):
-        url = settings.CONFIG.general.public_url
+        url = os.environ.get(
+            "LIBRETIME_GENERAL_PUBLIC_URL", settings.CONFIG.general.public_url
+        )
```
jooola, line 58: This is not needed, the config already handle env vars by itself.

docker-compose.yml
```diff
@@ -86,6 +86,8 @@ services:
     volumes:
       - ${LIBRETIME_CONFIG_FILEPATH:-./config.yml}:/etc/libretime/config.yml:ro
       - libretime_storage:/srv/libretime
+    environment:
+      LIBRETIME_GENERAL_PUBLIC_URL: http://nginx:8080
```
jooola, line 90: Hmm this feels risky, could we pass the env vars only when we run the bulk import ?

docs/admin-manual/library.md
```diff
@@ -13,4 +13,14 @@ To scan a directory and import the files into the library, you can use the follo
 sudo -u libretime libretime-api bulk_import --path PATH_THE_DIRECTORY_TO_SCAN
 ```
 
-See the command usage to get available options.
+If you're running from a docker environment, use
+```bash
+docker-compose run --rm api libretime-api bulk_import --path PATH_THE_DIRECTORY_TO_SCAN
```
jooola, line 18:
```suggestion
docker-compose run --rm -e LIBRETIME_GENERAL_PUBLIC_URL=http://nginx:8080 api libretime-api bulk_import --path PATH_THE_DIRECTORY_TO_SCAN
```

Ideally I would move the import process to python, but this might do the trick for now.

jooola (changes requested):

api/libretime_api/storage/management/commands/bulk_import.py
kmahelona: in my testing bulk import tries to call `http://localhost:8000/api` etc. but the docker container can't actually access that. It needs to use http://nginx:8080 instead to access the API to run the import.

docker-compose.yml
kmahelona: The suggestion below makes sense rather than changing the compose. The documentation helps ensure that the right URL is set.

api/libretime_api/storage/management/commands/bulk_import.py
kmahelona: `settings.CONFIG.general.public_url = http://localhost:8000/` is the default when you spin up the compose environment.

PR #2941, opened 2024-02-07
feat(legacy): custom intro/outro playlists

caveman99:
### Description

Having a global intro and outro playlist in settings is not very flexible for special programming. This adds an override intro/outro playlist per show. If it is not set, the global one is used. also it's ignored if there's no autloading at all.

...not yet, but i will.

### Testing Notes

**What I did:**
Schedule 2 shows, one without defining custom lists, one with defining custom lists. one hour before the show starts it should be populated correctly. If you define a global list it shojuld be replaced with the per-show list.


caveman99:
Remark: this was an absolute nightmare to implent through all layers and abstractions. I full heartedly support ditching the legacy UI.

legacy/application/models/airtime/map/CcPlaylistTableMap.php
```diff
@@ -56,6 +56,8 @@ public function buildRelations()
     {
         $this->addRelation('CcSubjs', 'CcSubjs', RelationMap::MANY_TO_ONE, array('creator_id' => 'id', ), 'CASCADE', null);
         $this->addRelation('CcShow', 'CcShow', RelationMap::ONE_TO_MANY, array('id' => 'autoplaylist_id', ), 'SET NULL', null, 'CcShows');
+        $this->addRelation('CcShow', 'CcShow', RelationMap::ONE_TO_MANY, array('id' => 'intro_playlist_id', ), 'SET NULL', null, 'CcShows');
+        $this->addRelation('CcShow', 'CcShow', RelationMap::ONE_TO_MANY, array('id' => 'outro_playlist_id', ), 'SET NULL', null, 'CcShows');
```
jooola, line 60: Please update the legacy schema.xml file and regenerate these files with the make propel-gen command. You will have to run a bunch of make -C legacy format and pre-commit run --all to fix the formatting.

api/libretime_api/schedule/models/show.py
```diff
@@ -69,6 +69,28 @@ def live_enabled(self) -> bool:
...
+    intro_playlist = models.ForeignKey(
+        "schedule.Playlist",
+        on_delete=models.DO_NOTHING,
+        blank=True,
+        null=True,
+        db_column="intro_playlist_id",
+        related_name="intro_playlist",
+    )
+
+    override_intro_playlist = models.BooleanField(db_column="override_intro_playlist")
```
jooola, line 81: Why use a boolean field AND a nullable ID? Can't we consider a null ID as "disabled" value?

legacy/application/forms/AddShowAutoPlaylist.php
```diff
@@ -34,6 +34,36 @@ public function init()
...
+        $this->addElement('checkbox', 'add_show_override_intro_playlist', [
+            'label' => _('Override Intro Playlist ?'),
+            'required' => false,
+            'class' => 'input_text',
+            'decorators' => ['ViewHelper'],
+        ]);
+
+        $introPlaylistSelect = new Zend_Form_Element_Select('add_show_intro_playlist_id');
+        $introPlaylistSelect->setLabel(_('Select Intro Playlist'));
+        $introPlaylistSelect->setMultiOptions(Application_Model_Library::getPlaylistNames(true));
```
jooola, line 48: I don't know how expensive `Application_Model_Library::getPlaylistNames(true)` is, but maybe we can store the result in a variable and not call the DB twice?

legacy/application/models/Show.php
```diff
@@ -169,6 +169,58 @@ public function setAutoPlaylistId($playlistid)
         $show->setDbAutoPlaylistId($playlistid);
     }
 
+    public function getHasOverrideIntroPlaylist()
+    {
+        $show = CcShowQuery::create()->findPK($this->_showId);
+
+        return $show->getDbOverrideIntroPlaylist();
+    }
```
jooola, line 177: This feels expensive for a single get, I wonder if the show is not already prefilled, or maybe we should cache the show in a property.

legacy/application/models/airtime/map/CcPlaylistTableMap.php
caveman99: Thanks for the hint, i was not aware these were generated instead of modelled by hand.

api/libretime_api/schedule/models/show.py
caveman99: This comes from the UI logic in the show builder. the has_autoplaylist works the same way (and hides parts of the UI when unchecked). I will change though, since it saves 2 clicks when editing.

jooola:
Have you tested this with smartblocks? Is there any chances that the final show length is way of because didn't account for the added intro/outro duration? Or was this already implemented in the initial work for the global intro/outro?

caveman99:
> Have you tested this with smartblocks? Is there any chances that the final show length is way of because didn't account for the added intro/outro duration? Or was this already implemented in the initial work for the global intro/outro?

This is already implemented in the existing logic. This just swaps out the ID's of the intro and outro playlist before processing if a custom one is defined. The only real change needed for this feature is the one in `legacy/application/common/AutoPlaylistManager.php`. The rest ist UI and DB plumbing.

caveman99:
Is there something required on my side to get this moving again? :-) I know it's a huge patch but most of this is database changes to add the two fields to store the info, partly auto-generated.

UI looks like this after the change (unused parts are folded away)

grafik

paddatrapper (approved):

paddatrapper (commented):
This LGTM, can you please update the docs?

caveman99:
@paddatrapper docs added and a minor fix to the screen display. I forgot the checkboxes from my screenshot were gone with a mod, so only the dropdowns remain. This change clarifies that 'None' means 'take the default from the global settings'.

Issue #2670, opened 2023-08-23
Autoloading playlists failing randomly


matias-tecnosoul:
### Describe the bug

We had random failures with “autoloading playlists” not being loaded, leaving the shows without tracks.

The shows are  working fine with the lists being autoloaded correctly but they've failed a couple of time in the last week.
We tried recreating the blocks and lists but the issue happened again.
I've asked in the [forum](https://discourse.libretime.org/t/autoloading-playlists-failing-randomly/1547) and one other person had the issue as well.
Thanks!!

### To reproduce

We couldn't find how to reproduce the issue. The issue happened in 2 LT installs on different machines. One install is fresh and the other is an upgrade from older versions.  On the other hand it didn´t happen in 2 other installs, both upgraded.

### Expected behavior

Shows being autoloaded from the configured autoload playlists.

### Relevant log output or error messages

```shell
The error ocurred in the 20:00hs show.

Liquidsoap logs

2023/08/21 19:56:45 [s3:3] Prepared "/var/lib/libretime/playout/scheduler/418.mp3" (RID 0).
2023/08/21 19:56:45 [lang:3] timeout --signal=KILL 45 libretime-playout-notify media '14899' &
2023/08/21 19:56:45 [lang:3] Using message format 0
2023/08/21 19:56:45 [lang:3] Using message format 0
2023-08-21 19:56:45,713 | INFO     | libretime_playout.notify.main:media:72 - Sending currently playing media id '14899'
2023/08/21 19:59:59 [cue_cut_9618:3] Cueing out...
2023/08/21 19:59:59 [dummy:3] Source failed (no more tracks) stopping output...
2023/08/21 19:59:59 [dummy(dot)2:3] Source failed (no more tracks) stopping output...
2023/08/21 19:59:59 [switch:blank+schedule:3] Switch to map_metadata:offline with forgetful transition.
2023/08/21 19:59:59 [lang:3] transition called...
2023/08/21 20:00:00 [server:3] New client: 127.0.0.1.
2023/08/21 20:00:00 [server:3] Client 127.0.0.1 disconnected.
2023/08/21 20:02:11 [server:3] New client: 127.0.0.1.
2023/08/21 20:02:11 [lang:3] web_stream.get_id
2023/08/21 20:02:11 [server:3] Client 127.0.0.1 disconnected.
2023/08/21 20:08:51 [server:3] New client: 127.0.0.1.

Playout logs

[...]

PR #2946, opened 2024-02-08
feat: build schedule events exclusively in playout

jooola:
### Description

Build and use the schedule events only in playout, the events generated by legacy are not used anymore.

This ensure that we don't have to maintain 2 different implementation in 2 different languages. We still need the php function to run to make sure the side effects of this function are executed (filling the schedule in the DB).


paddatrapper:
Is this PR affected by https://github.com/libretime/libretime/issues/2670#issuecomment-2028035833?

jooola:
I would have to test, I can deploy this to test it out.

jooola:
I have been running this in production, combined with #2987 and we didn't have any blanks.

PR #2954, opened 2024-02-15
docs: get libfdk-aac from non-free source for debian 11

caveman99:
This adds reference to a second file needed for debian 11.

### DIFF
--- a/docs/admin-manual/tutorials/setup-liquidsoap-aac-streaming.md
+++ b/docs/admin-manual/tutorials/setup-liquidsoap-aac-streaming.md
@@ -29,8 +29,8 @@ wget https://github.com/savonet/liquidsoap/releases/download/v1.4.4/liquidsoap-v
 For Debian 11 ('Bullseye'), first enable non-free package source for libfdk-aac support:
 
 ```bash
-sudo sed -i 's/main/main non-free/' /etc/apt/sources.list
-sudo apt update
+sudo apt install software-properties-common
+sudo apt-add-repository -c non-free
 ```
 
 Then use the following file:
--- a/docs/admin-manual/tutorials/setup-liquidsoap-aac-streaming.md
+++ b/docs/admin-manual/tutorials/setup-liquidsoap-aac-streaming.md
@@ -6,7 +6,7 @@ This tutorials walks you though the steps required to replace the liquidsoap pac
 
 :::warning
 
-Replacing the liquidsoap package has security implications, since this will remove the package from the system's package manager. This means that the package manager will not be able to update the liquidsoap package in the future. This includes backports of security fixes.
+Replacing the liquidsoap package has security implications, since this will remove the package from the system's package manager. This means that the package manager will not be able to update the liquidsoap package (and also the necessary fdk-aac1 lib on Debian 11) in the future. This includes backports of security fixes.
 
 Libretime is NOT compatible with Liquidsoap 2.x at the time of this writing. Future versions of Libretime will support Liquidsoap 2.x which will render these instructions obsolete.
 
@@ -35,10 +35,23 @@ wget http://ftp.debian.org/debian/pool/non-free/f/fdk-aac/libfdk-aac1_0.1.6-1_am
 
 ## 2. Install and replace the liquidsoap package
 
-Install the package(s) using `apt`, then remove the old liquidsoap dependencies. Make sure there are no other deb files in the directory before running the following commands.
+Install the package(s) using `apt`:
+
+For Ubuntu 20.04 LTS ('focal'), use the following command:
 
 ```bash
-sudo apt -y install ./*.deb
+sudo apt -y install ./liquidsoap-v1.4.4_1.4.4-ubuntu-focal-amd64-1_amd64.deb
+```
+
+For Debian 11 ('Bullseye'), use the following command:
+
+```bash
+sudo apt -y install ./liquidsoap-v1.4.4_1.4.4-debian-stable-amd64-1_amd64.deb ./libfdk-aac1_0.1.6-1_amd64.deb
+```
+
+Then remove the old liquidsoap dependencies on both distros.
+
+```bash
 sudo apt -y autoremove
 ```
 
--- a/docs/admin-manual/tutorials/setup-liquidsoap-aac-streaming.md
+++ b/docs/admin-manual/tutorials/setup-liquidsoap-aac-streaming.md
@@ -26,18 +26,19 @@ For Ubuntu 20.04 LTS ('focal'), use the following file:
 wget https://github.com/savonet/liquidsoap/releases/download/v1.4.4/liquidsoap-v1.4.4_1.4.4-ubuntu-focal-amd64-1_amd64.deb
 ```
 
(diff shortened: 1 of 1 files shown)


docs/admin-manual/tutorials/setup-liquidsoap-aac-streaming.md
```diff
@@ -26,18 +26,19 @@ For Ubuntu 20.04 LTS ('focal'), use the following file:
 wget https://github.com/savonet/liquidsoap/releases/download/v1.4.4/liquidsoap-v1.4.4_1.4.4-ubuntu-focal-amd64-1_amd64.deb
 ```
 
-For Debian 11 ('Bullseye'), use the following file:
+For Debian 11 ('Bullseye'), use the following files: (Debian 11 ships fdk-aac2, so you'll need to download fdk-aac1 from Debian 10)
```
paddatrapper, line 29: This is not a good recommendation, though I'm not sure what the alternative is. For 1 you risk making a [FrankenDebian](https://wiki.debian.org/DontBreakDebian#Don.27t_make_a_FrankenDebian) and 2 you then don't get security updates for this library.

docs/admin-manual/tutorials/setup-liquidsoap-aac-streaming.md
```diff
@@ -26,18 +26,19 @@ For Ubuntu 20.04 LTS ('focal'), use the following file:
...
 ```
 
 ## 2. Install and replace the liquidsoap package
 
-Install the package using `apt`, then remove the old liquidsoap dependencies:
+Install the package(s) using `apt`, then remove the old liquidsoap dependencies. Make sure there are no other deb files in the directory before running the following commands.
 
 ```bash
-sudo apt -y install ./liquidsoap-v1.4.4_1.4.4-*-amd64-1_amd64.deb
+sudo apt -y install ./*.deb
```
paddatrapper, line 41: Rather explicitly list both debs

docs/admin-manual/tutorials/setup-liquidsoap-aac-streaming.md
caveman99: It's the same point we have with the main liquidsoap binary. We already have red boxes all over this doc saying it is unsupported, and if it breaks you own both parts. I just stumbled upon this today cause i had to do this on debian proper for the first time. I don't see the real difference between pulling in a 3rd party deb and an oldoldstable one. Both means you are responsible for the result.

docs/admin-manual/tutorials/setup-liquidsoap-aac-streaming.md
caveman99: yes, i was hesitant there, the brevity talked me into doing it this way. will reconsider. :-)

docs/admin-manual/tutorials/setup-liquidsoap-aac-streaming.md
caveman99: I added a comment to the caution.

docs/admin-manual/tutorials/setup-liquidsoap-aac-streaming.md
caveman99: Separated between debian and ubuntu

docs/admin-manual/tutorials/setup-liquidsoap-aac-streaming.md
```diff
@@ -26,18 +26,32 @@ For Ubuntu 20.04 LTS ('focal'), use the following file:
 wget https://github.com/savonet/liquidsoap/releases/download/v1.4.4/liquidsoap-v1.4.4_1.4.4-ubuntu-focal-amd64-1_amd64.deb
 ```
 
-For Debian 11 ('Bullseye'), use the following file:
+For Debian 11 ('Bullseye'), use the following files: (Debian 11 ships fdk-aac2, so you'll need to download fdk-aac1 from Debian 10)
```
jooola, line 29: Could link me where you found out that fdk-aac1 is required instead of fdk-aac2 ? I've used the one in the non-free repository without any problem.

Why did you not propose to enable the non free repository ?

caveman99: The error message when installing the liquidsoap 1.4.4 deb prompted me. I just tried installing libfdk-aac2 instead of 1, but as soon as i try to uninstall 1 it wants to remove the new liquidsoap as well.

docs/admin-manual/tutorials/setup-liquidsoap-aac-streaming.md
```diff
@@ -26,18 +26,32 @@ For Ubuntu 20.04 LTS ('focal'), use the following file:
...
 
 ```bash
 wget https://github.com/savonet/liquidsoap/releases/download/v1.4.4/liquidsoap-v1.4.4_1.4.4-debian-stable-amd64-1_amd64.deb
```
jooola, line 32: I believe there is a mistake here, we should use the deb package for `testing` as bullseye 11 was testing at the time of the release of this package. This might be a reason why fdk-aac1 is required, as it would be for the older version of Debian.

jooola: https://github.com/savonet/liquidsoap/releases/download/v1.4.4/liquidsoap-v1.4.4_1.4.4-debian-testing-amd64-1_amd64.deb

caveman99: Ah, ok, that makes sense. I will make a quick test on the instance i have up and running and alter the docs accordingly.

jooola:
I believe the missing part is to enable the non-free repository in Debian (and probably also Ubuntu) to have access to the ` fdk-aac2` package.

caveman99:
the ubuntu focal deb links against libfdk-aac1 (tested with ldd on a running system) which lives in multiverse. On my LTS server install the repository is enabled by default. For debian you are right. Turns out i installed on a provider image that had non-free already enabled.

docs/admin-manual/tutorials/setup-liquidsoap-aac-streaming.md
```diff
@@ -26,10 +26,17 @@ For Ubuntu 20.04 LTS ('focal'), use the following file:
 wget https://github.com/savonet/liquidsoap/releases/download/v1.4.4/liquidsoap-v1.4.4_1.4.4-ubuntu-focal-amd64-1_amd64.deb
 ```
 
-For Debian 11 ('Bullseye'), use the following file:
+For Debian 11 ('Bullseye'), first enable non-free package source for libfdk-aac support:
 
 ```bash
-wget https://github.com/savonet/liquidsoap/releases/download/v1.4.4/liquidsoap-v1.4.4_1.4.4-debian-stable-amd64-1_amd64.deb
+sudo sed -i 's/main/main non-free/' /etc/apt/sources.list
+sudo apt update
```
jooola, line 33:
```suggestion
sudo apt install software-properties-common
sudo apt-add-repository -c non-free
```

PR #2955, opened 2024-02-16
fix(legacy): allow updating track types code

caveman99:
This is a minor nag i had. Since moving to an indexed track_type model for 3.0 the constraint of the shortlabel being immutable is no longer valid, since it's not used as index any more. This commit removes the readonly flag from that form field.

### DIFF
--- a/legacy/public/js/airtime/tracktype/tracktype.js
+++ b/legacy/public/js/airtime/tracktype/tracktype.js
@@ -15,12 +15,6 @@ function populateForm(entries) {
   $("#visibility").val(visibility_value);
 
   $("#analyze_cue_points").prop("checked", entries.analyze_cue_points);
-
-  if (entries.id.length != 0) {
-    $("#code").attr("readonly", "readonly");
-  } else {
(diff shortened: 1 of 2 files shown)


legacy/public/js/airtime/tracktype/tracktype.js
```diff
@@ -156,6 +150,7 @@ function initTracktypeData() {
     description: "",
     visibility: "1",
     id: "",
+    analyze_cue_points: true,
```
jooola, line 153: This seem unrelated ?

caveman99: Err yes. This is part of a bigger thing i am working on. However this field is not initialized in the code as-is. Default for this is 'true' in the schema, so i added it.

jooola: Please keep this is the "bigger thing" you are working on.

caveman99: Removed, sorry.

jooola (approved):

PR #2960, opened 2024-02-20

caveman99:
Description

Testing Notes

Playground server with new code successfully executed DELETE api call.

### DIFF
--- a/api/libretime_api/storage/views/file.py
+++ b/api/libretime_api/storage/views/file.py
@@ -1,13 +1,13 @@
 import logging
 import os
+from os import remove
 
 from django.conf import settings
 from django.http import HttpResponse
-from django.shortcuts import get_object_or_404
 from django.utils.encoding import filepath_to_uri
-from rest_framework import viewsets
+from rest_framework import status, viewsets
 from rest_framework.decorators import action
-from rest_framework.serializers import IntegerField
+from rest_framework.exceptions import APIException
 
 from ...schedule.models import Schedule
 from ..models import File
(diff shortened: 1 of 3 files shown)


jooola:
We are missing an error schema for our API error responses. Returning a simple HTTP status code is not enough.

api/libretime_api/storage/views/file.py
```diff
@@ -29,3 +34,24 @@ def download(self, request, pk=None):  # pylint: disable=invalid-name
...
+
+    def destroy(self, request, *args, **kwargs):  # pylint: disable=invalid-name
+        pk = kwargs.get("pk", None)
+        pk = IntegerField().to_internal_value(data=pk)
+
+        file = get_object_or_404(File, pk=pk)
+
+        # Check if the file is scheduled to be played in the future
+        if Schedule.is_file_scheduled_in_the_future(file_id=pk) is True:
+            return HttpResponse(status=409)
```
paddatrapper, line 46: Let's include a useful error message in these responses

paddatrapper:
> We are missing an error schema for our API error responses. Returning a simple HTTP status code is not enough.

Do we want to define it here or merge this with basic human-readable errors for now and spend time defining it outside this?

jooola:
We can implement a custom exceptions handler in the future, but I refactored the code to raise API exceptions instead of returning empty responses.

@caveman99 Let me know if you are ok with this implementation, and I'll be happy to merge this.

caveman99:
@jooola looks good to me ... :-)

PR #2995, opened 2024-04-17
fix(legacy): allow deleting file with api token

caveman99:
when calling DELETE /rest/media/<id> the call fails with 'unknown error' if it's not within a GUI session. The StoredFile delete method checks for user permissions regardless of if a user is even known.

### DIFF
--- a/legacy/application/models/StoredFile.php
+++ b/legacy/application/models/StoredFile.php
@@ -396,6 +396,7 @@ SQL;
 
         // if we get here from the REST API, there's no valid user. APIKEY is validated already.
         if ($userInfo = Zend_Auth::getInstance()->getStorage()->read()) {
+            // This call will throw "Trying to get property 'id' of non-object"
             $user = new Application_Model_User($userInfo->id);
             $isAdminOrPM = $user->isUserType([UTYPE_SUPERADMIN, UTYPE_ADMIN, UTYPE_PROGRAM_MANAGER]);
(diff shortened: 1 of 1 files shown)


caveman99:
Remark: i am not sure if this is a security risk, but i don't see any other path to this point. you either have an authenticated session or a vaild REST key.

legacy/application/models/StoredFile.php
```diff
@@ -394,16 +394,18 @@ public function delete($quiet = false)
...
+        if ($userInfo = Zend_Auth::getInstance()->getStorage()->read()) {
+            // This call will throw "Trying to get property 'id' of non-object"
+            $user = new Application_Model_User($userInfo->id);
+            $isAdminOrPM = $user->isUserType([UTYPE_SUPERADMIN, UTYPE_ADMIN, UTYPE_PROGRAM_MANAGER]);
+            if (!$isAdminOrPM && $this->getFileOwnerId() != $user->getId()) {
+                throw new FileNoPermissionException();
+            }
+            $file_id = $this->_file->getDbId();
+            Logging::info($file_id);
+            Logging::info('User ' . $user->getLogin() . ' is deleting file: ' . $this->_file->getDbTrackTitle() . ' - file id: ' . $file_id);
```
paddatrapper, line 407: Shall we log a delete info message when deleted via API token, just without user information?

jooola (approved):

PR #3011, opened 2024-05-14
Fix container startup order in docker-compose.yml

olivermeyer:
### Description

This PR fixes the startup order of containers in the Docker deployment.

The initial issue I was having was that the `playout` container sometimes timed out while trying to get the liquidsoap version from the `liquidsoap` container. Restarting the `playout` container fixed it, because by that time `liquidsoap` was ready, but it's less than ideal. To fix this, I suggest adding a healthcheck to `liquidsoap`, and adding it to the `playout`'s dependencies with the `service_healthy` condition.

While looking into this, I noticed that logs are often noisy at startup because containers are using the default dependency mode (`container_started`) and ignoring healthchecks. To improve this, I'm switching all dependencies to `service_healthy`, except for the dependency from `legacy` to `nginx` since `legacy` doesn't have a healthcheck.

The downside of this approach is that startup seems to take longer, because containers actually wait for upstream services to be running, rather than just waiting for them to start. I don't think this has an _actual_ impact on startup time though, because downstream services would not work until upstream services are running anyway.

### Testing Notes

**What I did:**

I applied these changes to my deployment a week ago and I've been monitoring it since. The issue with the `playout` container hasn't occurred since, and the logs are cleaner on startup.

`docker compose up` and check that everything is running.


docker-compose.yml
```diff
@@ -41,20 +44,24 @@ services:
...
-      - rabbitmq
+      rabbitmq:
+        condition: service_healthy
     volumes:
       - ${LIBRETIME_CONFIG_FILEPATH:-./config.yml}:/etc/libretime/config.yml:ro
       - libretime_playout:/app
     environment:
       LIBRETIME_GENERAL_PUBLIC_URL: http://nginx:8080
+    healthcheck:
+      test: liquidsoap --version
```
paddatrapper, line 55: This isn't a useful check to see if liquidsoap is running because this will return true independently of the actual running process. This verifies that the liquidsoap binary in the container exists and at least somewhat runs, but doesn't check the specific process that libretime-liquidsoap spawns.

LibreTime liquidsoap is a fairly simple script that starts a liquidsoap process and then communicates with it via telnet. A better check would be to verify that the file `radio.liq` exists in the working directory ([created](https://github.com/libretime/libretime/blob/5d2ad4303938055457654e9ab85318c1dccbe2ab/playout/libretime_playout/liquidsoap/main.py#L43-L52) as part of the startup process) and potentially that the liquidsoap process is running/can be communicated with via telnet.

olivermeyer: Interesting - `liquidsoap --version` seemed like a decent proxy for the container being ready for two reasons:
* With that healthcheck, the liquidsoap container only reached `healthy` state ~30s after creation
* With the dependency between liquidsoap and playout, I didn't see any of these error logs `WARNING  | libretime_playout.liquidsoap.client._client:wait_for_version:62 - could not get version: [Errno 111] Connection refused` anymore, which is the behaviour I'm trying to prevent

That being said, I changed the healthcheck to test that the `radio.liq` file exists and see the same behaviour.

jooola: I believe there is a 'on started' hook (to comfirm) that we can use to actually use to toggle a ready state. Let it be a file or a http endpoint with a variable, not sure which is cleaner.

This might be useful from playout as well to know when liquidsoap ready is.

olivermeyer: > I believe there is a 'on started' hook (to comfirm) that we can use to actually use to toggle a ready state

I just checked the liquidsoap docs and found this `on_start` [hook](https://www.liquidsoap.info/doc-1.4.4/reference.html#on_start). I guess that's the one you're referring to. I can see how using it might be the most robust implementation, but I personally don't have the bandwidth to figure out how to do it right now.

Perhaps the current implementation is good enough? It's an improvement over no dependency at all and, as far as I can tell, it solves the immediate issue of playout failing to start because liquidsoap isn't ready.

Curious to hear your thoughts :)

paddatrapper: This is a good first step. I agree that `on_start` is probably more robust, but can be implemented later down the line

jooola:
Thanks a lot for the contribution, but I am unsure this is the fix for the initial problem. If the playout service has a timeout during the initialization phase, we should probably handle this problem first.

The services should be resilient to a missing dependencies (startup or crash), and have a built-in retry mechanism.

@olivermeyer If we fix the playout init problem, do you think the dependency chain is still worth adding ?

dakriy:
Running libretime in a docker swarm environment has issues because nginx tries to resolve a container that does not exist before it starts, and so it caches in invalid DNS name, and the whole legacy app won't load until the nginx container is restarted. Docker swarm does not have the `depends_on`  property and everything starts at once. Would this possibly fix that issue?

PR #3015, opened 2024-05-15
feat(legacy): add filename block criteria

### LINKED ISSUE #3014: Filter tracks by filename

Filename can have useful information to filter by. Especially when none of the ID3 tags are applicable and description doesn't feel right. Required for a radio station to go live with LibreTime.

Add a track filter to filter by original filename.

Using the description or other ID3 tags but it just doesn't feel right to use an ID3 tag randomly for something it was not meant for.

### Additional context

_No response_

dakriy:
### Description

Adds filename block criteria because sometimes you just need the filename and the ID3 tags won't do. Also attempted to clean up some of the criteria data because all the duplication and copy paste was driving me crazy.

**This is a new feature**:

Adds filename block criteria

**I not have updated the documentation to reflect these changes**:

I did not feel any documentation update was needed as I did not find each individual criteria documented and this is just adding another criteria. It should also be fairly self explanatory.

### Testing Notes

**What I did:**

I started up a LibreTime docker stack and uploaded several mp3 files. I then created a block and added a filename filter to filter to one of the mp3 files and verified that the filename filter worked as expected. 

### **Links**

Closes  #3014


jooola:
The refactor is awesome! Thanks for the work!

I was expecting the filename criteria to have different modifiers, for example contains or start with ? Seem like we have numerical modifiers?

image

jooola:
Also, we might want to order the criteria a bit, not sure how though. Maybe it's best for another PR.

dakriy:
@jooola Thanks! I was worried you might not like the refactor.

As for ordering, if you have an ordering you want I can sneak it in real quick, but if there needs to be more discussion then I would prefer to do it in a different PR. 

Not sure why you are seeing different filters. I have tried it on multiple computers and this is what I see:
image

Do you need to re-run composer autoload files or something? Although that wouldn't make much sense for JS assets.

dakriy:
Maybe it was the assets file checksum. I updated that so hopefully that fixes your issue?

jooola:
Tested this again, indeed, might have been related to the assets checksum.

Merging now.

RobertWaterloo:
Being able to create a Smartblock based on filename is great, but we need to know the filenames and paths. Can they be made visible in the track editing screen as a non-editable field?

(does this need a new entry as a feature request?)

--Bob.

RobertWaterloo:
Being able to create a Smartblock based on filename is great, but we need to know the filenames and paths. Can they be made visible in the track editing screen as a non-editable field?

(does this need a new entry as a feature request?)

-Bob.

dakriy:
I'll make a PR to fix the filename thing. In the mean time you can use `contains`.

PR #3019, opened 2024-05-21
feat(legacy): implement subset sum solution to show scheduling

### LINKED ISSUE #3018: Reduce dead air time at end of block scheduling

When running a radio station it is generally a good idea to reduce dead air time. The current algorithm for adding tracks to a block/show can leave a lot of dead air time at the end as it doesn't use a very good algorithm. Required for a radio station to go live with LibreTime.

Adding tracks to a show until it is full while making it as full as possible is a well known problem in computer science. It is the [Subset Sum Problem](https://en.wikipedia.org/wiki/Subset_sum_problem). There is a decent solution on Wikipedia, why not use it?

Leaving dead air time. Not so great.

### Additional context

_No response_

dakriy:
### Description

When running a radio station it is generally a good idea to reduce dead air time. The current algorithm for adding tracks to a block/show can leave a lot of dead air time at the end as it doesn't use a very good algorithm. Adding tracks to a show until it is full while making it as full as possible is a well known problem in computer science. It is the [Subset Sum Problem](https://en.wikipedia.org/wiki/Subset_sum_problem). This PR implements a Randomized Greedy with Local Improvement (RGLI) approximation solution for the Subset Sum Problem. The new algorithm is only used when sort type is random and overflow is not enabled and there is no limit on the number of tracks that can be used.

**This is a new feature**:

Improvement on an existing feature. 

**I have not updated the documentation to reflect these changes**:

I did not update the documentation because the current scheduling algorithm is not currently documented and its existing features have not changed.

### Testing Notes

**What I did:**

I first attempted a fully polynomial time approximation scheme solution, however it is really bad at finding good solutions for high density values and can kinda slow the more tracks/time you have. So I instead implemented an RGLI which is O(nlogn) and has been giving much better results. 

I implemented the solution in a separate project and tested it and timed the values with a normal distribution of 500 songs with a mean of 3 minutes and a standard deviation of 1 minute. With a show size of 1 hour the algorithm took around 10-15 ms to run. When adjusting the block size and track size the algorithm still was pretty quick to run. Am going to be testing on an instance with lots of tracks later, will update PR when I have done that.

### **Links**

Closes #3018 

https://en.wikipedia.org/wiki/Subset_sum_problem#Fully-polynomial_time_approximation_scheme

https://web.stevens.edu/algebraic/Files/SubsetSum/przydatek99fast.pdf

### DIFF
--- a/legacy/application/models/Block.php
+++ b/legacy/application/models/Block.php
@@ -1333,89 +1333,83 @@ SQL;
         $info = $this->getListofFilesMeetCriteria($show);
         $files = $info['files'];
         $limit = $info['limit'];
-        $repeat = $info['repeat_tracks'];
-        $overflow = $info['overflow_tracks'];
-
-        $insertList = [];
-        $totalTime = 0;
-        $totalItems = 0;
+        $repeat = $info['repeat_tracks'] == 1;
+        $overflow = $info['overflow_tracks'] == 1;
+        $isRandomSort = $info['sort_type'] == 'random';
+        $blockTime = $limit['time'];
+        $blockItems = $limit['items'];
 
         if ($files->isEmpty()) {
-            return $insertList;
+            return [];
         }
 
         // this moves the pointer to the first element in the collection
         $files->getFirst();
         $iterator = $files->getIterator();
 
-        $isBlockFull = false;
-
-        while ($iterator->valid()) {
+        /**
+         * @var Track[] $tracks
+         */
+        $tracks = [];
+        $maxItems = 500;
+        while ($iterator->valid() && count($tracks) < $maxItems) {
             $id = $iterator->current()->getDbId();
             $fileLength = $iterator->current()->getCueLength();
             $length = Application_Common_DateHelper::calculateLengthInSeconds($fileLength);
-            // if the block is setup to allow the overflow of tracks this will add the next track even if it becomes
-            // longer than the time limit
-            if ($overflow == 1) {
-                $insertList[] = ['id' => $id, 'length' => $length];
-                $totalTime += $length;
-                ++$totalItems;
-            }
-            // otherwise we need to check to determine if the track will make the playlist exceed the totalTime before
-            // adding it this could loop through a lot of tracks so I used the totalItems limit to prevent
-            // the algorithm from parsing too many items.
 
-            else {
-                $projectedTime = $totalTime + $length;
-                if ($projectedTime > $limit['time']) {
-                    ++$totalItems;
-                } else {
-                    $insertList[] = ['id' => $id, 'length' => $length];
-                    $totalTime += $length;
-                    ++$totalItems;
-                }
-            }
-            if ((!is_null($limit['items']) && $limit['items'] == count($insertList)) || $totalItems > 500 || $totalTime > $limit['time']) {
-                $isBlockFull = true;
-
-                break;
-            }
+            $tracks[] = new Track($id, $length);
 
             $iterator->next();
         }
 
-        $sizeOfInsert = count($insertList);
+        /**
+         * @var Track[] $insertList
+         */
+        $insertList = [];
+        $totalTime = 0;
 
-        // if block is not full and repeat_track is check, fill up more
-        // additionally still don't overflow the limit
-        while (!$isBlockFull && $repeat == 1 && $sizeOfInsert > 0) {
-            Logging::debug('adding repeated tracks.');
-            Logging::debug('total time = ' . $totalTime);
-
-            $randomEleKey = array_rand(array_slice($insertList, 0, $sizeOfInsert));
-            // this will also allow the overflow of tracks so that time limited smart blocks will schedule until they
-            // are longer than the time limit rather than never scheduling past the time limit
-            if ($overflow == 1) {
-                $insertList[] = $insertList[$randomEleKey];
-                $totalTime += $insertList[$randomEleKey]['length'];
-                ++$totalItems;
-            } else {
-                $projectedTime = $totalTime + $insertList[$randomEleKey]['length'];
-                if ($projectedTime > $limit['time']) {
-                    ++$totalItems;
+        if ($isRandomSort && !$overflow && $blockItems === null) {
+            $minTrackLength = min(array_map(fn(Track $item) => $item->length, $tracks));
+            do {
+                $solution = SSPSolution::solve($tracks, 0.2, $blockTime);
+                $insertList = array_merge($insertList, $solution->tracks);
+                $totalTime += $solution->sum;
+            } while ($repeat && ($blockTime - $totalTime) > $minTrackLength);
+            shuffle($insertList);
+        } else {
+            $isFull = fn() => $blockItems !== null && count($insertList) >= $blockItems || $totalTime > $blockTime;
+            $addTrack = function (Track $track) use ($overflow, $blockTime, &$insertList, &$totalTime) {
+                if ($overflow) {
+                    $insertList[] = $track;
+                    $totalTime += $track->length;
                 } else {
-                    $insertList[] = $insertList[$randomEleKey];
-                    $totalTime += $insertList[$randomEleKey]['length'];
-                    ++$totalItems;
+                    $projectedTime = $totalTime + $track->length;
+
+                    if ($projectedTime <= $blockTime) {
+                        $insertList[] = $track;
+                        $totalTime += $track->length;
+                    }
                 }
+            };
+
+            foreach ($tracks as $track) {
+                $addTrack($track);
+
+                if ($isFull()) break;
             }
 
-            if ((!is_null($limit['items']) && $limit['items'] == count($insertList)) || $totalItems > 500 || $totalTime > $limit['time']) {
-                break;
+            $sizeOfInsert = count($insertList);
+
+            while (!$isFull() && $repeat && $sizeOfInsert > 0) {
+                Logging::debug('adding repeated tracks.');
+                Logging::debug('total time = ' . $totalTime);
+                $track = $insertList[array_rand(array_slice($insertList, 0, $sizeOfInsert))];
+
+                $addTrack($track);
             }
         }
 
-        return $insertList;
+        return array_map(fn(Track $track) => ['id' => $track->id, 'length' => $track->length], $insertList);
     }
 
     /**
@@ -1790,7 +1784,7 @@ SQL;
         try {
             $out = $qry->setFormatter(ModelCriteria::FORMAT_ON_DEMAND)->find();
 
-            return ['files' => $out, 'limit' => $limits, 'repeat_tracks' => $repeatTracks, 'overflow_tracks' => $overflowTracks, 'count' => $out->count()];
+            return ['files' => $out, 'limit' => $limits, 'repeat_tracks' => $repeatTracks, 'overflow_tracks' => $overflowTracks, 'count' => $out->count(), 'sort_type' => $sortTracks];
         } catch (Exception $e) {
             Logging::info($e);
         }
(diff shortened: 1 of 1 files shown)


dakriy:
Switched to RGLI as the fully polynomial approximation scheme gave unwanted results when input tracks were highly clustered around the same length. The RGLI is also giving better results at much faster speeds.

dakriy:
@jooola ran tests locally so hopefully it'll pass now

dakriy:
@jooola The build failed on a file that I never even touched, and correct me if I'm wrong but I feel fixing typos in the docs is out of scope for this PR. I can fix it if you want but I would like to avoid any merge conflicts. I did pull in the most recent master though.

jooola:
Yes, I'll handle this. Do not lose too much time :)

dakriy:
Ok thanks! Was there anything else you needed me to do for this PR?

paddatrapper:
Fixed in #3027. Can you merge/rebase on `main`?

dakriy:
Merged @paddatrapper

jooola:
Would it be possible to have a parameter to overshoot the total length of the block?

Based on the abstract of the document, I understand that the algorithm's goal is to find a subset of tracks with a total closest to the total length of the block, but never greater. But I'd prefer to always overshoot it, so we can always fade out a track if needed.

Ideally, we want to have the length closest to the block length, and always greater than the block length.

jooola:
Also, I wonder how random this algorithm actually is? I'd prefer the best randomness and overshoot than be closer, but always pick the same tracks.

I didn't fully understand the algorithm yet, but I understood that it tries to add the longest available tracks to the list, and if it doesn't fit, try with a shorter track, this until the total duration if filled. How can we make sure that the added tracks will always be as random as a simpler algorithm?

dakriy:
@jooola sorry for the incoming wall of text. Please let me know if you want me to run any other tests or have any questions about or don't understand any of this.

## Overshoot
For the “overshoot” concern, I did not remove the old “overflow” parameter. And when it is set it uses the old algorithm. The reasoning being, if overshoot or overflow is desired, you don’t need an efficient algorithm as you’re just going to fade out whenever. If you paired this algorithm with an overflow, in most cases you’d be fading out right at the beginning in the first few seconds (or milliseconds) of a song which wouldn’t make much sense anyway. That is why this algorithm is not run when overflow is desired. This algorithm was built for when you DON’T want overshoot and fade and you just want the scheduler to figure out scheduling just right so that the show ends right when the next show is supposed to start as smoothly as possible. As previously mentioned you still can overshoot and fade if you want. That functionality isn’t going away. This is just an optimization in the case of not overshooting.



## Algorithm Details
You had a few incorrect ideas about how the algorithm works. To summarize the paper and simplify a bit I will walk through building up the algorithm starting with the simplest case:

The simplest case was implemented before (the random greedy algorithm (RG)). You just randomly select files until you hit the target length or there are no more files throwing out any tracks that would be too long.

This is a pretty good way of doing things and gets decent results but it can be improved quite a bit. 

The next logical step is to do the same RG selection many times (randomizing the input each time so you get different results) and select the best outcome. So you do the random selection say 50 times randomizing input track order each time and find the one that comes the closest to your target length. 

That is also pretty good and gets you really good results, but still can be improved by going back over all the tracks after you have made your initial selection and replacing tracks with ones that are slightly longer to get you slightly closer to the input time if there are any. This is the local improvement (LI) step of the RGLI algorithm. It does the RG trials many times but after each trial it sees if it can make it a little bit better. The best trial out of all of these is selected and returned.



## Randomness
I am assuming you are talking about randomness of the set of selected tracks (ignoring order) as that would make the most sense here. The randomness highly depends on the size of input data, length of input data, and the target length. The implemented algorithm does a random search through the entire space of combinations and finds the best one from all of it’s trials (with some local improvement). Since it is by nature a random search the output will be random, unless you have a low number of tracks and a total track length close to (but greater than) the desired target show length. 

If the input tracks are less then the desired length there is no variation because all tracks get selected every time. No discernible difference from the old algorithm.

If the input tracks in the database is large (tracks in db > 500), the db selects 500 random every time anyway and you get a completely different set each time anyway. No difference from the old algorithm, except you get much better scheduling.
I did some entropy calculations on some runs.  I focused on the more interesting cases where 10 < N < 500. A brief introduction to entropy if you are not familiar:

Entropy is a measure of randomness. The higher the entropy, the more random something is. Told you it was brief!

### Data

In my testing I was computing entropy across 100 runs of the algorithms on the same items. If the items that came out were the exact same as another run (ignoring order) that counts against the randomness. It is important to keep in mind that a lot of the differences between the algorithms were around where the total length of the tracks is only slightly longer than the total length. So you get a lot of repetition between runs, they just aren’t the exact selection. For example, when testing a target length of an hour, and 20 tracks (around 3+ mins each), you end up with 15 or 16 chosen per run. Now, 20 choose 15 is 15504 different ways to do that. Now most of those have repetition between them but aren’t the exact same set. So for my testing I treated them as completely different as we were interested in randomness across distinct sets of runs. 

The testing was done with random tracks from my library (My library contains 347ish hours of audio). For almost all cases the algorithms produced equally random results. The randomness differs in a few different edge cases It is mostly noticable when you have a low number of tracks and a total track length close to (but greater than) the desired target show length. For pretty much all other edge cases the randomness is similar. Let me know if you want the code used to make the data.

Now for the data!!
image
image
image
image


### Code used to generate data

```kotlin
import com.github.doyaaaaaken.kotlincsv.dsl.csvWriter
import java.io.File
import javax.sound.sampled.AudioSystem
import javax.sound.sampled.UnsupportedAudioFileException
import kotlin.math.log2
import kotlin.random.Random
import kotlin.system.measureNanoTime
import kotlinx.coroutines.Dispatchers
import kotlinx.coroutines.async
import kotlinx.coroutines.runBlocking
import kotlinx.coroutines.withContext
import org.tritonus.share.sampled.file.TAudioFileFormat

private fun <T> benchmark(block: () -> T): T {
    val result: T
    val time = measureNanoTime {
        result = block()
    }
    println("Took ${time / 1000000.0} ms")

    return result
}

fun main() {
//    entropyTesting()
    entropySweepNTarget1H()
    entropySweepNTarget2H()
    entropySweep20NTarget1H()
    entropySweepTargetN100()
    entropySweepTargetN500()
}

// interested in:
// entropy for sweeping item size items compared to entropy for full randomness for a high target
// entropy for sweeping target length with fixed size

// Entropy could be measured in two different ways:
// same set chosen every time (this is what we are doing)
// same song chosen across times

private fun Iterable<Pair<Int, Pair<Double, Double>>>.write(name: String) {
    csvWriter().open(name) {
        writeRow(listOf("N", "RGLI", "Random"))
        forEach { row ->
            writeRow(row.first, row.second.first, row.second.second)
        }
    }
}

private fun entropySweepNTarget1H() {
    val result = sweepItems(60 * 60.0)

    result.write("sweepNTarget1H.csv")
}

private fun entropySweepNTarget2H() {
    val result = sweepItems(60 * 60.0 * 2)

    result.write("sweepNTarget2H.csv")
}

private fun entropySweep20NTarget1H() {
    val result = sweepItems(60 * 60.0, 10..30 step 1)

    result.write("sweep20NTarget1H.csv")
}

private fun entropySweepTargetN20() {
    val result = sweepTarget(20, 2000..6000 step 100)

    result.write("sweepTargetN20.csv")
}

private fun entropySweepTargetN100() {
    val result = sweepTarget(100, 22000..25000 step 100)

    result.write("sweepTargetN100.csv")
}

private fun entropySweepTargetN500() {
    val result = sweepTarget(500)

    result.write("entropySweepTargetN500.csv")
}


private fun entropyTesting() {
    val rand = Random(38462894653)
    val items = 200
    val musics = getMusic(items, rand).map(Music::toItem)

    val (rgliS, randS) = benchmark {
        computeEntropyFor(musics, 3600.0, 50000)
    }
    println("RGLI entropy: $rgliS bits")
    println("Random entropy: $randS bits")
}

private fun sweepItems(target: Double, range: IntProgression = (10..500 step 10), trials: Int = 100) = range.map { num ->
    println("On step $num")
    val items = getMusic(num).map(Music::toItem)
    num to computeEntropyFor(items, target, trials)
}

private fun sweepTarget(itemsCount: Int, range: IntProgression = (60..3600 step 100), trials: Int = 100) = range.map { num ->
    println("On step $num")
    val items = getMusic(itemsCount).map(Music::toItem)
    num to computeEntropyFor(items, num.toDouble(), trials)
}

private fun computeEntropyFor(
    items: List<Item>,
    target: Double,
    trials: Int,
): Pair<Double, Double> {
    val rgliSSP = RGLISSP()

    val (rgli, random) = runBlocking {
        val rgliJob = async {
            withContext(Dispatchers.Default) {
                (1..trials).map {
                    rgliSSP.solveSSP(items, target).items.toSet()
                }.entropyEven()
            }
        }
        val randomJob = async {
            withContext(Dispatchers.Default) {
                (1..trials).map {
                    RandomSSP.solveSSP(items.shuffled(), target).items.toSet()
                }.entropyEven()
            }
        }

        val one = rgliJob.await()
        val two = randomJob.await()
        one to two
    }

    return rgli to random
}

data class Music(
    val name: String,
    val length: Double,
    val path: String,
) {
    fun toItem() = Item(this, length)
}

private fun getMusic(count: Int, random: Random = Random.Default): List<Music> {
    return File("~/Music")
        .walkTopDown()
        .filter {
            it.isFile && it.extension == "mp3"
        }
        .map { file ->
            val fileFormat = AudioSystem.getAudioFileFormat(file)
            val duration = if (fileFormat is TAudioFileFormat) {
                val properties: Map<*, *> = fileFormat.properties()
                val key = "duration"
                val microseconds = properties[key] as Long?
                (microseconds!! / 1000.0 / 1000.0)
            } else {
                throw UnsupportedAudioFileException()
            }

            Music(
                file.name,
                duration,
                file.path,
            )
        }
        .filter {
            it.length in (1.0 * 60)..(20.0 * 60) && random.nextDouble() < 0.1
        }
        .take(count)
        .toList()
}

private fun <T> Iterable<T>.entropy(probabilityDistribution: (T) -> Double): Double =
    toSet().sumOf { item ->
        val prob = probabilityDistribution(item)
        -prob * log2(prob)
    }

fun <T> Collection<T>.entropyEven() = entropy { item ->
    val total = count { item == it }
    total.toDouble() / size.toDouble()
}

private fun distributionTest() = runBlocking {
    val avgSongLength = 60.0 * 3.0
    val songStandardDeviation = 60.0
    val rand = java.util.Random()
    var id = 0
    val nums = buildList {
        repeat(500) {
            add(
                Item(
                    id++,
                    rand.nextGaussian(avgSongLength, songStandardDeviation).coerceAtLeast(60.0)
                )
            )
        }
    }

    val target = 60.0 * 60.0 * 24.0

    val result = benchmark { RGLISSP().solveSSP(nums, target) }

    val resultSum = result.sum
    println("Result Sum: $resultSum")
    println("off by ${target - resultSum} seconds")
    println("elements: $result")
}

data class Solution(
    val items: List<Item> = emptyList(),
    val sum: Double = items.sumOf(Item::length),
) {
    fun add(item: Item) = Solution(items + item, sum + item.length)
    fun isCloseEnough(delta: Double) = delta < 0.25
    fun replace(old: Item, new: Item) = Solution(items.map { if (it == old) new else it })
}

data class Item(
    val id: Any,
    val length: Double,
)

fun interface SSPAlgorithm {
    fun solveSSP(items: List<Item>, target: Double): Solution
}

object RandomSSP : SSPAlgorithm {
    override fun solveSSP(items: List<Item>, target: Double): Solution {
        return items.fold(Solution()) { solution, item ->
            val new = solution.add(item)
            if (new.sum <= target) new else solution
        }
    }
}

class RGLISSP(val trials: Int = 50) : SSPAlgorithm {
    override fun solveSSP(items: List<Item>, target: Double): Solution {
        var best = Solution()
        repeat(trials) {
            val shuffled = items.shuffled()

            val initial = shuffled.fold(Solution()) { acc, item ->
                val new = acc.add(item)
                if (new.sum <= target) new else acc
            }

            val localImprovement = initial.items.shuffled().fold(initial) { solution, item ->
                val delta = target - solution.sum
                if (solution.isCloseEnough(delta)) return@fold solution

                val replacement = (items - solution.items.toSet()).filter {
                    it.length > item.length && it.length - item.length <= delta
                }.maxByOrNull(Item::length) ?: return@fold solution

                solution.replace(item, replacement)
            }

            if (localImprovement.sum > best.sum) {
                best = localImprovement
            }
            if (best.sum == 0.0) {
                return best
            }
        }

        return best
    }
}
```

jooola:
Thanks for the detailed explanation! I really appreciate the effort. Sorry that it took me so long to get back to you.

I did indeed mix up the different algorithms in my understanding!

The code looks good, but before I merge this, I'll let it run on my instances to see how it behaves in a real scenario.

dakriy:
@jooola any update on this?

dakriy:
Got some feedback from my station operator on this change. He said that before we were getting dead air times of up to a minute or more. Now he says he hasn't seen dead air time of more than 1 second, and it's often the exact time.

jooola:
I've had it running in our infra as well, but haven't had the chance to come back to the PR yet. I'll try to find some time soon.

dakriy:
I would really like for this to go out in 4.3.0, would that be possible?

jooola (approved):

jooola:
Thanks for the contribution and the patience :)

PR #3037, opened 2024-06-19
perf: optimize the rabbitmq health check

conet:
### Description

Running `rabbitmq-diagnostics` every 30s is a resource intensive operation wasting cpu cycles

**This is a new feature**:

No

Not necessary

### Testing Notes

**What I did:**

_We reduced the CPU load generated by the composition to about half of the previous value_

_Run the current composition for a while, record cpu usage then run the new health check and observe much lower cpu usage. Or simply compare the execution time of the two commands in the rabbitmq container. The second command does not perform all checks but I think it is good enough considering the performance gain._


jooola:
Hey thanks for the contribution!

Could you provide us the data that back your proposal up? I don't see this as a huge performance improvement, and we lose some functionality.

Unless you have a real scenario where this is causing a problem, I doubt we will merge this.

You would also use a docker-compose.override.yml file to use your own health check command.

conet:
I have graphs from machines running 20+ compositions and with this change the load and cpu usage was actually reduced to half. I'm not looking to have this PR merged, I'm happy to use this optimization on our deployments but I wanted to share this with the community, a 50% cpu usage reduction ads up. I will not share the graphs but I can share this easily reproducible test:

```
$ docker compose exec rabbitmq bash
# time for in in {1..100}; do rabbitmq-diagnostics -q ping; done
Ping succeeded
Ping succeeded
...
Ping succeeded

real	0m40.225s
user	2m33.354s
sys	0m17.481s
# time for in in {1..100}; do nc -z 127.0.0.1 5672; done
real	0m0.027s
user	0m0.018s
sys	0m0.010s
```

Everyone is free to run whatever they want but I would say that one of the goals of any good project is to ship optimum defaults. Running `rabbitmq-diagnostics -q ping` is really bad for a command that is run every 30s (the default healthcheck interval), it's worse if one runs multiple rabbitmq sevices. I'm not sure what is the lost functionality here, is there a case when rabbitmq is down but it keeps the amqp port open? Anyway whoever recommended  `rabbitmq-diagnostics -q ping` as a good healtcheck probe (I found stack overflow threads about this) is responsible for a lot of CPU usage out there.

jooola:
Thanks for the data!

Merging now

PR #3038, opened 2024-06-19
perf: optimize the api image health check

conet:
### Description

_Having curl installed in the API container improves operations and the efficiency of the health check. Launching python to do a single http call is an overkill._

**This is a new feature**:

_No_

_Not necessary_

### Testing Notes

**What I did:**

_Compare cpu time of the new health check command to the old health check command_

_See above_


Dockerfile
```diff
@@ -174,8 +175,7 @@ CMD ["/usr/local/bin/gunicorn", \
 ARG LIBRETIME_VERSION
 ENV LIBRETIME_VERSION=$LIBRETIME_VERSION
 
-HEALTHCHECK CMD ["python3", "-c", \
-    "import requests; requests.get('http://localhost:9001/api/v2/version').raise_for_status()"]
+HEALTHCHECK CMD ["curl", "-f", "http://localhost:9001/api/v2/version"]
```
jooola, line 178:
```suggestion
HEALTHCHECK CMD ["curl", "--fail", "http://localhost:9001/api/v2/version"]
```

jooola:
Thanks!

PR #3054, opened 2024-07-05
docs: use new docker compose command

hwalker928:
Improve the Docker install documentation, along with the `docker compose` commands.

### DIFF
--- a/docs/admin-manual/install/install-using-docker.md
+++ b/docs/admin-manual/install/install-using-docker.md
@@ -14,6 +14,12 @@ This guide walk you though the steps required to install LibreTime on your syste
 
 Before installing LibreTime, you need to make sure that [Docker](https://docs.docker.com/engine/) is installed on your operating system and **up-to-date**.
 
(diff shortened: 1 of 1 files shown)


docs/admin-manual/install/install-using-docker.md
```diff
@@ -14,6 +14,12 @@ This guide walk you though the steps required to install LibreTime on your syste
 
 Before installing LibreTime, you need to make sure that [Docker](https://docs.docker.com/engine/) is installed on your operating system and **up-to-date**.
 
+For a quick install of Docker, you can execute the command below:
+
+```bash
+curl -sSL https://get.docker.com/ | CHANNEL=stable bash
```
paddatrapper, line 20: This is not a good way of installing software. Piping random things from the Internet to bash is never a good solution

paddatrapper (changes requested):

paddatrapper (approved):

paddatrapper:
Thanks!

PR #3082, opened 2024-09-28
feat(analyzer): parse comment fields from mp3 files

### LINKED ISSUE #526: Use standard ID3 "comment" metatag field instead of "description"

dakriy:
### Description

Upload comments from mp3 files into libretime `comments` and `description` fields.

**This is a new feature**:

Yes

No none required

### Testing Notes

**What I did:**

I uploaded tracks that contained comments into LibreTime and checked the database to ensure that the `comments` and `description` fields were correctly populated. I then went to the UI and confirmed that the description field had the MP3 comment in it inside of the metadata editor. I then uploaded some files that did not have comments to make sure I did not break any existing functionality. 

Follow the steps in what I did

### **Links**

Fixes #526


dakriy:
@jooola The tests fail on my machine for the analyzer project, specifically this part where the test does seem incorrect:

https://github.com/libretime/libretime/blob/60db15e8e5a572d4d9cfe854219e12bc25232618/analyzer/tests/fixtures/__init__.py#L156

On my computer to get the test to pass I have to set channels to 1 which should be expected.

paddatrapper:
Ignore the failing pre-commit - looks like there is a mismatched version somewhere between generation and checking the API schema

PR #3101, opened 2024-10-19
fix: intro/outro playlist unset was impossible

caveman99:
this reinstates the boolean fields in the database from the original PR to work around a foreign key contraint. THE UI remains unchanged


caveman99:
How can i pass that API Schema check? It goes through all commits and chokes on the old one, ignoring the fact it is correct now.

paddatrapper:
We need to fix that check for PRs. It'll be fine on main when this is merged because I'll squash the commits down

paddatrapper (approved):

Issue #3121, opened 2024-12-09
Running database migration after importing Airtime database deletes the data 


HighTechHarmony:
### Describe the bug

When trying to migrate from airtime per [the migrate from airtime document ](https://libretime.org/docs/admin-manual/install/migrate-from-airtime/), I import the airtime database, run the migration, and I am then left with a blank, fresh install of libretime, with the restored data from airtime being totally absent.  I also checked to see if the airtime database was somehow present under a different name, such as ```airtime``` instead of ```libretime```, this did not appear to be the case.

### To reproduce

On the previous airtime server to be migrated, the airtime database is backed up using a command similar to the following:
``` sudo -u postgres pg_dump --no-owner --no-privileges airtime > airtime.sql ```

1. Do a plain install of Ubuntu 20.04.5-amd64 to a VM.  The aforementioned database dump file is copied to the file system somewhere on the new VM.
2. Set the timezone, configure static IP, setup a DNS entry for "libretime" to resolve to this host,
3. Do all system updates ```apt-get update && apt dist-upgrade``` and reboot
```sudo apt-get install postgresql-contrib```
4. Download https://github.com/libretime/libretime/releases/download/4.2.0/libretime-4.2.0.tar.gz and extract
```cd libretime``` and run the installer:
```sudo ./install --listen-port 8080 http://libretime:8080```
5. Wait for installation to complete, no errors are seen during the process
6. Restore the airtime database dump file, per the documentation: ``` sudo -u libretime libretime-api dbshell < airtime.sql```
7. Observe that this takes some time, and from the output that tables appear to be created and altered
NOTE: Despite being specified in the documentation, I skipped restoring the storage itself for brevity at this point, because I conjectured that it would have no impact on the particular issue I am dealing with. 
8. Verify that there is data present in the libretime database.  Specifiically, open the dbshell:
``` sudo -u libretime libretime-api dbshell ```
Connect to libretime database:
``` \c libretime ```
Select some know data from a table:
``` select * from cc_show limit 10```
Observe that there are shows in the table that were imported from the airtime database dump.  
9. Exit the dbshell and proceed with the documentation order to apply database migrations:
``` sudo -u libretime libretime-api migrate```
10. Reenter the dbshell and repeat the previous test to select from cc_show
``` select * from cc_show limit 10```
11. Observe that there are no shows in the table.

Furthermore it appears evident after proceeding with the remainder of the documented procedure, authenticating and browsing the web console, that the whole system is in an out-of-the-box state.

[...]

Issue #2563, opened 2023-05-25
New install and migration from Airtime: database restore error


comiconomenclaturist:
### Describe the bug

I've just installed Debian 11 on a server and then installed libretime 3.0.2.
I then ran `sudo -u libretime libretime-api dbshell < airtime-backup` but I got the error:

```
2023-05-25 11:40:37.846 | CRITICAL | libretime_shared.config._base:__init__:50 - 1 validation error for Config
general -> public_url
  invalid or missing URL scheme (type=value_error.url.scheme)
```

The airtime backup is from Airtime 2.5.1 - is this supported? Any idea how I can import the database?

### To reproduce

1. Install OS
2. Install Libretime 3.0.2
3. Restore Airtime database

### Expected behavior

The database should be restored.

### Relevant log output or error messages

```shell
2023-05-25 11:40:37.846 | CRITICAL | libretime_shared.config._base:__init__:50 - 1 validation error for Config
general -> public_url
  invalid or missing URL scheme (type=value_error.url.scheme)
```

### LibreTime version

3.0.2

### Installation method and OS / Environment

Operating system: Debian 11
Method: Installer script

### Client Environment

Operating system:
Browser:

paddatrapper:
You need to configure the public url in the configuration file. See https://libretime.org/docs/admin-manual/setup/configuration/#general

comiconomenclaturist:
Thanks. I don't have a resolvable FQDN just yet so I tried putting in the local IP. 
I now get the error:

```
\connect: FATAL:  database "airtime" does not exist
CommandError: "psql -U libretime -h localhost -p 5432 libretime" returned non-zero exit status 2.
```

Is it possible to gert Libretime up and running without the FQDN and SSL certs for testing?

paddatrapper:
Looks like your configuration file has different DB connection details that what the DB has. You don't need SSL certs and using localhost as the public_url initially will work. Are you following the migration guide? https://libretime.org/docs/admin-manual/setup/migrate-from-airtime/

comiconomenclaturist:
Yes, that's the guide I've been following. I did the initial install, and no setup. The next step was just trying to restore the Airtime DB, but the new install's DB is called libretime, whereas the Airtime backup is called airtime. I guess that's the problem? What would be the best way to solve this? Rename libretime to airtime in psql and update `/etc/libretime/config.yml`?

paddatrapper:
[...]

PR #3123, opened 2024-12-11
fix(legacy): migrations from airtime 2.5.1

HighTechHarmony:
### Description

This fixes various problems in legacy migrations that were preventing a successful database migration from Airtime 2.5.1.  Previously, following [the procedure](https://libretime.org/docs/admin-manual/install/migrate-from-airtime/) using the migrations provided in the Libretime 4.2.0 installer, without the fixes in this PR, would either fail completely, or cause all of the imported data to be completely deleted. 

_migrations.py
If schema_version is not found in the table cc_prefs, it then checks for system_version to have a value of '2.5.1' (in case this is an airtime 2.5.1 migration which will not have any schema_version in cc_pref).  If found, it prevents loading of the schema file, which is critical to preserving the imported Airtime data.

0006_2_5_5
Removed a redundant addition of the image_path and description columns to cc_show (done in earlier migration 003_2_5_2)

0015_2_5_17
Fixed a syntax error with adding the artwork column to cc_files

0023_3_0_0_alpha_9_1
Removed a redundant addition of the artwork column to cc_files (done in earlier migration 0015_2_5_7)

### Documentation Changes

The [airtime migration documentation](https://libretime.org/docs/admin-manual/install/migrate-from-airtime/) already suggests a procedure to be followed, it just didn't work because of problems within these migrations.  The procedure as documented should now work for  those coming from Airtime 2.5.1.

### Testing Notes

**What I did:**

I attempted to migrate an actual airtime 2.5.1 database from a production system containing a large amount of shows, tracks, and users.  I observed that the migration completed without errors, and that the expected system state was achieved within Libretime. Specifically, the calendar, library, authentication, and other aspects are populated with the data that was present in the migrated Airtime database, and Libretime is able to function using this data.

Install Libretime 4.2.0.  Restore a sample postrgresql database backup from an Airtime 2.5.1 server.  Apply the database migration.  Restart the services.  Login and view the library, calender, etc.

### **Links**

Closes: #3121  
May also be related to, or even close (as a duplicate): #2563

### DIFF
--- a/api/libretime_api/legacy/migrations/0023_3_0_0_alpha_9_1.py
+++ b/api/libretime_api/legacy/migrations/0023_3_0_0_alpha_9_1.py
@@ -4,6 +4,10 @@ from django.db import migrations
 
 from ._migrations import legacy_migration_factory
 
+"""
+This migration is currently an empty placeholder for 3.0.0-alpha.9.1.  Please do not remove it.  There are
+currently no actions, but it needs to remain intact because of the incremental nature of the migrations.
+"""
 UP = """
 
 """
--- a/api/libretime_api/legacy/migrations/0023_3_0_0_alpha_9_1.py
+++ b/api/libretime_api/legacy/migrations/0023_3_0_0_alpha_9_1.py
@@ -4,10 +4,12 @@ from django.db import migrations
 
 from ._migrations import legacy_migration_factory
 
-"""
-This migration is currently an empty placeholder for 3.0.0-alpha.9.1.  Please do not remove it.  There are
-currently no actions, but it needs to remain intact because of the incremental nature of the migrations.
-"""
+# This migration is currently a placeholder for 3.0.0-alpha.9.1.
+# Please do not remove it.  There are currently no actions, but it
+# needs to remain intact so it does not fail when called from the
+# migrations script. Any future migrations that may apply to
+# 3.0.0-alpha.9.1 will be added to this file.
+
 UP = """
 
 """
--- a/api/libretime_api/legacy/migrations/0023_3_0_0_alpha_9_1.py
+++ b/api/libretime_api/legacy/migrations/0023_3_0_0_alpha_9_1.py
@@ -5,11 +5,11 @@ from django.db import migrations
 from ._migrations import legacy_migration_factory
 
 UP = """
-ALTER TABLE cc_files ADD COLUMN artwork VARCHAR(4096);
+
 """
 
 DOWN = """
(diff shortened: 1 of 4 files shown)


api/libretime_api/legacy/migrations/0023_3_0_0_alpha_9_1.py
```diff
@@ -5,11 +5,11 @@
 from ._migrations import legacy_migration_factory
 
 UP = """
-ALTER TABLE cc_files ADD COLUMN artwork VARCHAR(4096);
```
paddatrapper, line 8: This is now a blank migration. Can it be removed completely?

paddatrapper: Then add a comment explaining why it needs to stay to avoid it being removed in future

HighTechHarmony: Added a comment in latest push

api/libretime_api/legacy/migrations/_migrations.py
```diff
@@ -19,8 +19,17 @@ def get_schema_version():
...
-        return row[0] if row else None
+        if row: return row[0]
 
+        # Check to see if this is an airtime 2.5.1 migration which will not return a schema_version
+        # We look for system_version to have a value of 2.5.1
+        with connection.cursor() as cursor:
+            cursor.execute("SELECT valstr FROM cc_pref WHERE keystr = 'system_version'")
+            row = cursor.fetchone()
+            if row and row[0] == '2.5.1':
+                return '0'  # A low schema version that is not None
```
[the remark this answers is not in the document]
HighTechHarmony: Latest push consolidates this into a single query. get_schema_version returns the system_version if it is airtime 2.5.1, enough to prevent the database from being nuked/paved by 0001_initial.

paddatrapper (approved):

api/libretime_api/legacy/migrations/_migrations.py
```diff
@@ -11,15 +11,27 @@ def get_schema_version():
 
     Don't use django models as they might break in the future. Our concern is to upgrade
     the legacy database schema to the point where django is in charge of the migrations.
+
+    An airtime 2.5.1 migration will not have schema_version, in that case, we look for
+    system_version to have a value of 2.5.1 and return that as the schema version value
+    (really just needs to be anything besides None, so that the next migration doesn't overwrite the database)
```
paddatrapper, line 17: This line just needs to be broken up into 2 to fix the linting error

Issue #3113, opened 2024-11-23
libretime-api service can't start from a fresh install


HighTechHarmony:
### Describe the bug

The libretime-api service doesn't seem to run from a fresh libretime 4.2.0 tarball install following the [install instructions](https://libretime.org/docs/admin-manual/install/install-using-the-installer).
tarball used: https://github.com/libretime/libretime/releases/download/4.2.0/libretime-4.2.0.tar.gz

### To reproduce

1. Do a plain install of Ubuntu 20.04.5-amd64 to a VM 
2. Set the timezone, configure static IP, setup a DNS entry for "libretime" to resolve to this host, 
3. Do all system updates `apt-get update && apt dist-upgrade` and reboot
4. `sudo apt-get install postgresql-contrib`
5. Download https://github.com/libretime/libretime/releases/download/4.2.0/libretime-4.2.0.tar.gz and extract
6. `cd libretime` and run the installer: 
`sudo ./install http://libretime`
7. Wait for installation to complete, no errors are seen during the process
8. Restart libretime services: `sudo systemctl restart libretime.target`
9. Observe that all expected services are running, according to systemd `sudo systemctl --all --plain | egrep 'libretime|nginx|php.*-fpm'`
10. Observe that libretime-playout threads are unable to access the API
11. Observe a 503 error when trying to access the Libretime web interface
12. Observe /var/libretime/api.log file size is 0 bytes

### Expected behavior

Expected that the Libretime web interface would be accessible after restarting the services.  Also that the api starts and runs, the playout system is able to use it, and that logging info would be seen in the /var/log/libretime/api.log file.

### Relevant log output or error messages

```shell
`root@libretime:~# sudo systemctl --all --plain | egrep 'libretime|nginx|php.*-fpm'`
`libretime-analyzer.service   loaded    active   running   `
`LibreTime Media Analyzer Service`
  `libretime-api.service    loaded    active   running   `
`LibreTime API Service`
  `libretime-liquidsoap.service    loaded    active   running   `
`LibreTime Liquidsoap Service`
`libretime-playout.service    loaded    active   running   `
`LibreTime Playout Service                                                   `
`  libretime-worker.service   loaded    active   running   `
`LibreTime Worker Service                                                           `
 ` nginx.service    loaded    active   running   A high performance web server and a reverse proxy server`
`  php7.4-fpm.service      loaded    active   running   `
`The PHP 7.4 FastCGI Process Manager`
`  libretime-api.socket    loaded    active   running  `
[...]

PR #3126, opened 2024-12-15
fix(docs): clarify install documentation around specifying a non-default port

HighTechHarmony:
### Description

In the Admin Manual section dealing with running the installer and specifying a port other than the default, there was some incongruity in how the public URL should be specified. The port needs to be provided both as an argument to `--listen-port` _as well in the public URL_.  In particular, when used as shown, one of the examples as previously written would lead to the API not being able to start after an otherwise successful installation run, causing confusion.

### Testing Notes
I tested the example as shown in the updated installation document, and the system (including the API) was able to start successfully. 

### **Links**

Closes: #3113


paddatrapper:
Are you testing this with a non-default client-facing port or is the non-default port only behind the reverse proxy?

HighTechHarmony:
> Are you testing this with a non-default client-facing port or is the non-default port only behind the reverse proxy?

_Behind_ the reverse proxy, the part where the port is specified in argument(s) to the installer (or via env).  Without specifying the port in both places, the API will not even be able to start.

paddatrapper:
Hmm... Then this configuration is going to produce broken user-facing links in things like emails. The actual issue is that the API should not be using the `public_url` to connect to the service. A work around is to allow it to correctly connect to your reverse proxy. In Docker we fixed this by overriding the `LIBRETIME_GENERAL_PUBLIC_URL` passed to the API container:

https://github.com/libretime/libretime/blob/cf172d5c7c4aeb152a6a822eb2142c2cb9306fea/docker-compose.yml#L32

HighTechHarmony:
> Hmm... Then this configuration is going to produce broken user-facing links in things like emails. The actual issue is that the API should not be using the `public_url` to connect to the service. A work around is to allow it to correctly connect to your reverse proxy. In Docker we fixed this by overriding the `LIBRETIME_GENERAL_PUBLIC_URL` passed to the API container

Heard;  I think the main part of this issue is that the documentation is a little lean at best (misleading at worst?), for someone trying to set up on a non-standard port without a reverse proxy.  Then. if it's misconfigured, the resulting API startup issue is a bit tricky to troubleshoot, as there is no easy indication that this is what the problem is, and how to fix it.  The Troubleshooting section offers some tips on viewing log files from each service, but in this case the API seems to fail before it starts logging, and the web interface returns a 503, so there is not much to go on.

What would have helped me is a paragraph discussing alternative ports, where there are URL and port config settings that need to be considered (with examples), and something indicating that this can result in the API not starting up, for those who are searching the docs for answers.  An separate paragraph would also be good to cover the specifics of configuring for docker, that you pointed out.

paddatrapper:
What about including such a section or tutorial instead? Talk about how to configure things and what parts of the install need to be changed

PR #3127, opened 2024-12-22
feat(api): added filters on genre & md5 for files api

kmahelona:
### Description
Added filters for genre and md5 to the files API, e.g. `/api/v2/files?genre=soul`

**This is a new feature**: Yes

**I have updated the documentation to reflect these changes**: No
There should be a schema and docs that are generated automatically. I don't know where that is.

### Testing Notes

**What I did:**
- Used docker to deploy locally
- Confirmed filters work at http://localhost:8080/api/v2/files?genre=Soul

- `make clean dev`
- Upload some files!
- Visit http://localhost:8080/api/v2/files
- You can use the filters
Screenshot 2024-12-23 at 01 36 01
Screenshot 2024-12-23 at 01 35 56

- See above
- wrote tests to confirm filters work

### DIFF
--- a/api/libretime_api/storage/tests/views/test_file.py
+++ b/api/libretime_api/storage/tests/views/test_file.py
@@ -94,15 +94,3 @@ class TestFileViewSet(APITestCase):
         path = "/api/v2/files?genre=R%26B"
         results = self.client.get(path).json()
         self.assertEqual(len(results), 1)
-
-        path = "/api/v2/files?limit=1"
-        results = self.client.get(path).json()
-        self.assertEqual(len(results["results"]), 1)
-
-        path = f"/api/v2/files?md5={file.md5}"
-        results = self.client.get(path).json()
-        self.assertEqual(len(results), 1)
-
-        path = "/api/v2/files?genre=Soul"
-        results = self.client.get(path).json()
-        self.assertEqual(len(results), 1)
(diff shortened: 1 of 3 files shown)


api/libretime_api/storage/tests/views/test_file.py
```diff
@@ -61,3 +61,32 @@ def test_destroy_invalid(self):
...
+
+        path = "/api/v2/files"
+        results = self.client.get(path).json()
+        self.assertEqual(len(results), 2)
+
+        path = f"/api/v2/files?md5={file.md5}"
+        results = self.client.get(path).json()
+        self.assertEqual(len(results), 1)
+
+        path = "/api/v2/files?genre=Soul"
```
paddatrapper, line 90: Does the old failure here for R&B imply that you cannot filter on genres with & in them?

paddatrapper (commented):
The schema check will continue to fail, even though you've fixed it. The solution is either to ignore it or to squash your commits

api/libretime_api/storage/tests/views/test_file.py
kmahelona: Not really. I wrote "R&B" but one would have to properly encode the & to %26. I guess browsers will by default convert the space to %20. So if we used this in the front end or via the API, you'd have to make sure the query params are url encoded properly. I guess I was being lazy and thought to just change to "Soul"

paddatrapper (approved):
LGTM

paddatrapper:
Failing test is fixed in main, so merging

PR #3144, opened 2025-03-18
fix(legacy): scheduled show length should not add track fade times

dakriy:
### Description

When generating a show that contains a playlist with dynamic blocks, the show will be incorrectly under scheduled because the fade time was added onto the total track length. Fade should not add time to an overall track length as then it wouldn't be a fade.

**This is a new feature**:

No

No, none needed.

### Testing Notes

**What I did:**

I created a playlist with 3 dynamic blocks:

1. 10 minute length dynamic block
2. 1 item dynamic block
3. fill rest of show dynamic block

I then created a show with the defined auto loading playlist. I then saw that on auto-load the show was not under scheduled.

See testing notes


legacy/application/models/Scheduler.php
```diff
@@ -1346,14 +1346,11 @@ public function removeItems($scheduledItems, $adjustSched = true, $cancelShow =
...
-        foreach ($files as $file) {
-            $timeLength += Application_Common_DateHelper::playlistTimeToSeconds($file['cliplength']);
-            $timeLength += $file['fadein'];
-            $timeLength += $file['fadeout'];
-        }
-
-        return $timeLength;
+        return array_reduce(
+            $files,
+            fn ($acc, $file) => $acc + Application_Common_DateHelper::playlistTimeToSeconds($file['cliplength']),
```
paddatrapper, line 1351: This should include either the `fadein` or `fadeout` consistently. Otherwise it will underestimate the time length. For example, if you have 2 tracks with the characteristics fade in 15 seconds, clip length of 30s and a fade out of 15s. This will calculate the length as 60s, when it should be 75s.

dakriy: @paddatrapper I guess I don't understand how fade is supposed to work. Could you help clarify it for me?

I thought that it was just a fade in/fade out time applied to a track. So if you have a track that is 30 seconds with 10 second fade on either end, the first 10 seconds will be fading in, and the last 10 seconds will be fading out, but the whole length doesn't change. And a cross fade would take away time from a track if anything.

dakriy: @paddatrapper I was playing with fade in and fade out time, and I don't hear that extra 15 seconds you mentioned that should be there, so I must be doing something wrong.

1. I setup the fade in/fade out time to be 30 seconds
2. I scheduled a show with some tracks in it
3. I noted that I heard each track fading in/out
4. I also noted that the show length and track lengths were each as advertised and DID NOT include an extra fade in or fade out time.


Also from the scenario you mentioned, what if fade in and fade out times were different lengths? Which one should you include if it's only one?

paddatrapper: This is the line that makes me thing that the clip length is the time between fades - https://github.com/libretime/libretime/blob/373140fc875f3f9c7cf5c62e7260de88e6bf155c/legacy/application/models/Scheduler.php#L236. 

Here the `cliplength` property is calculated as the difference between the fade out point and the fade in. So I'm not sure how to handle differing fade times. We could err on the side of using the default `crossfadeDuration` maybe?

paddatrapper (approved):

PR #3146, opened 2025-03-18
feat: configure initial framework for new web UI

paddatrapper:
### Description

This implements a Vue-based frontend utilising Vuetify, Storybook and Cypress. It is based on #2546 with some of the work @jooola implemented in https://github.com/jooola/libretime/tree/webui.

I still need to implement an About page that can then be integrated into the PHP UI.

**This is a new feature**:

Yes. Though this will not be integrated much when deployed, as it is the initial framework.

_Are there documentation changes required as a result of these changes? See
https://github.com/libretime/libretime/wiki/Documentation-Requirements_

### Testing Notes

**What I did:**

```bash
cd frontend
yarn dev
yarn storybook
yarn cypress open
```

You can see the Help page at http://localhost:3000/dashboard/help


paddatrapper:
This is pretty much now at a point where it can be reviewed. I have not integrated it into the legacy UI, mainly because I am at a loss as to how exactly to do that. If anyone has any suggestions that would be great!

zklosko:
Is this still a project we're working on? I wanted to know if we still wanted help with this.

paddatrapper:
@zklosko yes I still think that the current PHP UI is unsustainable. The larger issue for me is that I no longer run LT anywhere, so I haven't been able to dedicate much time at all to working on it. Everything has languished whenever @jooola hasn't been around. I don't see my availability changing much in the near term either. Please do go ahead if you are keen to work on this. I would love to see something land and grow, I just don't have the time to drive it.

PR #3149, opened 2025-03-21
feat(playout): high availablility playout

dakriy:
### Description

Allow playout block to be duplicated for high availability.

**This is a new feature**:

Yes

Yes

### Testing Notes

**What I did:**

I configured a show, started up 3 playout blocks, each providing one stream to Icecast. I then verified that I could hear each audio output in Icecast and verified that each one responded correctly to events sent from LibreTime (show cancel, next track, editing the schedule). I then verified I could start and stop any one of the playout blocks without interrupting either of the other streams. 

See above.

### DIFF
--- a/docs/admin-manual/tutorials/setup-high-availability-playout.md
+++ b/docs/admin-manual/tutorials/setup-high-availability-playout.md
@@ -40,7 +40,7 @@ The sections that need configured are
 
 - [General](../configuration.md#general)
   - Only `public_url` and `api_key` are needed. `cache_ahead_hours` currently has
-  no effect as playout is hardcoded to cache ahead 1 day.
+    no effect as playout is hardcoded to cache ahead 1 day.
 - [RabbitMQ](../configuration.md#rabbitmq)
 - [Playout](../configuration.md#playout)
   - Make sure to configure the associated liquidsoap host for this block.
--- a/docs/admin-manual/tutorials/setup-high-availability-playout.md
+++ b/docs/admin-manual/tutorials/setup-high-availability-playout.md
@@ -1,0 +1,54 @@
+---
+title: How to setup high availability playout
+---
+
+This tutorial walks you through the steps required to setup playout for high availability.
+
+:::info
+
+We assume you have a way to consume multiple audio inputs (for example icecast). For terrestrial HA you might consider 
+having an audio processor with switching and silence detection capabilities.
+
+:::
+
+## 1. Initial LibreTime setup
+
+When initially setting up LibreTime, first setup the components in
+the [Create the schedule block](../../contributor-manual/design/architecture.md#create-the-schedule) along with
+the message queue.
+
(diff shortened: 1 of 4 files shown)


paddatrapper (commented):
I'm not sure if we should link to something more concrete than the architecture docs? I get why you have - it makes it independent of the installation method, but maybe add some examples for Docker (I'm assuming that's how you're running this.). It's definitely easier to document than the installer approach.

dakriy:
My thought process was that if you're wanting to do HA, then you are a much more advanced user. And more concrete details are going to make it harder to figure out what exactly needs to happen. If you are doing HA, a lot of times it's not just a simple docker install. In my case I'm going to be running 3 playout blocks. 2 of them are going to be bare metal installs on a VM.

Additionally it's hard to give a one size fits all for HA stuff as everyone's requirements are different. I was just trying to provide general instructions and information for how to make it happen. Any specifics that are given are not likely to be useful for anybody else.

I linked to the architecture docs because it was the single most useful thing for trying to figure out how to do HA. Even then it left more questions than answers, and I had to spend a few days in the code to feel like I knew what was going on.

If you still think specifics are needed, I can put something together with the docker install, but I'm worried it will make things less clear. Your call though.

jooola:
I really like your experiment, but could we put all those behind feature flags, e.g. a flag in the config file or using en vars.

I believe there will be more to change to support all features in HA, so I'd go with an incremental approach and make sure we don't break the existing.

For example, playout is also sending information back to the API, which this PR does not handle yet.

dakriy:
What else is missing for HA? I would like to turn this from experiment into production ready. Only thing I saw for playout contacting API was notifying liquid soap was ready, which in this scenario is fine. I will look for other instances.

dakriy:
@jooola 

Hey, so I agree and generally like putting new features that change things behind feature flags like you were suggesting. However, in this case current documented behavior does not change at all, and so adding a feature flag would just enable functionality that could just work out of the box because it doesn't change any of the default behavior. No current production systems would break by changing the queue to a fanout as behavior is the exact same in a 1:1 scenario.

The fanout was the only real functional change. The other commit removing the `self.liquidsoap.clear_all_queues()` call doesn't change any behavior because it is called on startup in `fetch.py:main` and the queue is immediately pruned based on the current LT schedule via a `update_schedule` command therefore no clearing is necessary here anyway.

Is a feature flag really needed for just changing the Rabbit Queue to a fanout which doesn't change or break any current documented functionality?

dakriy:
@jooola 

I now have a separate use case for this code change besides high availability. I am setting up a program that does some extra stuff dealing with schedule and playout, and I would like to listen to the rabbitmq queue to know when the schedule gets updated. This code change allows for being able to sidecar without affecting libretime's normal operations and still be able to hook into events internally.

dakriy:
Ok, I split this into 3 PR's. See #3161 #3162 and #3163

PR #3160, opened 2025-06-06
fix: when metadata has a newline playout stops

dakriy:
### Description

If the title or artist metadata in libretime has a newline, that file is not played as telnet takes the newline as a command enter in the middle of the command.

**This is a new feature**:

No

No

### Testing Notes

**What I did:**

I spun up a fresh local LibreTime instance, uploaded a track with a newline in the creator field. I then created a show and added the track to the show and verified that the track did not play with error:

```
playout-1  | 2025-06-06 17:41:49,888 | ERROR    | libretime_playout.liquidsoap.client._client:_set_var:50 - ERROR: unknown command, type "help" to get a list of commands.
```

I then added my fix and verified that the track did play as it was supposed to.

See the above section


dakriy:
Might be good to do that there too although it won't fix playout issues with files already uploaded with newlines in the metadata. Apparently there are many files already in our library with a newline in some of the metadata fields :man_facepalming:  (such as title or creator), and this is the only thing i've seen break so far. Mostly because the liquidsoap connection is over telnet and a newline sends the command.

So I figured might as well put this in for anybody who has that sort of thing in their libraries already as it was not so obvious what the problem was when playout would just stop....

PR #3162, opened 2025-06-06
docs: document how to setup HA playout

dakriy:
### Description

With the fanout queue change, playout blocks can now be duplicated to provide multiple playout sources giving playout redundancy.

**This is a new feature**:

Not really, just documentation changes

Yes

### **Links**

Requires #3161 to be merged first

### DIFF
--- a/docs/admin-manual/tutorials/setup-high-availability-playout.md
+++ b/docs/admin-manual/tutorials/setup-high-availability-playout.md
@@ -40,7 +40,7 @@ The sections that need configured are
 
 - [General](../configuration.md#general)
   - Only `public_url` and `api_key` are needed. `cache_ahead_hours` currently has
-  no effect as playout is hardcoded to cache ahead 1 day.
+    no effect as playout is hardcoded to cache ahead 1 day.
 - [RabbitMQ](../configuration.md#rabbitmq)
 - [Playout](../configuration.md#playout)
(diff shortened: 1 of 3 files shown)


dakriy:
I branched off of #3161 so once it gets merged its changes will disappear from this PR

dakriy:
PR is failing because master is failing I believe

paddatrapper:
Yeah, the failure isn't related to your changes

dakriy:
Have been running HA playout for a week in production without issues so far.

dakriy:
This is instructions for how to setup HA at the liquidsoap (LQ) level. With a VERY quick glance at the links you posted it seems LQ is still a single point of failure. You can still configure each LQ as you would without any HA, it just allows you to have two of them with duplicate setups so you can do things like update liquidsoap without having any downtime. You can also configure two separate liquidsoap playout servers to send audio to two different sinks. For example, you can configure one to host your playout for icecast and one for playout on a physical device out through an audio card. You may not want actions on one to affect the other. 

This PR is basically documenting how to duplicate the "Play the schedule" block that is documented and talked about on this existing page: https://libretime.org/docs/contributor-manual/design/architecture/

Mainly the page says 

> Since the Play the schedule has its own requirements in terms of logic and uptime, it's handled separately from the Create the schedule block. This block needs to be able to be duplicated in a high availability context.

But it doesn't give any information on how to actually do that.

PR #3165, opened 2025-06-08
fix: remove potential deadlock in liquidsoap client

jooola:
### Description

Remove the liquidsoap connection lock to remove any chance to run into a deadlock. With this change, the client is not thread safe anymore.

Also handle connection error exception when talking to liquidsoap, e.g. when liquidsoap is restarted and closes the telnet connection.

**This is a new feature**:

No

Not relevant.

### Testing Notes

- Restart liquidsoap while playout is talking to it.
- See the connection being dropped and playout that might run into a deadlock.

BEGIN_COMMIT_OVERRIDE
fix(playout): remove potential deadlock in liquidsoap client (#3165)
END_COMMIT_OVERRIDE

### DIFF
--- a/playout/libretime_playout/liquidsoap/1.4/ls_lib.liq
+++ b/playout/libretime_playout/liquidsoap/1.4/ls_lib.liq
@@ -152,7 +152,7 @@ def input.http_restart(~id,~initial_url="http://dummy/url")
                           log(string_of(server.execute("#{id}.start"))) ;
                           (-1.)
                         else 0.5 end})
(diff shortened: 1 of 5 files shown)


BrixSat:
https://github.com/libretime/libretime/pull/3165 this worked for me

BrixSat:
It does not always work, i have problems where dj's where not disconnected automatically.

BrixSat:
nope to recent for me to saw it. applied now lets see :) thanks

PR #3176, opened 2025-06-17
feat(legacy): make autoloading playlist autofill lead time configurable

### LINKED ISSUE #3174: Configure autoloading playlist load ahead time

Autoloading playlists ruin cache ahead time. In the playout service it is hard coded to be 1 day, but if you make heavy use of autoloading playlists, in reality it is only 1 hour ahead.

I would like to be able to configure in LibreTime how far ahead auto-loading shows are loaded. Additionally the show time should be used instead of load time for resolution of the `now` macro and such.

_No response_

### Additional context

_No response_

dakriy:
### Description

Autoloading playlists ruin cache ahead time. In the playout service it is hard coded to be 1 day, but if you make heavy use of autoloading playlists, in reality it is only 1 hour ahead.

**This is a new feature**:

Yes

Yes

### Testing Notes

**What I did:**

I set the time to an invalid interval and saw the expected error message pop up. I then configured a lead time of 5 days and created a show 2 days out. I then visited the `/api/invalid` endpoint to get tasks to trigger and saw my show had been scheduled.

See what I did section.

### **Links**

Closes #3174

### DIFF
--- a/docs/user-manual/playlists.md
+++ b/docs/user-manual/playlists.md
@@ -26,7 +26,7 @@ If you want to edit the playlist content or metadata later, you can find it by *
 
 ### Auto loading playlists
 
-LibreTime will schedule tracks from a selected playlist an hour before a show is scheduled to air. This is a great way to automatically schedule weekly shows which are received via. podcasts.
+By default, LibreTime will schedule tracks from a selected playlist an hour before a show is scheduled to air. This is a great way to automatically schedule weekly shows which are received via. podcasts. This can be configured with the `general.autoload_lead_time` configuration option.
 
 ## Creating a Smartblock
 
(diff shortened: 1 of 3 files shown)


caveman99:
If you schedule "current affairs" type shows that are ingested as they are produced (e.g. a hourly news bulletin fed via intro playlist) would that not schedule old content? I was wondering the same on #3178

dakriy:
I'm not sure how those types of shows work, as I haven't messed with them. At the very least with this PR you could just not touch the setting and you'd get the current behavior as this shouldn't change any default behavior.

dakriy:
Actually, thinking about when I was digging in that code a while ago I believe those shows just have a URL that liquidsoap connects to and streams from at the time of playout. So it shouldn't affect them at all. I could be wrong though it has been a while since I've looked at that. But that was how I thought it worked.

PR #3178, opened 2025-06-18
fix(playout): obey cache ahead hours

### LINKED ISSUE #3177: Playout does not use the cache_ahead_hours as is documented.
### Describe the bug

Playout does not obey the `general.cache_ahead_hours` configuration option. It is hard coded to 1 day.

### To reproduce

1. Set cache ahead to 1 minute
2. See playout still caches stuff further out than 1 minute

### Expected behavior

I would expect the configuration option to do what it says:

```
  # How many hours ahead Playout should cache scheduled media files.
  # > default is 1
  cache_ahead_hours: 1
```

```shell

```

### LibreTime version

4.4.0

Operating system: Ubuntu 24.04 LTS
Method: Docker compose

### Installation details

See default docker-compose.yml file in the repo for developers.

### Client Environment

Client in this case is the playout docker container so:

Operating system: Debian GNU/Linux 11 (bullseye)
Browser: python-requests/2.32.3

### Screenshots

_No response_

dakriy:
### Description

Playout does not obey the general.cache_ahead_hours configuration option. It is hard coded to 1 day.

**This is a new feature**:

No

Yes

### Testing Notes

**What I did:**

I created a show in libretime that went for 8 hours. I then set the cache ahead time to 1 hour and noted that only 1 hour's worth of tracks were downloaded. I then set the cache_ahead_hours to 10 and saw that all files in the show were downloaded.

See what I did

### **Links**

Closes #3177

### DIFF
--- a/playout/libretime_playout/player/schedule.py
+++ b/playout/libretime_playout/player/schedule.py
@@ -34,11 +34,11 @@ def insert_event(events: Events, event_key: str, event: AnyEvent) -> None:
     events[key] = event
 
 
-def get_schedule(api_client: ApiClient) -> Events:
+def get_schedule(api_client: ApiClient, cache_ahead_hours: int) -> Events:
     stream_preferences = StreamPreferences(**api_client.get_stream_preferences().json())
 
     current_time = datetime.utcnow()
-    end_time = current_time + timedelta(days=1)
+    end_time = current_time + timedelta(hours=cache_ahead_hours)
 
     current_time_str = current_time.isoformat(timespec="seconds")
     end_time_str = end_time.isoformat(timespec="seconds")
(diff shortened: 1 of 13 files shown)


playout/libretime_playout/player/schedule.py
```diff
@@ -34,11 +34,11 @@ def insert_event(events: Events, event_key: str, event: AnyEvent) -> None:
     events[key] = event
 
 
-def get_schedule(api_client: ApiClient) -> Events:
+def get_schedule(api_client: ApiClient, cache_ahead_hours: int) -> Events:
     stream_preferences = StreamPreferences(**api_client.get_stream_preferences().json())
 
     current_time = datetime.utcnow()
-    end_time = current_time + timedelta(days=1)
```
dakriy, line 41: I would agree, except for the fact that it was incorrectly stating that the default was 1, and the default has really been 24 hours for quite some time. So the default behavior is not changed.

See here for the old behavior. 1 day = 24 hours.

PR #3239, opened 2026-04-02
feat(installer): add Debian 12 (Bookworm) detection and PHP 8.2 packages

k054:
### Description

adding debian bookworm support to the installed

modifyed files

`install` and  `api/packages.ini` : adding `bookworm` to the list of os versions 
`legacy/packages.ini` : php8.2 packages
`legacy/install/php-fpm/libretime-legacy.conf` : language stuff

**This is a new feature**:

Yes, starts some needed changes to install LT 4.5 on debian 12 (bookworm)

no documentation changes needed other than "it now supports debian 12 installation"

### Testing Notes

**What I did:**

Very little changes. Like 8 modified lines and 15 new ones (packages names)

_What did you do to validate this PR?_

along with other commits gets LT 4.5 running on debian 12

as mentioned, very few changes

### DIFF
--- a/legacy/install/php-fpm/libretime-legacy.conf
+++ b/legacy/install/php-fpm/libretime-legacy.conf
@@ -20,3 +20,6 @@ php_admin_value[upload_tmp_dir] = /tmp
 php_admin_value[request_order] = GPC
 php_admin_value[session.gc_probability] = 0
 php_admin_value[session.auto_start] = 0
+
+env[LANG] = en_US.UTF-8
+env[LC_ALL] = en_US.UTF-8
--- a/.github/workflows/_python-test.yml
+++ b/.github/workflows/_python-test.yml
@@ -34,6 +34,11 @@ jobs:
           restore-keys: |
             ${{ inputs.release }}-pip-${{ inputs.context }}
 
+      - name: Apply vine Python 3.11 patch
+        if: inputs.release == 'bookworm'
+        run: |
+          _vine_site=$(python3 -c "import site; print(site.getsitepackages()[0])")
+          patch -p1 --forward --reject-file=- -d "$_vine_site" < playout/vine-python311.patch || true
       - name: Test
         run: make test-coverage
         working-directory: ${{ inputs.context }}
--- a/.github/workflows/_python-test.yml
+++ b/.github/workflows/_python-test.yml
@@ -4,45 +4,40 @@ on:
       context:
         required: true
         type: string
-
       release:
         required: true
         type: string
-
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}-test-${{ inputs.release }}
   cancel-in-progress: true
-
 jobs:
   test:
     runs-on: ubuntu-latest
-
     container:
       image: ghcr.io/libretime/libretime-dev:${{ inputs.release }}
       options: --user 1001:1001
     defaults:
       run:
         shell: bash
-
     steps:
       - uses: actions/checkout@v5
-
       - uses: actions/cache@v4
         with:
           path: ~/.cache/pip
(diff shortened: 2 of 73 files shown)


paddatrapper:
Can you please add CI tests for PHP 8.2 and Python 3.11?

k054:
@paddatrapper sure thing :-)

k054:
The bookworm CI tests are failing because they also need the vine patch — vine/five.py crashes on Python 3.11 before pytest even starts. 

I can add a step to _python-test.yml that applies the patch after pip install when running on bookworm:

```
      - name: Apply vine Python 3.11 patch
        if: inputs.release == 'bookworm'
        run: |
          _vine_site=$(python3 -c "import site; print(site.getsitepackages()[0])")
          patch -p1 --forward --reject-file=- -d "$_vine_site" < playout/vine-python311.patch || true


```

Should I add it to this PR or to the vine PR (#3243)?

paddatrapper:
Add it here, seeing as it relates to the CI tests

k054:
CI is down to 2 failures, both pre-existing:

API / lint: mypy missing types-requests stub
Legacy / test (8.2): composer.lock and test dependencies require PHP ≤7.x

All other checks pass now.
Would you like me to tackle those in separate PRs? The API one also fails on main and looks like a low hanging fruit. The Legacy one is broader. it needs updated test dependencies and a regenerated composer.lock, which is a task on its own.

paddatrapper:
@k054 thanks. Yes - those can be separate PRs.

k054:
@paddatrapper cool, will do!

is there anything missing for this PR to be merged?

k054:
> @k054 thanks. Yes - those can be separate PRs.

https://github.com/libretime/libretime/pull/3249

paddatrapper:
@k054 are you working on the PHP 8.2 dependency update? I think that's the only thing blocking merging this, as until that is fixed, the CI will fail

k054:
> @k054 are you working on the PHP 8.2 dependency update? I think that's the only thing blocking merging this, as until that is fixed, the CI will fail

Fixing that would brake retrocmpatibility with debian < 12. Is that ok?

paddatrapper:
Is there no way to support both?

k054:
> Is there no way to support both?

Hey! The thing is we're on phpunit 5 (ancient) and we'd need to upgrade to phpunit 9 to support both PHP 7.4 and 8.2. The problem is phpunit 9 breaks compatibility with phpunit/dbunit, which is also abandoned with no PHP 8.2 compatible version available...

As a side note, dbunit is at first sight only used by 3 integration tests/files covering scheduling and block logic.  It's not part of the application itself, so LibreTime would continue to run fine. What we'd lose is test coverage on some core scheduling logic until those tests are rewritten without dbunit.

My approach would be to handle this as a separate PR: upgrade phpunit to 9.x and fix/rewrite what's needed to replace dbunit with direct database setup. 

That would add another PR to this bundle but I'll have some time this week to work on it if that works for you. What do you think?

paddatrapper:
@k054 thanks for your work on this. I'm happy with you replacing dbunit. The fewer dead dependencies the better!

legacy/install/php-fpm/libretime-legacy.conf
```diff
@@ -20,3 +20,6 @@ php_admin_value[upload_tmp_dir] = /tmp
 php_admin_value[request_order] = GPC
 php_admin_value[session.gc_probability] = 0
 php_admin_value[session.auto_start] = 0
+
+env[LANG] = en_US.UTF-8
+env[LC_ALL] = en_US.UTF-8
```
jooola, line 25: Why is this needed?

PR #3242, opened 2026-04-02
fix(legacy): add PHP 8.2 vendor compatibility patch for ZF1

k054:
### Description

_Short summary of what is the issue and the solution._

Adds a BIG patch file applied automatically by the installer after composer install. Fixes PHP 8.2 incompatibilities in vendored Propel1 and ZF1:

This patch addresses a lot of changes to unmaintained vendor files so they work on php8.2.

_Do the changes in this PR implement a new feature?_

**What I did:**

- Add #[\ReturnTypeWillChange] to Propel1 interface method overrides
- Add #[\AllowDynamicProperties] to 50 ZF1 classes
- Fix PropelPDO::query() signature for PHP 8 PDO compatibility
- Fix PropelOnDemandCollection ArrayObject sort method signatures

### DIFF
--- a/install
+++ b/install
@@ -742,6 +742,11 @@ else
   chown -R "$LIBRETIME_USER:$LIBRETIME_USER" "$LEGACY_WEB_ROOT"
 fi
 
+if [[ -f "$LEGACY_WEB_ROOT/vendor-zf1-php82.patch" ]]; then
+  info "applying PHP 8.2 vendor compatibility patch"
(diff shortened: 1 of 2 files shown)


paddatrapper:
@k054 please move Propel changes to https://github.com/libretime/propel1 and update the dependency we pull in from there instead of patching it here.

k054:
Moved Propel1 fixes to libretime/propel1. PR: https://github.com/libretime/propel1/pull/4

And I've also updated this PR to keep only the ZF1 fixes (renamed to vendor-zf1-php82.patch). Once the propel1 PR is merged, LibreTime's composer.json will pull in the fixed version directly.

PR #3243, opened 2026-04-02
fix(playout): add vine/five.py Python 3.11 compatibility patch

k054:
### Description

inspect.formatargspec was removed in Python 3.11. vine 1.3.0 (pulled in by kombu==4.6.11) imports it unconditionally, crashing libretime-playout on Debian 12. Adds a patch file with a formatargspec polyfill, applied automatically by the installer after the playout pip install.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -124,6 +124,10 @@ COPY playout .
 RUN --mount=type=cache,target=/root/.cache/pip \
     pip install --editable .[sentry]
 
+RUN set -eux \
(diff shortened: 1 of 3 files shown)


k054:
I didn't run any test on the docker container and I don't really know how it's being build but if you I'll give it a try this week and let you know.

k054:
@paddatrapper We'd need to add this to the Dockerfile for it to apply the patch

https://github.com/k054/libretime/commit/4cc1ca20b7bf4f66a1b85fec7e8d7651dd72c161

paddatrapper (approved):

PR #3249, opened 2026-04-29
fix(api): disable import-untyped mypy error code

k054:
mypy is treating [import-untyped] as a separate error code from ignore_missing_imports, causing the lint CI job to fail. 

Adding disable_error_code = ["import-untyped"] to api/pyproject.toml resolves it.

Fixes the failing lint job in: https://github.com/libretime/libretime/actions/runs/25017877262

### DIFF
--- a/api/pyproject.toml
+++ b/api/pyproject.toml
@@ -28,6 +28,7 @@ plugins = [
   "mypy_drf_plugin.main",
 ]
 ignore_missing_imports = true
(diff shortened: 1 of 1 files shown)


paddatrapper (commented):
Can you remove the extra whitespace added here?

k054:
> Can you remove the extra whitespace added here?

done

paddatrapper (approved):

PR #3250, opened 2026-04-29
fix(playout): remove trailing whitespace in vendor-zf1-php82.patch and vine-python311.patch files

k054:
The vine-python311.patch file had trailing whitespace on two blank lines, causing the Project / pre-commit CI check to fail on any PR touching that file.

if feels a bit strict for a CI to fail for that reason but I guess cleaner/shorter code is better somewhow...


k054:
removing trailing whitespaces from the vendor-zf1-php82.patch file too

paddatrapper (approved):

PR #3257, opened 2026-07-01
fix(legacy): use audio/flac mime type in jPlayer flac codec check

bbonora:
The audio preview player (jPlayer) silently refuses to play FLAC tracks in Chromium-based browsers.

### Cause

jPlayer probes browser support for each format via `canPlayType()` using its internal codec table, and its FLAC entry uses the obsolete `audio/x-flac` mime type:

```js
flac:{codec:"audio/x-flac",flashCanPlay:!1,media:"audio"}
```

Chromium-based browsers only recognize `audio/flac` — for `audio/x-flac`, `canPlayType()` returns `""`, so jPlayer treats FLAC as unsupported and never loads the track. Meanwhile the app-side `isAudioSupported()` check correctly tests the stored `audio/flac` mime and passes, so the preview window opens but nothing plays.

### Fix

Replace `audio/x-flac` with the standard `audio/flac` in the bundled `jquery.jplayer.min.js` (and the same stale string in `popcorn.jplayer.js` for consistency).

### Testing

Verified in a Docker dev environment: before the change, previewing an imported FLAC file opened the player but never played; after the change (browser hard-refresh), FLAC preview plays normally. MP3/OGG preview unaffected.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


paddatrapper:
Has this been fixed in newer versions of jplayer? If so, can we update to using them to avoid carrying a patch in our version?

